PR #1, opened 2017-11-07
Japanese translation for installing QISKit

rraymondhp:
This is an initial work of translating QISKit installation procedure into japanese


diego-plan9:
I'm unsure if it has already been taken into account, but please be aware that with the current setup all the content under the `documentation/` directory is overwritten automatically when a new stable release is created (by the Travis CI system on the `QISKit/qiskit-sdk-py` repository).

In practice, it would mean that the `documentation/_sources/ja` files included on this PR would get deleted next time a stable release is created: perhaps it would be a good idea to issue another PR against `qiskit-sdk-py` that includes the translated `.rst` files (under `doc/ja` or similar), and tweak a bit the build system so both the english and japanese versions of the technical documentation are generated automatically?

francabrera:
About the QISKit landing site, in #2 I have added localization to the site and added your japanese translation to the master branch which should be online soon (it takes some time for github pages to update).

For adding further languages users would only need to add their translations here: https://github.com/QISKit/qiskit.github.io/blob/master/locales.json

I'll leave this open in case @diego-plan9 wants to comment anything about the documentation.

rraymondhp:
Thanks. I have issues a PR at qiskit-sdk-py. You can close this PR. 
https://github.com/QISKit/qiskit-sdk-py/pull/140

PR #23, opened 2018-06-13
Updating pip install commands

juan-cb:
* According to https://github.com/QISKit/qiskit-acqua/blob/master/setup.py#L36 the command to install the `qiskit_acqua` library using pip would be `pip install qiskit_acqua`.

  * According to https://github.com/QISKit/qiskit-acqua-chemistry/blob/master/setup.py#L35 the command to install the `qiskit_acqua_chemistry` library using pip would be `pip install qiskit_acqua_chemistry`.

It seems the pypi links are still not working; this PR is only to help if those are the final names of the pip packages

### DIFF
--- a/src/pages/page-acqua.html
+++ b/src/pages/page-acqua.html
@@ -286,7 +286,7 @@
           <iron-pages selected="[[_selectedSectionId]]" attr-for-selected="section-id">
             <div section-id="home">
               <h3>[[localize('acqua.exampleTitle')]]</h3>
-              <pre class="terminal">$ pip install qiskit-acqua</pre>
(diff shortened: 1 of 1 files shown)


jaygambetta:
This is correct it needs fixing.

juan-cb:
The names in the setup have been changed right now to `qiskit-acqua` and `qiskit-acqua-chemistry`. So I think is better to update the pip command when they get published finally.

https://github.com/QISKit/qiskit-acqua/blob/master/setup.py#L36
https://github.com/QISKit/qiskit-acqua-chemistry/blob/master/setup.py#L35

If the name remains this, we will need to update the pypi badge (modifying the underscore)

PR #30, opened 2018-07-24
Add browse tutorial button

abdonrd:


abdonrd:
@alfrisch could you help me with the translation to German? 😃

PR #42, opened 2018-08-20
 Add Aqua Chemistry section

abdonrd:
You can check it here: https://qiskit-staging.mybluemix.net/aqua

qiskit aqua _ building algorithms for near

---

Also, @pistoia we need the GIFs as a videos. We can replace it in a new PR.


abdonrd:
@pistoia `AquaChemistry` should be `AQUAChemistry`, right?

And seems to `pyscf` package is missing.

PR #58, opened 2018-11-12
Fix Terra-0.6 sentence

nonhermitian:
- A minor change for better grammar.


abdonrd:
@alfrisch could you update the german? Thanks!

alfrisch:
I took a look and it is not a problem in German.

PR #59, opened 2018-11-12
a bit better sub-title

nonhermitian:
A slightly better sub-title wording.


nonhermitian:
@jaygambetta

abdonrd:
@alfrisch could you update the german? Thanks!

PR #61, opened 2018-12-11
Adding module to announce Q Camp. 

delapuente:
This is temporal and will be removed after 0.7


abdonrd:
The announcement will be shown for a long time? If so, we should use the `locales`.

delapuente:
No idea. I don't know if the idea is to remove it when 0.7 is out. If we want to keep it and localize, I need to add the localization ids. I'll do in a different PR.

abdonrd:
@delapuente I have already done it in this same PR. 🙂

PR #65, opened 2018-12-20
switch to import * in terra and aer examples

nonhermitian:
As discussed with @jaygambetta, we should move to * imports in the terra and aer examples.


abdonrd:
I'm not an expert in Python, but as a developer, doing an `import *` does not seem like a good idea 😕

jaygambetta:
ok im fine not doing these it is just the examples get rather complicated with the long imports quantumcircuit, quantumregister, classcialregister, exectue, aer, ibmq ... But lets close for now.

PR #76, opened 2019-01-08
Update Qiskit colors

abdonrd:


abdonrd:
@jaygambetta updated! https://qiskit-staging.mybluemix.net

PR #85, opened 2019-02-16
[Proposal] to fix issue 86

Borjagodoy:
I have detected that the error are in the componente '<code-base>' lit-html part.js method can't get this node if it's in double nesting or more. For example in the qiskit aer page we have this code: 

image

This code is working because the componente have only one nesting. But nevertheless in the aqua page we have this code.

image

We have double nesting and the lit html library get as node the text, the text isn't a html node and the code element.setAttribute failed because a string don't have this method.

I propose this fast solution but we need review the componente '<code-sample>' in all page.

### DIFF
--- a/src/components/page-aqua.js
+++ b/src/components/page-aqua.js
@@ -272,12 +272,14 @@ class PageAqua extends localize(i18next)(LitElement) {
         <div class="row limited-width">
           <div class="description">
             <h3>${i18next.t('pages.aqua.chemistryGui')}</h3>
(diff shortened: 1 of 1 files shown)


abdonrd:
Thanks @Borjagodoy!

I find the real issue (https://github.com/Qiskit/qiskit.org/commit/8db0300e1e467ba0eb9f86b9f8b0c8d925740511#diff-98d896d2173434dc860959ed43e91f49L286), so closing now.

PR #90, opened 2019-02-20
Remove target blank on documentation link

### LINKED ISSUE #89: Documentation link
Currently, when you press the documentation link in the head menu, open a new windows/tab. 

We need to open in same page/tab.

carlosazaustre:
Link in local is a not found, because the web works a SPA and documentation is other app, but in production should be work fine.

Fixed #89

### DIFF
--- a/src/components/app-shell.js
+++ b/src/components/app-shell.js
@@ -163,7 +163,7 @@ class AppShell extends localize(i18next)(connect(store)(LitElement)) {
             <a href="/aqua" ?selected=${this.page === 'aqua'}>Aqua</a>
           </nav>
           <nav class="second">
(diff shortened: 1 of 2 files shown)


abdonrd:
I just deployed to staging, and it doesn't work as you expected: https://qiskit-staging.mybluemix.net

We need to intercept the router.

Btw, to try the production server you can do this:
https://github.com/Qiskit/qiskit.org#build-application-for-production

carlosazaustre:
> I just deployed to staging, and it doesn't work as you expected: https://qiskit-staging.mybluemix.net

New changes just uploaded. On staging seems work fine. https://qiskit-staging.mybluemix.net

PR #91, opened 2019-02-20
Remove tutorial links on Elements pages

### LINKED ISSUE #88: Remove the tutorial link
image

Can we remove the tutorial link in all elements? I think that it is not clear that these all link to the tutorials.

carlosazaustre:
- All the tutorial links on Element landing pages are removed.
- Tutorials link is now on header navbar.

You can see how looks below:
 
image

Resolves #88


src/components/app-shell.js
```diff
@@ -163,6 +163,7 @@ class AppShell extends localize(i18next)(connect(store)(LitElement)) {
             <a href="/aqua" ?selected=${this.page === 'aqua'}>Aqua</a>
           </nav>
           <nav class="second">
+            <a rel="noopener" href="https://nbviewer.jupyter.org/github/Qiskit/qiskit-tutorial/blob/master/index.ipynb">${i18next.t('tutorials')}</a>
```
abdonrd: Makes sense to add a `target="_blank"` to not to lose the user, since is an external link?

abdonrd:
Also we need to do something with the menu on smaller screens...
Now with Documentation & Tutorials we have many navigation elements.

carlosazaustre:
> Also we need to do something with the menu on smaller screens...
> Now with Documentation & Tutorials we have many navigation elements.

This issue is referenced in #93

PR #202, opened 2019-08-22
new text for qiskit version 0.12

nonhermitian:


jaygambetta:
Can you add German as well. Personally I would delete the German translation as it is too much work

nonhermitian:
Nein. I know zero German.  I would agree that we remove it.  But that should be in another PR.

jaygambetta:
ok then I let you remove in another pr. Dont forget so that they stay in sync.

PR #204, opened 2019-08-23
Removed German translation

SooluThomas:
@delapuente Can you please check to make sure that the updates made in this PR will completely remove German translations? Thank you.


src/components/app-shell.js
```diff
@@ -285,7 +285,7 @@ class AppShell extends localize(i18next)(connect(store)(LitElement)) {
                 aria-label="Language"
                 .value=${i18next.languages[0]}>
               <option value="en" ?selected=${i18next.languages[0] === 'en'}>English</option>
-              <option value="de" ?selected=${i18next.languages[0] === 'de'}>German</option>
+              <!-- <option value="de" ?selected=${i18next.languages[0] === 'de'}>German</option> -->
```
delapuente: If this is dead code, we should remove it instead of commenting it. Is there plans for recovering it later? What's the point of having a selector with only one choice?

SooluThomas: I'm looking forward to `ja` translations

PR #226, opened 2019-10-06
Add both IQX and pip actions to main header

nonhermitian:
Screen Shot 2019-10-06 at 11 17 52 AM


jaygambetta:
I don't think this is a good idea.

nonhermitian:
The alternative is to have a single button and a local install and iqx section in the docs. This is what pytorch does for example. But both should be cited as places to play with qiskit.

jaygambetta:
i am ok if in the install we have a section on getting started with the iqx.

PR #261, opened 2019-10-22
Move away from cloud foundry for website deploy

mtreinish:
This PR reworks the deployment setup of the qiskit.org website away
 from using cloud foundry. The current deployment model is problematic
for several reasons. Primarily the regular downtime caused by doing an
updated deploy is noticeable and disruptive. Additionally, there is no
reason to host anything in an application the qiskit.org website is 100%
static with nothing needed to run on the server for it to work. There is no
reason we should host this content as an application running in cloud foundry.
This commit starts the process of changing our deployment model by
switching how we deploy updates to the repo. Now instead of using cloud
foundry this just uploads the static content to ibm cloud's s3 compatible
cloud object store. This provides a direct host for that content which
can be viewed in a web browser, behind a CDN. This should significantly
improve the hosting story for qiskit.org and update will now result in 0
downtime.

From here the next step will be to remove all the server pieces from this
repo and make this website about just the content that gets hosted.


mtreinish:
I've added the necessary travis secrets to the config so when this merges it should be able to decrypt the config file and upload the content.

mtreinish:
The travis failure is expected because we can't run a deploy job from an outside fork (travis will not use secrets). I have the same job setup and running in my fork though you can see it here: https://travis-ci.com/mtreinish/qiskit.org/builds/132995306

PR #262, opened 2019-10-22
Tweak polymer config to optimize for deployment

mtreinish:
This commit tweaks the settings in the polymer build to optimize the
output artifacts for running in production. Prior to this commit we were
hosting each artifact individually and without minimization. This
resulted in a lot of round trips to get each individual file. This
starts bundling all the code together so that there are fewer things to
pull and also minimizing the files to reduce the size of the files being
sent over the wire.


mtreinish:
The travis failures are expected because the deploy job only works on commits from branches on this repo.

 I have deployed this as a test to https://qiskit-test.quantum-computing.ibm.com where you can see that the number of files we're pulling is significantly smaller.

mtreinish:
> Can we use the [`deploy` stage of Travis](https://docs.travis-ci.com/user/deployment/script/) in combination with [`on` and `branch` options](https://docs.travis-ci.com/user/deployment#conditional-releases-with-on) for preventing the branches to fail in Travis?

That is independent of this PR, I am about to push another PR that reorganizes our travis job structure so that we actually run validation on open PRs and don't attempt to deploy where we can't. But it should not block this commit which is about improving what we have deployed.

mtreinish:
The travis changes are here: https://github.com/Qiskit/qiskit.org/pull/265

mtreinish:
Also you do realize that there are not actually any tests in this repo right? The test job just echoes there are no tests and returns 0

delapuente:
Yes. I'm afraid that comes from far. Happy if you want to remove that stage.

PR #361, opened 2019-12-20
Add metrics and disambiguate events

### LINKED ISSUE #256: Admins want to make informed decisions based on site metrics
Integrate with analytics for providing better metrics.

y4izus:
Fixes #256 

We need an index page for events the same way as we have for experiment to make the metrics works properly. Now are sending the data to https://debug-api.segment.com/v1/p because the mixin should be added to this index page and we don't have one. @lerongil , @delapuente , are we going to do an index for events? If yes, I will create an issue for that, but if not, I have to take another look to how to send the data

For the rest:
- Education page
- Advocates page
- Experiments index page
- Each experiment
the tracking is done properly

### DIFF
--- a/components/MdContent.vue
+++ b/components/MdContent.vue
@@ -9,11 +9,11 @@ export default class extends Vue {
 
   templateRender: Function | null = null
 
-  render (createElement) {
+  render(createElement) {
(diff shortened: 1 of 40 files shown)


delapuente:
This is done, @francabrera would you mind to take a look before merging?

PR #380, opened 2020-01-07
[ISSUE-377] Review Rel attribute

### LINKED ISSUE #377: Review rel attribute values for tag <a>
There are three attribute values for `rel` that we have to have in mind:
- `noopener`: use it on all links with target _blank to avoid a malicious attack coming from the opened tab. _**We are already using it**_
- `noreferrer`: use it as similar as `noopener`, especially for older browsers. It makes sense to use them both. It will change the referred page from "coming from X" to "coming from direct search" (keep in mind if it's used for internal links)
 - `nofollow`: use it for SEO purposes to not to pass the link juice/PageRank to the linked page

marinaaisa:
Closes #377 

As @y4izus was telling in the issue description, I have added `rel="noopener noreferrer"` to all the links having `target="_blank"`.

I have been searching for information on the issue as well. As you can read here, I haven't added any `nofollow` attribute because nofollow should only be used in these 2 cases:
- Indicate to Google that a link is not editorial (sponsored, submitted by users, etc)
- Prevent a page of a competitor that you link to from ranking

Also I have added the attribute `rel="me"` for social media links. Because it tells Google who manage those accounts.
Extended explanation [here](https://webmasters.stackexchange.com/questions/52877/should-i-use-rel-nofollow-for-social-media-links).

### DIFF
--- a/components/Menu.vue
+++ b/components/Menu.vue
@@ -1,7 +1,7 @@
 <template>
   <nav>
     <section class="ibm-research">
-      <a href="https://qiskit.org" class="text-logo" target="_blank" rel="noopener"><img class="logo" alt="Qiskit logo" src="https://qiskit.camp/img/qiskit-logo.png">&nbsp;Qiskit</a>
+      <a href="https://qiskit.org" class="text-logo" target="_blank" rel="noopener noreferrer"><img class="logo" alt="Qiskit logo" src="https://qiskit.camp/img/qiskit-logo.png">&nbsp;Qiskit</a>
     </section>
     <div class="convertible-menu">
       <input class="hamburger" type="checkbox">
--- a/components/Menu.vue
+++ b/components/Menu.vue
@@ -1,7 +1,7 @@
 <template>
   <nav>
     <section class="ibm-research">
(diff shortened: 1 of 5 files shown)


y4izus:
Sorry @marinaaisa, I forgot the label `discussion`. I put the PR on "on hold" because we should discuss some things before:
- We should ask if from the social accounts they want to know from where the traffic is coming. In this case, we should remove the `noreferrer` attribute (with it, the traffic will see as "they come from direct search" which is not the case). @delapuente, @lerongil , @CatherineKlauss  do you know who manage the statistics of the social accounts?
- There are other reasons to use `nofollow` as you can read [here](https://www.searchenginejournal.com/use-nofollow-links/). We don't have links only in the menu bar and in the footer, we also have on the experiments pages and on the events pages. We should analyze this links to take a decision about the `nofollow` attribute (notice that this links are generated from a markdown file, so there are more thing involve in this decision) 

Sorry again for the misunderstanding with the labels

P.S: Good point the `rel="me"` attribute! I didn't know it! :)

delapuente:
Closing for inactivity. I won't delete the branch yet.

PR #389, opened 2020-01-09
Create modular footer

### LINKED ISSUE #374: Make the footer modular and editable
Following #298 solution, provide the means for editing the footer content, meaning if we would like to add a new item it can be slotted in.

y4izus:
Fixes #374 

The new structure for the footer elements is in the `menuLinks.ts` file with the name `FOOTER_ELEMENTS`.

I have also created some vue components to make the footer more modular. I don't like the name `FooterSubsection` for one of the component but I haven't find something better. Suggestions are welcome :).

In the future (if we are agree with this new approach, we should review the menu bar to do the same)


components/footers/LinksList.vue
```diff
@@ -0,0 +1,56 @@
...
+    <li v-for="el in elements" :key="el.label">
+      <nuxt-link
+        v-if="el.linkType==='nuxtLink'"
+        class="footer-link"
+        :to="el.url"
+      >
+        {{ el.label }}
+      </nuxt-link>
+      <a
+        v-else
```
delapuente: For implementing `Cta` components, we also use `a` or `nuxt-link` ones. We decide when to use one or another, and also what value `target` should be by [inspecting the URL](https://github.com/Qiskit/qiskit.org/blob/master/components/ctas/Cta.vue) and using [dynamic components](https://vuejs.org/v2/guide/components.html#Dynamic-Components).

This way, if the link is something like `/advocates`, a `nuxt-link` component is used with `target="_self"`. If the URL looks like `https://something.com/external`, then we use an `a` tag with `target="_blank"` and `rel="noopener"`.

However, qiskit.org hosts two static sites inside: API documentation and the textbook. These require using the `a` tag without opening a new window and no `rel` values.

We solve this with a flag, as you do with `nuxtLink`/`aLink` but instead of being related to the underlaying implementation, it is related with the nature of the content and so, it is called `is-static`.

My recommendation is to adhere to the same solution and refactor the solution in a follow-up. If you prefer not to alter your solution right now, that's fine also. Just remember to open another issue for consolidating both strategies and refactor them. It is up to you.

y4izus: I have created [this issue](https://github.com/Qiskit/qiskit.org/issues/405) to unified both solutions

constants/menuLinks.ts
```diff
@@ -2,10 +2,28 @@ type SegmentData = {
   action: string
 }
 
+/**
+ * To use this type:
```
delapuente: What syntax are you using here?

I would adhere to the JSDoc syntax which I think TSDoc is based. Telling exactly what is the syntax for TSDoc is hard. It is spread across several sources. I've found a good starting point at [this StackOverflow answer](https://stackoverflow.com/questions/23072286/where-is-the-syntax-for-typescript-comments-documented) but I've also found more [examples](https://github.com/microsoft/tsdoc/tree/master/playground/src/samples) in the TSDoc repository.

Looking at how you documented the fields separately, there is an issue in the TSDoc repository about [documenting types](https://github.com/TypeStrong/typedoc/issues/278) that suggests converting the type into an `interface` and document each property. Could you use that style, please? You can describe the type in a general way, i.e. `Represent a navigation link for the menus of the site` (I don't know why, but imperative mood is the usual form here) and then you can go deeper field by field.

PR #390, opened 2020-01-09
[Issue 359] Enable applying Component decorator to abstract classes.

### LINKED ISSUE #359: Add type annotations to the segment mixin.
Add TypeScript annotations to qiskit-community/community.qiskit.org#105

delapuente:
Closes #359

Abstract page components are convenient in this project to enforce all
pages to add the `routeName` property for tracking purposes, and `title`
and `description` for element-specific pages.

However, the Component decorator that comes with Nuxt is not ready for
decorating abstract classes. The `abstract` modifier is TypeScript
specific and has no impact during runtime. An abstract class is the same
as a regular class, from a JavaScript perspective. This PR augments the
TypeScript definition of Component to allow abstract classes to be
decorated.

It also introduces the interface `TrackedPage` which is implemented by
the `QiskitPage` page component so all the concrete descendants must
implement the requirements imposed by `TrackedPage`. In this case, the
existence of the `routeName` property.

### DIFF
--- a/mixins/segment-analytics.ts
+++ b/mixins/segment-analytics.ts
@@ -1,5 +1,27 @@
-import Vue from 'vue'
+import Vue, { ComponentOptions } from 'vue'
 import { Component } from 'vue-property-decorator'
+import { Route } from 'vue-router'
+
+// Patch the Component decorator to work with abstract classes.
+// @ts-ignore is needed for avoiding the error:
+// "Overload signatures must all be exported or non-exported."
+// See https://github.com/microsoft/TypeScript/issues/21566#issuecomment-362462824
+declare module 'vue-property-decorator' {
+  abstract class _Abstract extends Vue {}
+  // @ts-ignore
+  function Component<V extends Vue>(options: ComponentOptions<V> & ThisType<V>): <VC extends typeof _Abstract>(target: VC) => VC;
+  // @ts-ignore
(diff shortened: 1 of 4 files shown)


mixins/segment-analytics.ts
```diff
@@ -1,5 +1,27 @@
...
+// See https://github.com/microsoft/TypeScript/issues/21566#issuecomment-362462824
+declare module 'vue-property-decorator' {
+  abstract class _Abstract extends Vue {}
+  // @ts-ignore
+  function Component<V extends Vue>(options: ComponentOptions<V> & ThisType<V>): <VC extends typeof _Abstract>(target: VC) => VC;
+  // @ts-ignore
+  function Component<VC extends typeof _Abstract>(target: VC): VC;
+}
+
+/**
```
y4izus: This comment is a docstring, so more information of use should be provided. As it only has `routeName` we can use the information of the documentation:
"The `routeName` would be a unique name for your route that would allow us to identify the page (category+page) regardless of the URL or the Title of the page changing over the time, that will help us to keep our queries the same over time. Examples: support-guides, notebooks-list, etc.

Please ensure the name is unique on each of your pages and follows kebab-case."

delapuente: You're right. Done in #d16da3c

PR #414, opened 2020-01-20
Filter README.md out when listing files

### LINKED ISSUE #230: Fix the `README.md` problem at experiments during building
Although successfull, the [build process in Travis](https://travis-ci.com/qiskit-community/community.qiskit.org/builds/131399369) throws an error related to the `README.md` at `~/content/experiments/`, probably related to trying to parse it as an actual experiment.

delapuente:
Fix #230

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -173,16 +173,25 @@ const config: Configuration = {
   generate: {
     dir: 'public',
     routes: (function () {
-      const events = fs.readdirSync(path.resolve(__dirname, 'content', 'events'))
-        .filter(isMarkdownContent)
-        .map(filename => `/events/${path.parse(filename).name}`)
-      const experiments = fs.readdirSync(path.resolve(__dirname, 'content', 'experiments'))
-        .filter(isMarkdownContent)
-        .map(filename => `/experiments/${path.parse(filename).name}`)
+      const events = getContentUrls('events')
+      const experiments = getContentUrls('experiments')
       return events.concat(experiments)
 
-      function isMarkdownContent (filename: string): boolean {
-        return path.extname(filename) === '.md' && filename.toUpperCase() !== 'README.MD'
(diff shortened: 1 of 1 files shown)


nuxt.config.ts
```diff
@@ -174,12 +174,16 @@ const config: Configuration = {
...
-        .filter(filename => path.extname(filename) === '.md')
+        .filter(isMarkdownContent)
         .map(filename => `/events/${path.parse(filename).name}`)
       const experiments = fs.readdirSync(path.resolve(__dirname, 'content', 'experiments'))
-        .filter(filename => path.extname(filename) === '.md')
+        .filter(isMarkdownContent)
         .map(filename => `/experiments/${path.parse(filename).name}`)
       return events.concat(experiments)
+
+      function isMarkdownContent (filename: string): boolean {
```
y4izus: Could you change the function name to something that represent what is doing? It's called `isMarkdownContent` but it check also if it's the `README` file

nuxt.config.ts
```diff
@@ -174,12 +174,16 @@ const config: Configuration = {
     dir: 'public',
     routes: (function () {
       const events = fs.readdirSync(path.resolve(__dirname, 'content', 'events'))
-        .filter(filename => path.extname(filename) === '.md')
+        .filter(isMarkdownContent)
```
y4izus: The way to get the events and the experiments is the same, only changing the string `events` to `experiments`. Could you create a function with the functionality in common?

delapuente: I considered it non related to this PR but I will do when adressing the other one.

nuxt.config.ts
delapuente: The idea was to highlight the `README` is not content, thus the name. I'll give another take.

delapuente:
@y4izus ready for another look!

Issue #326, opened 2019-11-25
Find a better methodology for CSS so we stop having problems with scoping


marinaaisa:

y4izus:
Could you specify more about this issue? I've already fixed one problem about the scoping and I wanted to know if it's the same. I found that the scope in some component wasn't working because of the compilation of the selectors from something like `.a` to something like `.a[data-v-f3f3eg9]`. If we want to have styles for selectors that are child of `.a` we should use a `deep selector`. See this [explanation](https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors).

For more information: 
- The issue I created (for the experiment page): #325 
- The [pull request with the solution](https://github.com/qiskit-community/community.qiskit.org/pull/119) to the problem

marinaaisa:
Hey @y4izus, since @delapuente explained you about what I'm doing, I will guess that your question is already answered. Anyways, for anyone who wants to understand what this issue is about:
I'm searching for CSS methodologies/architectures to:
- Solve component encapsulation problem
- Minimize CSS size on production to improve our web performance
- Solve naming problem
- Only use variables based on a design system

delapuente:
We solved the problem of scoping in #343, closing this in favour of testing new methodologies in #327

PR #441, opened 2020-03-03
Event card for desktop

### LINKED ISSUE #433: Implement an individual entry
Implement the event card following the design on [proposal 1](https://www.figma.com/file/GKMCqzcq7li1ZVi2OZtn0m/Qiskit-Events-index?node-id=0%3A1)

y4izus:
Closes #433 

To see the component in action go to https://qiskitorg-git-fork-y4izus-yg-event-card.qiskit.now.sh/events.

This is how it looks in different browsers:
- Firefox
event-card-firefox

- Chrome
event-card-chrome

- Safari
event-card-safari

### DIFF
--- a/components/cards/EventCard.vue
+++ b/components/cards/EventCard.vue
@@ -16,15 +16,15 @@
 
         <footer>
           <div class="event-card__info-detail">
-            <Map20 class="event-card__icon-map" />
+            <Map20 class="event-card__icon" />
             <span class="event-card__place">{{ place }}</span>
           </div>
           <div class="event-card__date-and-arrow">
             <div class="event-card__info-detail">
-              <Calendar20 class="event-card__icon-calendar" />
+              <Calendar20 class="event-card__icon" />
               <span class="event-card__date"><time>{{ date }}</time></span>
             </div>
-            <ArrowRight20 class="event-card__icon-arrow-right" />
+            <ArrowRight20 class="event-card__icon event-card__icon--purple" />
           </div>
         </footer>
       </div>
@@ -53,24 +53,6 @@ export default class extends Vue {
 </script>
 
 <style lang="scss" scoped>
-@mixin config-icons($prefix, $colors...) {
-  @each $color in $colors {
-    .#{$prefix}#{nth($color, 1)} {
-      fill: nth($color, 2);
-      position: relative;
-      top: .3em;
-      width: 1.25rem;
-      height: 1.25rem;
-    }
-  }
-}
-
-@include config-icons('event-card__icon-',
-    'map' var(--white),
-    'calendar' var(--white),
-    'arrow-right' var(--purple-60)
-);
-
 .card-link {
   text-decoration: none;
 }
@@ -79,10 +61,10 @@ export default class extends Vue {
   height: 15.875rem;
   width: 100%;
   background-color: var(--gray-70);
+  color: var(--white);
   display: flex;
 
   &__content {
-    color: var(--white);
     width: 50%;
     margin: 1rem;
     display: flex;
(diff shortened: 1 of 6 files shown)


components/cards/EventCard.vue
```diff
@@ -0,0 +1,130 @@
...
+
+        <footer>
+          <div class="event-card__info-detail">
+            <Map20 class="event-card__icon-map" />
+            <span class="event-card__place">{{ place }}</span>
+          </div>
+          <div class="event-card__date-and-arrow">
+            <div class="event-card__info-detail">
+              <Calendar20 class="event-card__icon-calendar" />
+              <span class="event-card__date">{{ date }}</span>
```
marinaaisa: It would be great if you can add the HTML's `time` tag to it. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time

components/cards/EventCard.vue
```diff
@@ -0,0 +1,130 @@
...
+            </div>
+            <ArrowRight20 class="event-card__icon-arrow-right" />
+          </div>
+        </footer>
+      </div>
+
+      <div
+        class="event-card__media"
+        :style="`background-image: ${decorate(image)};`"
```
marinaaisa: Is there any reason why you are not just writing
```
:style="`background-image: url(${image});`"
```
and delete the decorate function? 🤔

components/cards/EventCard.vue
```diff
@@ -0,0 +1,130 @@
...
+    const bgEffects = [
+      `url(${image})`
+    ]
+    return bgEffects.join(', ')
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@mixin config-icons($prefix, $colors...) {
```
marinaaisa: It's cool to use SCSS mixins but in this case you don't need them and you ended up generating more unnecessary CSS.

If you know that
```
     position: relative;
      top: .3em;
      width: 1.25rem;
      height: 1.25rem;
```
are shared thought all the icons, you should apply to all of them the class `event-card__icon` with the shared properties and then add a BEM's modifier on this class with the prefix `--`. Check more about BEM's modifier [here](http://getbem.com/naming/).


My proposed solution:

**HTML**
```
<Map20 class="event-card__icon" />
<Calendar20 class="event-card__icon" />
<ArrowRight20 class="event-card__icon event-card__icon--purple" />

```
**CSS**
```
.event-card__icon {
      fill: currentColor;
      position: relative;
      top: .3em;
      width: 1.25rem;
      height: 1.25rem;
}
.event-card__icon--purple {
      color: var(--purple-60)
}
```

Notice that I'm using `currentColor`. This will allow you to get the parent color for your SVG.

components/cards/EventCard.vue
```diff
@@ -0,0 +1,130 @@
...
+    'calendar' var(--white),
+    'arrow-right' var(--purple-60)
+);
+
+.card-link {
+  text-decoration: none;
+}
+
+.event-card {
+  height: 15.875rem;
```
marinaaisa: Try to use only 2 decimals on rems. 3 are too much and you don't need to be that precise.

components/cards/EventCard.vue
marinaaisa: Also I think that if you know that the path will always start by `/images/events/` you should try something as:
```
:style="`background-image: url(/images/events/${image});`"
```

Does it make sense?

marinaaisa: Another solution is creating a computed property:
HTML: 
```
:style="backgroundImage"
```

JS: 
```
get backgroundImage () {
    return `background-image: url(/images/events/${this.image});`
  }
```

components/cards/EventCard.vue
```diff
@@ -0,0 +1,130 @@
...
+  decorate (image) {
+    const bgEffects = [
+      `url(${image})`
+    ]
+    return bgEffects.join(', ')
+  }
+}
+</script>
+
+<style lang="scss" scoped>
```
marinaaisa: Using BEM properly as you are doing, there is no need to apply `scoped` to the styles.

delapuente: Let's keep this. I cannot assure we are using BEM everywhere. We still need to decide something regarding #326

delapuente:
Nice done, thank you.

PR #445, opened 2020-03-06
List all events

### LINKED ISSUE #434: Implement the list of all events
It should list both qiskit events and community events

y4izus:
Fix #434 

Although you can already see the list of the event cards, this PR is related only to the logic. Concerns related to the design will be in another issue

### DIFF
--- a/pages/events/index.vue
+++ b/pages/events/index.vue
@@ -30,6 +30,15 @@ import { Component } from 'vue-property-decorator'
 import QiskitPage from '~/components/qiskit/QiskitPage.vue'
 import EventCard from '~/components/cards/EventCard.vue'
 
+interface Event {
+  type: String,
+  title: String,
+  image: String,
+  place: String,
+  date: String,
+  to: String
+}
+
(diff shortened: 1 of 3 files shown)


pages/events/index.vue
```diff
@@ -39,6 +45,13 @@ import EventCard from '~/components/cards/EventCard.vue'
     return {
       title: 'Qiskit Experiments'
     }
+  },
+
+  async asyncData () {
+    const jsonEventsFile: any = await import('~/content/events/events-previews.json')
```
delapuente: Please, combine these two lines in one by extracting the `default` property here. You can leave the second annotation.

If you think the line would be doing too much, extract it to a function a try to only add type annotations to signatures.

y4izus: I think my way is easy to read, but it's true that `jsonEventsFile` doesn't represent what it is. I'm going to replace `jsonEventsFile` with `eventsModule`. I'm going to remove the `any` type too

PR #470, opened 2020-03-26
Filter events by type and location

### LINKED ISSUE #435: Filter list of events
Only the logic to filter the events. Two types of filter:
- By location: America, Europa, Asia, Africa
- By event type: Camp, Unconference, Hackathon

y4izus:
Fix #435


store/modules/events.ts
```diff
@@ -0,0 +1,77 @@
+type Event = {
```
delapuente: Non-blocking: in the context of Web applications, I tend to mistake `Event` with a JS event. Would it be possible to rename it to `CommunityEvent` or something similar?

store/modules/events.ts
```diff
@@ -0,0 +1,77 @@
...
+  },
+  mutations: {
+    setEvents (state, events) {
+      state.items = events
+    },
+    addFilter (state, payload) {
+      const { filter, filterValue } = payload
+      const filters = state[filter]
+
+      state[filter] = [...filters, filterValue]
```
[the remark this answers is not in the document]
y4izus: I prefer to use `state[filter].push(filterValue)`

store/modules/events.ts
```diff
@@ -0,0 +1,66 @@
...
+    },
+    addFilter (state, payload) {
+      const { filter, filterValue } = payload
+
+      state[filter].push(filterValue)
+    },
+    removeFilter (state, payload) {
+      const { filter, filterValue } = payload
+
+      state[filter].splice(state[filter].indexOf(filterValue), 1)
```
delapuente: Be careful, `.splice(start, 1)`  with a negative value for `start` will remove 1 item [starting from the end](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) and so, trying to remove a non-present item will remove the last filter.

store/modules/events.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+  },
+  mutations: {
+    setEvents (state, events) {
+      state.communityEvents = events
+    },
+    addFilter (state, payload) {
+      const { filter, filterValue } = payload
+
+      state[filter].push(filterValue)
```
delapuente: Per offline conversations, remember to ensure you cannot add the same filter more than once.

pages/events/index.vue
```diff
@@ -108,11 +104,39 @@ type Event = {
...
+    }),
+    isFilterChecked (filter, filterValue) {
+      const typeFilters = (this as any).typeFilters
+      const locationFilters = (this as any).locationFilters
+
+      return filter === 'locationFilters'
+        ? locationFilters.includes(filterValue)
+        : typeFilters.includes(filterValue)
+    },
+    updateFilter (filter, filterValue, selectFilter) {
```
delapuente: Can we reflect the fact that `selectFilter` act as a flag renaming it to `isSelected` or something similar?

PR #476, opened 2020-04-01
Show and filter past events in the events index page

### LINKED ISSUE #465: Show past events in the events index page

y4izus:
Fix #465


pages/events/index.vue
```diff
@@ -181,6 +185,10 @@ export default class extends QiskitPage {
       ? commit('addFilter', payload)
       : commit('removeFilter', payload)
   }
+
+  selectTab (selectedTab: boolean) {
```
delapuente, line 189: According to the [Carbon Vue story book](http://vue.carbondesignsystem.com/?path=/story/components-cvtabs--default), it seems the method's parameter is a number representing the tab index. Is there a reason for using `boolean` instead of `number`?

store/modules/events.ts
```diff
@@ -1,6 +1,8 @@
 export default {
   state: {
-    communityEvents: [],
+    activeTab: 0,
```
delapuente, line 3: Notice how the store knows about details of the view (it knows there is a tab, but could be a combo box, or a radio button or whatever), coupling both responsibilities. I'm thinking we could use a more semantic way of indicating if we want to filter past or future events by creating a type `'past'|'future'` for the active collection, replacing `activeTab`.

store/modules/events.ts
```diff
@@ -33,8 +42,13 @@ export default {
...
   mutations: {
-    setEvents (state, events) {
-      state.communityEvents = events
+    setEvents (state, payload) {
+      const { events, eventsList } = payload
+
+      state[events] = eventsList
+    },
+    setActiveTab (state, payload) {
+      state.activeTab = payload
```
delapuente, line 51: If you're not destructuring the payload cause it is a simple type, it'd be better to use a more meaningful name.

store/modules/events.ts
```diff
@@ -52,8 +66,12 @@ export default {
     }
   },
   actions: {
-    async fetchEvents () {
-      const eventsModule = await import('~/content/events/events-previews.json')
+    async fetchFutureEvents () {
```
delapuente, line 69: Curious, I would have expected to fetch just one list, then filter by date. However, since we are caching the lists once a day, I don't see anything bad with this approach. :+1:

store/modules/events.ts
y4izus: I think `setActiveTab` and `state.activeTab` give enough meaningful name. I like the use of `payload` and how it follows a consistency with the rest of mutation methods

PR #495, opened 2020-04-08
Apply requested changes for filter past events

### LINKED ISSUE #477: Implement requested changes for filter past events
The requested changes can be seen [here](https://github.com/Qiskit/qiskit.org/pull/476)

y4izus:
Fix #477


store/modules/events.ts
```diff
@@ -15,7 +16,7 @@ export { CommunityEvent, CommunityEventType, WorldLocation }
 
 export default {
   state: {
-    activeTab: 0,
+    communityEventsShowed: 'upcoming',
```
delapuente, line 19: Typo: it is `Shown` instead of `Showed`

delapuente, line 19: I like `upcoming` although, in the collection, we use `future`. Can you use the same for both? I don't mind which of them.

delapuente, line 19: To highlight the fact that the store should know nothing about the view, I'd like to avoid using the verb "show". We can use `activeSet` for referring to which of `futureCommunityEvents` or `pastCommunityEvents` are the filters acting on. I think the name, in addition to the possible values are enough to convey the meaning.

store/modules/events.ts
```diff
@@ -1,3 +1,4 @@
+type CommunityEventsToShowByDate = 'past'|'upcoming'
```
delapuente, line 1: If changing the name, the name of the type can be simplified to `CommunityEventSet` or alike.

store/modules/events.ts
```diff
@@ -62,8 +63,8 @@ export default {
 
       state[events] = eventsList
     },
-    setActiveTab (state, payload) {
-      state.activeTab = payload
+    setCommunityEventsShowed (state, payload) {
```
delapuente, line 66: You can use your new type to annotate `payload` and ensure it is called with one of the possible values.

store/modules/events.ts
y4izus: I'm going to use `upcoming`

PR #514, opened 2020-04-13
Adds placeholder when no events

### LINKED ISSUE #496: Implement placeholder for no events.
Now, when we filter in the events page and there are no results, nothing is shown:

Screen Shot 2020-04-08 at 16 14 51

We should have this message instead:
**Plan an event in your area today! (Insert CTA)**

There is another [issue to implement the CTA](https://github.com/Qiskit/qiskit.org/issues/507) because we are waiting for some info. Meanwhile, we can advance work just implementing the placeholder text

y4izus:
Fix #496 

Result:

Screen Shot 2020-04-13 at 10 47 39

### DIFF
--- a/pages/events/index.vue
+++ b/pages/events/index.vue
@@ -75,7 +75,7 @@
             :to="event.to"
           />
         </div>
-        <div v-else>
+        <div v-else class="event-page__results">
           <h3>Plan an event in your area today!</h3>
         </div>
       </div>
--- a/pages/events/index.vue
+++ b/pages/events/index.vue
@@ -76,7 +76,9 @@
           />
         </div>
         <div v-else class="event-page__results">
-          <h3>Plan an event in your area today!</h3>
+          <p class="event-page__no-events-msg">
+            No events: Plan an event in your area today!
+          </p>
         </div>
       </div>
     </div>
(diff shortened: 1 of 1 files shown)


pages/events/index.vue
```diff
@@ -132,6 +135,10 @@ export default class extends QiskitPage {
   routeName: string = 'events'
   windowWidth: Number = 0
 
+  get hasEvents (): boolean {
+    return this.$store.getters.filteredEvents.length !== 0
```
delapuente, line 139: You don't need the `$store.getters` part because you already mapped it into the object in [L109](https://github.com/Qiskit/qiskit.org/pull/514/files#diff-f5bccdcfec2d036fb031c17293a33db1R109).

pages/events/index.vue
```diff
@@ -75,6 +75,9 @@
             :to="event.to"
           />
         </div>
+        <div v-else class="event-page__results">
+          <h3>Plan an event in your area today!</h3>
```
delapuente, line 79: This copy is just information and I don't think it has a title connotation. Could you change to a paragraph, please?

delapuente, line 79: Non-blocking: knowing this is a placeholder but just-in-case, could you make start the placeholder with something indicating there are no events matching this criterion. A simple `No events: plan an event...` is enough.

PR #515, opened 2020-04-13
Add no-picture.jpg fallback to events with no picture

### LINKED ISSUE #498: Implement a fallback picture for when there is no image.
If we can't get an image from Airtable, we should be prepared with a fallback picture. The event team has chosen this one:
Screen Shot 2020-04-08 at 16 42 46

**Note**: Is better to use the original picture (more quality), no this one

delapuente:
Fix #498


hooks/event-conversion-utils.ts
```diff
@@ -67,13 +68,31 @@ function getType (record: any): CommunityEventType {
...
+function getImageAttachment (attachments: any[]): any|null {
+  for (const oneAttachment of attachments) {
+    if (oneAttachment.type.startsWith('image')) {
+      return oneAttachment
+    }
+  }
+  return null
+}
+
+function getBestSizeUrl (imageAttachment: any): string {
```
y4izus, line 88: Suggestion to improve legibility:
```
function getImageUrl (imageAttachment: any): string {
    const { thumbnails } = imageAttachment
    const { large: largeThumbnail } = thumbnails
    const noThumbnailsForImage = !thumbnails
    const noLargeThumbnailsForImage = !largeThumbnail

    if (noThumbnailsForImage || noLargeThumbnailsForImage) {
        return imageAttachment.url
    }

    return largeThumbnail.url
}
```

delapuente: I cannot destructure `undefined` in case it does not exist, but I get the idea.

hooks/event-conversion-utils.ts
```diff
@@ -67,13 +68,31 @@ function getType (record: any): CommunityEventType {
...
-  return options[Math.floor(Math.random() * options.length)]
+function getImage (record: any): string {
+  const fallbackImage = '/images/events/no-picture.jpg'
+  const attachments = record.get(RECORD_FIELDS.image)
+  const imageAttachment = attachments && getImageAttachment(attachments)
+  const imageUrl = imageAttachment && getBestSizeUrl(imageAttachment)
+  return imageUrl || fallbackImage
+}
+
+function getImageAttachment (attachments: any[]): any|null {
```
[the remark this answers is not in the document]
delapuente: What about `findImageAttachment` since what I'm looking for is "an attachment of image type"? Does it make sense?

hooks/event-conversion-utils.ts
```diff
@@ -67,13 +68,31 @@ function getType (record: any): CommunityEventType {
...
+  for (const oneAttachment of attachments) {
+    if (oneAttachment.type.startsWith('image')) {
```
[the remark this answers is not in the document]
delapuente: I'll use an intermediate aux to make the meaning explicit.

hooks/event-conversion-utils.ts
```diff
@@ -67,13 +68,33 @@ function getType (record: any): CommunityEventType {
...
+
+function getBestSizeUrl (imageAttachment: any): string {
+  const { thumbnails } = imageAttachment
+  const { large: largeThumbnail } = thumbnails || {}
+
+  if (!thumbnails || !largeThumbnail) {
+    return imageAttachment.url
+  }
+
+  return imageAttachment.thumbnails.large.url
```
y4izus, line 97: Here you can use your `largeThumbnail` variable

hooks/event-conversion-utils.ts
```diff
@@ -67,13 +68,33 @@ function getType (record: any): CommunityEventType {
...
+    }
+  }
+  return null
+}
+
+function getBestSizeUrl (imageAttachment: any): string {
+  const { thumbnails } = imageAttachment
+  const { large: largeThumbnail } = thumbnails || {}
+
+  if (!thumbnails || !largeThumbnail) {
```
y4izus, line 93: Please, try to use positive conditional as you suggested [here](https://github.com/Qiskit/qiskit.org/pull/470#discussion_r400141906)

hooks/event-conversion-utils.ts
```diff
@@ -67,13 +68,33 @@ function getType (record: any): CommunityEventType {
...
+  for (const oneAttachment of attachments) {
+    const isImage = oneAttachment.type.startsWith('image')
+    if (isImage) {
+      return oneAttachment
+    }
+  }
+  return null
+}
+
+function getBestSizeUrl (imageAttachment: any): string {
```
y4izus, line 89: For me, this function name is unclear. You are not getting the "best size url" you are getting the url of the image you can find. Correct me if I'm wrong

hooks/event-conversion-utils.ts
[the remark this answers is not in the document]
y4izus: I think `findImageOnAttachment` is better

hooks/event-conversion-utils.ts
delapuente: 🤦

Issue #507, opened 2020-04-09
Insert the CTA in the placeholder when there is no events


y4izus:
Part of #496

delapuente:
The text should be:
> Nothing here yet - let’s make it happen!

CTA to be provided by the end of the week.

PR #520, opened 2020-04-14
Replace the placeholder text

delapuente:
Part of #507


delapuente:
I've updated the type face to make it match the style (productive) in the website although I'm not convinced of this. The elements use `body-short-01` but that's too small so I chose to emphasize a bit with `body-short-02`.

delapuente:
I'd appreciate another delayed review here, @y4izus

PR #524, opened 2020-04-15
Extend const symbolic pattern to world locations and derive types from these consts

delapuente:
The const pattern for using symbolic names instead of strings is useful for avoiding typos and taking advantage of coding tools so it is extended to world location values too.

However, its use with TS may introduce some unneeded repetition when trying to write the types with the string values the symbolic names try to represent:

```ts
const SYMBOLIC_NAMES = {
  symbolA: 'Symbol A',
  symbolB: 'Symbol B'
}
type Names = 'Symbol A'|'Symbol B'
```

This PR uses [`const` assertions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions) to derive the types from the const values.

### DIFF
--- a/store/modules/events.ts
+++ b/store/modules/events.ts
@@ -37,7 +37,7 @@ const WORLD_LOCATION_OPTIONS: WorldLocation[] = [
   WORLD_LOCATIONS.africa,
   WORLD_LOCATIONS.online
 ]
-const TYPE_OPTIONS: CommunityEventType[] = [
+const COMMUNITY_EVENT_TYPE_OPTIONS: CommunityEventType[] = [
   COMMUNITY_EVENT_TYPES.hackathon,
   COMMUNITY_EVENT_TYPES.camp,
   COMMUNITY_EVENT_TYPES.unconference,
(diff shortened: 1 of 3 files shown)


store/modules/events.ts
```diff
@@ -12,22 +30,14 @@ type CommunityEvent = {
...
-const LOCATION_CATEGORIES: WorldLocation[] = ['Americas', 'Asia Pacific', 'Europe', 'Africa', 'Online']
-const TYPE_CATEGORIES: CommunityEventType[] = [
+const WORLD_LOCATION_OPTIONS: WorldLocation[] = [
+  WORLD_LOCATIONS.americas,
+  WORLD_LOCATIONS.asiaPacific,
+  WORLD_LOCATIONS.europe,
+  WORLD_LOCATIONS.africa,
+  WORLD_LOCATIONS.online
+]
+const TYPE_OPTIONS: CommunityEventType[] = [
```
y4izus, line 40: I like the correspondence `WORLD_LOCATION_OPTIONS` => `WorldLocation`. Could you use the same for `TYPE_OPTIONS` replacing it with `COMMUNITY_EVENT_TYPE_OPTIONS` (=> `CommunityEventType`)??

Issue #528, opened 2020-04-16
Unify vocabulary regarding place & location, and event location and region


delapuente:
There is ambiguity when talking about where an event takes place. We have two classifications for that, one local, another broader. There is a mismatch between how we refer to these concepts in the Airtable and in the UI. Here is the equivalence table. We should agree on a common way to refer to it so we can communicate clearly, not only between us but with the users too:

| Concept | Example | in Airtable | in UI |
|---|---|---|---|
Local | Madrid, Spain | Event Location | place |
Broader | Europe | Region | location |

delapuente:
Per offline conversation, let's use Airtable vocabulary.

PR #531, opened 2020-04-16
Rename location into region, and place into location

delapuente:
Implements #528

### DIFF
--- a/tests/hooks/event-conversion-utils.spec.ts
+++ b/tests/hooks/event-conversion-utils.spec.ts
@@ -4,17 +4,20 @@ import {
   filterWithWhitelist,
   convertToCommunityEvent,
   getTypes,
+  getLocation,
+  getRegion,
   getDates,
   getImage
 } from '~/hooks/event-conversion-utils'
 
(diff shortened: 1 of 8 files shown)


delapuente:
@y4izus could you do a delayed review on this, please?

tests/hooks/event-conversion-utils.spec.ts
```diff
@@ -108,6 +114,59 @@ describe('filterByWhitelist', () => {
...
+    expect(getLocation(nowhereEvent)).toBe('TBD')
+  })
+
+  it('defaults in region if there is no location', () => {
+    const { americas } = WORLD_REGIONS
+    const noLocationEvent = new FakeRecord({
+      name: 'Fake Conference',
+      region: americas
+    })
+    expect(getLocation(noLocationEvent)).toBe(getRegion(noLocationEvent))
```
y4izus, line 158: It is a good practice to use a raw value in the matcher function instead of a function. Could you replace `getRegion(noLocationEvent)` with `americas`?

delapuente: I'll review the code, I think I have another instance of this out there.

PR #537, opened 2020-04-17
Unify legacycta and cta

### LINKED ISSUE #474: Unify LegacyCta and Cta if possible
`LegacyCta` and `Cta` components are quite similar. We should check if they can be unified and if yes, unify them

y4izus:
Fix #474


components/ctas/Cta.vue
```diff
@@ -19,7 +19,9 @@ import { Component, Prop } from 'vue-property-decorator'
 @Component
 export default class extends Vue {
   @Prop(String) to
+  @Prop(Boolean) primary
```
delapuente, line 22: Considering this is an intermediate refactor for splitting logic from style, I would not consider this as a blocker yet but I'd recommend reviewing this interface. If this is a tristate, I think it is better if the property is only one (for instance, `type`, and values are `primary` (default), `secondary` and `tertiary`). Or, in case you want to preserve the properties separated, we would need some prioritization rule such as "the highest type win" and add `cta--primary` y `cta--secondary`, only if there are no higher types.

components/ctas/Cta.vue
```diff
@@ -19,7 +19,9 @@ import { Component, Prop } from 'vue-property-decorator'
...
   @Prop(Boolean) secondary
+  @Prop(Boolean) terciary
```
delapuente, line 24: Typo: it is `tertiary`.

PR #556, opened 2020-04-22
Refactor CompactFeature to use BEM

### LINKED ISSUE #542: Apply BEM to CompactFeature component
Part of #489

y4izus:
Fix #542

### DIFF
--- a/components/ui/CompactFeature.vue
+++ b/components/ui/CompactFeature.vue
@@ -5,8 +5,12 @@
       class="compact-feature__icon"
       :src="`/images/icons/icon${icon}White@3x.png`"
     >
-    <h2 class="compact-feature__title">{{ title }}</h2>
-    <p class="compact-feature__description">{{ description }}</p>
+    <h2 v-if="title" class="compact-feature__title">
+      {{ title }}
+    </h2>
+    <p v-if="description" class="compact-feature__description">
+      {{ description }}
+    </p>
   </section>
 </template>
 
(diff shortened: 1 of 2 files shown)


components/ui/CompactFeature.vue
```diff
@@ -16,6 +17,8 @@ import { Component, Prop } from 'vue-property-decorator'
 @Component
 export default class extends Vue {
   @Prop({ type: String, default: '' }) icon
+  @Prop({ type: String, default: '' }) title
```
delapuente, line 20: Instead of defaulting in empty strings, let's make the properties completely optional as it's the case with the icon, and check for them to prevent the title and the paragraph from showing if there is no content. 
```suggestion
  @Prop({ type: String }) title: string|undefined
```

I'm including the type for another refactor I'm proposing later.

components/ui/CompactFeature.vue
```diff
@@ -5,7 +5,8 @@
       class="compact-feature__icon"
       :src="`/images/icons/icon${icon}White@3x.png`"
     >
-    <slot />
+    <h2 class="compact-feature__title">{{ title }}</h2>
```
delapuente, line 8: Let's use a `v-if` directive to prevent rending the element if empty.

pages/advocates/index.vue
```diff
@@ -91,6 +72,10 @@ import MapSection from '~/components/advocates/MapSection.vue'
 import AdvocateCard from '~/components/advocates/AdvocateCard.vue'
 import CompactFeature from '~/components/ui/CompactFeature.vue'
 
+type Benefit = {
+  [key: string]: string
```
delapuente, line 76: Since we know the shape of the `Benefit`, it would be better to have it fully typed. I mean:
```ts
type Benefit = {
  'icon': string,
  ...
}
```

However, trying not to repeat ourselves too much, we can use:

```ts
type Benefit = Pick<CompactFeature, 'id'|'title'|'description'>
```

PR #563, opened 2020-04-23
Unify Legacy and Community headers

### LINKED ISSUE #544: Apply BEM to LegacyPresentation component
Part of #489

### LINKED ISSUE #558: Unify LegacyPresentation
Part of #488

delapuente:
Fix #558 
Fix #544 

This PR does:
* Remove the legacy presentation.
* Rework the community presentation to be able to show the same layout element pages were using.
* Create a wrapper component `ElementPresentation` to factor out the header layout used in all elements.
* Refactor to BEM as part of the refactor.


components/elements/ElementPresentation.vue
```diff
@@ -0,0 +1,103 @@
...
+
+<style lang="scss" scoped>
+.element-presentation {
+  &__description {
+    @include body-long-04();
+    margin: 1.1rem 0;
+  }
+
+  &__image {
+    width: 200px;
```
y4izus, line 78: Per offline conversations, we would like to use `rem` instead `px` when possible

PR #564, opened 2020-04-23
Removes legacy section and merge functionality with page section

### LINKED ISSUE #559: Unify LegacySection
Part of #488

delapuente:
Fix #559 

This PR does:
* Remove `LegacySection` completely
* Refactor `PageSection` to be able of accomodating element sections layout.
* Refactor `PageSection` to BEM.
* Refactor all the element pages to BEM.
* Extract common element layout in `element-page.scss`
* The same happens with the community layout. It is factored out in `community-page.scss`
* Make extensive use of [Bem's mixes](https://en.bem.info/methodology/quick-start/#mix) to reuse the code from different blocks.

This PR does not:
* Refactor community pages into BEM (just partially, be able to factor out some styles).


components/ui/PageSection.vue
```diff
@@ -0,0 +1,79 @@
...
+    &_alone {
+      width: 100%;
+    }
+
+    @include mq($until: medium) {
+      width: 100%;
+    }
+  }
+
+  &__aside {
```
y4izus, line 71: I'm not sure if `aside` is an [Element](https://en.bem.info/methodology/key-concepts/#element) or a [Modifier](https://en.bem.info/methodology/key-concepts/#modifier), what do you think @delapuente ?

components/ui/PageSection.vue
```diff
@@ -0,0 +1,79 @@
...
+    width: 50;
```
y4izus, line 72: Could you add units to this property?

pages/aqua.vue
```diff
@@ -1,53 +1,42 @@
 <template>
-  <main>
+  <main class="element terra-element">
```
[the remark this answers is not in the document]
delapuente: You're right. It is `xxx-element` where `xxx` is each of the elements. Then the color definitions go inside `.xxx-element` selector.

components/ui/PageSection.vue
delapuente: Per offline conversations, we agreed on `aside` being an element although if the name is confusing, tell me and I'll change.

PR #568, opened 2020-04-28
Refactor SyntaxHighlight component and use BEM on it

### LINKED ISSUE #554: Apply BEM to SyntaxHighlight component
Part of #489

y4izus:
Fix #554

### DIFF
--- a/components/ui/SyntaxHighlight.vue
+++ b/components/ui/SyntaxHighlight.vue
@@ -18,9 +18,9 @@ import { Component, Prop } from 'vue-property-decorator'
 
 @Component
 export default class extends Vue {
-  @Prop({ type: String, default: 'python' }) lang
+  @Prop({ type: String, default: 'python' }) lang!: string
   @Prop(String) label
-  @Prop({ type: String, default: '' }) code
+  @Prop({ type: String, default: '' }) code!: string
 
   copyToClipboard () {
     navigator.clipboard.writeText(this.code)
--- a/components/ui/SyntaxHighlight.vue
+++ b/components/ui/SyntaxHighlight.vue
@@ -8,7 +8,7 @@
     >
       Copy
     </cv-button>
-    <pre v-highlightjs="code"><code :class="`${lang} syntax-highlight__code`" /></pre>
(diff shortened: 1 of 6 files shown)


components/ui/SyntaxHighlight.vue
```diff
@@ -1,12 +1,14 @@
...
+      size="small"
       title="Copy to clipboard"
+      class="syntax-highlight__copy-button"
       @click="copyToClipboard"
     >
       Copy
-    </button>
-    <pre v-highlightjs><code ref="code" :class="lang"><slot /></code></pre>
+    </cv-button>
+    <pre v-highlightjs="code"><code :class="`${lang} syntax-highlight__code`" /></pre>
```
delapuente, line 11: The [array syntax](https://vuejs.org/v2/guide/class-and-style.html#Array-Syntax) is more succint here and the preferred way when you need to mix dynamic and non-dynamic classes.

pages/aer.vue
```diff
@@ -121,6 +89,36 @@ export default class extends QiskitElementPage {
       description: 'Counts, Memory, Statevector, Unitary, Snapshots'
     }
   ]
+
+  codeExample: string = `from qiskit import QuantumCircuit, execute, Aer, IBMQ
```
delapuente, line 93: Remove the type annotation, it should be inferred from the template string.

components/ui/SyntaxHighlight.vue
```diff
@@ -18,10 +20,10 @@ import { Component, Prop } from 'vue-property-decorator'
 export default class extends Vue {
   @Prop({ type: String, default: 'python' }) lang
   @Prop(String) label
+  @Prop({ type: String, default: '' }) code
```
delapuente, line 23: Non-blocking: add the missing types for this and other properties. If you don't want to make it optional cause you allow the empty string, use:
```suggestion
  @Prop({ type: String, default: '' }) code!: string
```

components/ui/SyntaxHighlight.vue
y4izus: Per offline conversation, I'm going to use `class` and `:class` as shown [here](https://vuejs.org/v2/guide/class-and-style.html#Object-Syntax)

PR #582, opened 2020-05-04
Refactor qiskit org menu

### LINKED ISSUE #548: Apply BEM to QiskitOrgMenu component
Part of #489

y4izus:
Fix #548 

To refactor this component, I needed to do:
- Create a new SidebarMenu component
- Create a new MenuLink component
- Create a new mixin for menu
- Create a new mixin for menu links
- Use BEM on QiskitOrgMenu component
- Use BEM on SidebarMenu component
- Use BEM on MenuLink component

While refactoring I also added styles to show the user the active page for sidebar menu as shown in the image:
- Before:
Screen Shot 2020-05-04 at 13 15 28
- After:
Screen Shot 2020-05-04 at 13 16 06


components/layouts/QiskitOrgMenu.vue
```diff
@@ -1,353 +1,143 @@
...
-    color: $inverse-01;
+    &:focus {
+      .menu__overlay {
+        opacity: 0.5;
+      }
 
-    &.nuxt-link-active {
-      border-left: 4px solid $focus;
-      padding-left: calc(3rem - 4px);
+      .sidebar-menu {
```
delapuente, line 114: This component should style only `menu`-block-related elements. What we can do is to name the `<SidebarMenu>` component with a class such as `menu__side-menu` and stylish that.

components/layouts/QiskitOrgMenu.vue
```diff
@@ -1,353 +1,143 @@
...
-    <div
+  <div class="menu">
+    <section class="menu__mobile" tabindex="-1">
+      <Menu20 class="menu__link" />
+      <MenuLink v-bind="homeLink" />
+      <div class="menu__overlay" />
+      <SidebarMenu />
+    </section>
+    <section class="menu__main-level">
+      <nav class="menu__main-level_on-left">
```
delapuente, line 10: According to BEM, [a modifier cannot be used alone](https://en.bem.info/methodology/quick-start/#a-modifier-cant-be-used-alone) and it needs to [semantically describe what the CSS is doing](https://en.bem.info/methodology/faq/#how-to-choose-a-name-for-a-modifier). I think here we only need to name the `<nav>` elements with something like `menu__navigation-section` (perhaps coming with a better name but I hope you get the idea).

components/layouts/QiskitOrgMenu.vue
```diff
@@ -1,353 +1,143 @@
...
+        <MenuLink v-bind="homeLink" />
+        <div class="menu__separator" />
+        <MenuLink
+          v-for="link in qiskitElements"
+          :key="link.url"
+          :is-active="isActive(link)"
+          v-bind="link"
+        />
+      </nav>
+      <nav class="menu__main-level_on-right">
```
delapuente, line 20: It should be another instance of `menu__navigation-section`.

components/layouts/QiskitOrgMenu.vue
```diff
@@ -1,353 +1,143 @@
...
-            }"
-            :to="communitySubLink.url"
-          >
-            {{ communitySubLink.label }}
-          </nuxt-link>
-        </nav>
-      </section>
-    </div>
-  </section>
+      <nav class="menu__second-level_on-right">
```
delapuente, line 33: This would be another example of `menu__navigation-section`.

components/ui/menu/SidebarMenu.vue
```diff
@@ -0,0 +1,102 @@
+<template>
+  <nav class="sidebar-menu">
+    <h2 class="sidebar-menu__title">
+      Elements
+    </h2>
+    <MenuLink
+      v-for="link in qiskitElements"
+      :key="link.url"
+      :class="['sidebar-menu__link', isActive(link) && 'sidebar-menu__link_active']"
```
delapuente, line 9: Non-blocking: I think we should use the pattern:
```
class="sidebar-menu__link"
:class="isActive(link) && 'sidebar-menu__link_active'"
```
Or even better:
```
class="sidebar-menu__link"
:class="{'sidebar-menu__link_active': isActive(link)}"
```

To make a clear separation between dynamic and non-dynamic classes.

components/layouts/Menu.vue
```diff
@@ -0,0 +1,164 @@
+<template>
+  <div class="menu">
```
delapuente, line 2: [It is recommended](https://vuejs.org/v2/guide/components-registration.html#Component-Names) that all the components contain a hyphen. We are using UpperCamelCase for our components which means we should use multi-word names. This is because HTML reserves its right for creating new tags, and those will be one word.

I'm not complaining about the components that already are named with one word (although [I filed #585 to take care of it at some point) but I prefer not to lose those well named. Nevertheless, I agree that `QiskitOrgMenu` is useful no more and we can replace it with `TheMenu` (as [I just discover it is recommended](https://vuejs.org/v2/style-guide/#Single-instance-component-names-strongly-recommended) by Vue.js) or something that points this is the main menu and not any other menu.

mixins/menu.ts
```diff
@@ -0,0 +1,37 @@
...
+
+@Component
+export default class extends Vue {
+  homeLink: NavLink = HOME_LINK
+  qiskitElements: Array<NavLink> = ORDERED_QISKIT_ELEMENTS
+  learnMore: Array<NavLink> = [TUTORIALS_LINK, DOCUMENTATION_LINK]
+  communityLink: NavLink = COMMUNITY_LINK
+  communitySubLinks: Array<NavLink> = ORDERED_COMMUNITY_SUB_LINKS
+
+  isPathStartingWith (linkPath: string) {
```
delapuente, line 22: Please, annotate return types of functions.

components/ui/menu/SidebarMenu.vue
```diff
@@ -0,0 +1,102 @@
...
+
+<script lang="ts">
+import Vue from 'vue'
+import { Component } from 'vue-property-decorator'
+import MenuLink from '~/components/ui/menu/MenuLink.vue'
+import menuMixin from '~/mixins/menu'
+
+@Component({
+  components: { MenuLink },
+  mixins: [menuMixin]
```
delapuente, line 47: Non-blocking: we should try to use the [`Mixins` notation](https://github.com/kaorun343/vue-property-decorator#usage).

PR #590, opened 2020-05-07
Add TheHeroMoment component

### LINKED ISSUE #583: Add the grid.
Part of #572 

Uses CSS for creating a procedural grid.

### LINKED ISSUE #587: Implement layout of hero section
Part of #572

### LINKED ISSUE #595: Title & description
Part of #572 

Adjust copy, font style and spacing for title and description.

### LINKED ISSUE #596: New CTA
Part of #572

### LINKED ISSUE #597: GitHub section
Part of #572

### LINKED ISSUE #598: Add the illustration.
Part of #572

delapuente:
Fix #587, fix #583, fix #595, fix #596, fix #597 and fix #598 

In addition to adding `TheHeroMoment` component, this PR:
* Add SCSS variables for breakpoints.
* Uses Carbon colors.
* Start defining tokens from the Carbon white theme to use in combination with those of gray-90.
* Tries to be explicit with spacing, setting 0 values when needed.
* Uses CSS sub-properties not to overspecify and allowing more room for reuse with the help of BEM mixes.


assets/scss/mq.scss
```diff
@@ -2,14 +2,20 @@
 
 $mq-responsive: true;
 
+$small: 320px;
```
y4izus, line 5: [Here](https://developer.mozilla.org/en-US/docs/Web/CSS/@media), in the  Accessibility concerns section, it says that both `em` and `px` are valid units, but `em` works better if the user changes the browser text size. Moreover, we are removing all `px` from the web, I think we should be consistent with that

components/landing/TheHeroMoment.vue
```diff
@@ -0,0 +1,87 @@
...
+
+<style lang="scss" scoped>
+@import '~carbon-components/scss/globals/scss/typography';
+
+.the-hero-moment {
+  position: relative;
+  display: flex;
+  padding: 0 $spacing-05;
+
+  &__picture {
```
y4izus, line 42: Are you using this on the code? I don't see where. If you aren't using it, please remove it :)

components/landing/VersionInfo.vue
```diff
@@ -0,0 +1,78 @@
+<template>
+  <section class="qiskit-version-info">
+    <a
```
y4izus, line 3: Just a comment: When we will merge this branch with the master one, we should remember to use the AppLink component instead of `<a>`

pages/index.vue
```diff
@@ -254,6 +250,46 @@ export default class extends QiskitPage {
...
+      -1rem center,
+      -1rem center;
+    background-repeat:
+      repeat,
+      repeat;
+
+    overflow: hidden;
+  }
+
+  &__presentation-picture {
```
y4izus, line 276: Could you take a look to picture styles? When I scroll to the right, I get this:
Screen Shot 2020-05-12 at 12 48 36

pages/index.vue
```diff
@@ -254,6 +250,46 @@ export default class extends QiskitPage {
...
+    position: absolute;
+    width: 100%;
+    top: 0;
+    bottom: 0;
+    left: 40%;
+    left: unquote("max(40%, #{$small})");
```
y4izus, line 282: `left` attribute is set twice so, the first one is not used. We should have only one

pages/index.vue
```diff
@@ -254,6 +250,46 @@ export default class extends QiskitPage {
...
+    background: url('/images/menda.png') no-repeat;
```
y4izus, line 283: Maybe "menda" is not the best semantic word to use 😄

assets/scss/mq.scss
delapuente: Makes sense and I'd want to involve the design team for translating [Carbon breakpoints](https://www.carbondesignsystem.com/guidelines/2x-grid/basics#breakpoints) to the correct values in `em` (issue #600). Right now, I preffer leaving them as they are.

pages/index.vue
delapuente: That reminds us it is a placeholder. 😄 I'll clarify because the illustration is still WIP.

pages/index.vue
delapuente: CSS function `max` is not extensively supported yet. The browsers not supporting it will drop the rule and resort to the previous value. I'll add a comment to clarify.

assets/scss/mq.scss
y4izus: I left you a comment [in the issue](https://github.com/Qiskit/qiskit.org/issues/600#issuecomment-627817733) because we can get the `em` values from the information we have

y4izus:
I'm not merging because I have a last doubt. This is the proper style for the picture for tablet size?

Screen Shot 2020-05-13 at 10 38 58

I've approved the PR in order to not blocking you in case this is correct 😄. You could merge if you want

y4izus:
Sorry but reviewing this again I found more things that should be fixed:

- For large screens (for medium and small it works), GitHub, title, description, and the CTA are not aligned while there are in the Figma design:
--> Implementation

Screen Shot 2020-05-13 at 17 28 03

-->  Figma

Screen Shot 2020-05-13 at 17 31 05

- The grid on the background has bigger squares than this implementation (I'm not sure if this is the final design or not, but I wanted to comment it... just in case :) )

delapuente:
Regarding the alignment, I'm afraid it has to be the font glyphs since the boxes are perfectly aligned. Moreover, I've tried to manually fix but the first glyph offset depends of the font size and the correction stops working when changing the size.
Captura de pantalla 2020-05-14 a las 21 12 30

pages/index.vue
delapuente: Fortunately, we've been able to remove this and use a simple positioning scheme.

pages/index.vue
delapuente: The correct picture is in place. :)

assets/scss/mq.scss
delapuente: Done in this PR.

delapuente:
@y4izus we are merging this. I've implemented all the feedback including the accessibility recommendations for the breakpoints. Regarding the font, you were right, it was different although we agreed on moving to one of the Carbon Display fonts (which auto-compute the size in terms of the viewport size) and regarding the grid, we make it bigger and follow the 8px mini-unit of Carbon's grid.

PR #592, opened 2020-05-07
Unifies AppLink component

### LINKED ISSUE #584: Unifies MenuLink and FooterLink and use the result in Cta
MenuLink and FooterLink are almost the same and also the component used on the Cta component.

This issue is related to [this one](https://github.com/Qiskit/qiskit.org/issues/405). Once finished, issue #405 could be closed too

y4izus:
Fix #584 

This PR:
- Unifies all the links and Ctas
- Extracts links to a file


components/ui/AppLink.vue
```diff
@@ -1,13 +1,17 @@
...
     :href="url"
-    :to="isInternal(url) ? url : null"
+    :to="!isStatic && isInternal(url) ? url : null"
     :rel="isExternal(url) ? 'noopener' : null"
     :target="isExternal(url) ? '_blank' : null"
+    :title="title"
     @click="segment && $trackClickEvent(segment)"
   >
-    {{ label }}
+    <template v-if="label">
```
delapuente, line 11: The link already accepts general content in the `slot`, this shortcut is not necessary or it can be achieved by extension with a wrapper component. Leave the slot only, please.

components/ui/AppLink.vue
```diff
@@ -19,6 +23,8 @@ import MenuLinkMixin from '~/mixins/menuLink'
 export default class extends Mixins(MenuLinkMixin) {
   @Prop(String) url
   @Prop(String) label
```
delapuente, line 25: No idea why this was here, honestly. Can you remove it, please?

components/ui/AppLink.vue
```diff
@@ -19,6 +23,8 @@ import MenuLinkMixin from '~/mixins/menuLink'
 export default class extends Mixins(MenuLinkMixin) {
```
delapuente, line 23: Thanks to the unification, `MenuLinkMixin` is no longer needed, all the API regarding links can live under this component. Would you mind to move its methods here and remove it, please?

components/ui/AppLink.vue
```diff
@@ -1,13 +1,17 @@
...
   <component
-    :is="isInternal(url) ? 'nuxt-link' : 'a'"
+    :is="!isStatic && isInternal(url) ? 'nuxt-link' : 'a'"
     :href="url"
-    :to="isInternal(url) ? url : null"
+    :to="!isStatic && isInternal(url) ? url : null"
     :rel="isExternal(url) ? 'noopener' : null"
     :target="isExternal(url) ? '_blank' : null"
+    :title="title"
     @click="segment && $trackClickEvent(segment)"
```
delapuente, line 9: Non-blocking: when this was a menu, we wanted all the menus to have the option of tracking the click, now this is more general, I think we could get rid of it and move the responsibility to the consumer.

components/ui/AppLink.vue
```diff
@@ -0,0 +1,31 @@
...
+import Vue from 'vue'
+import { Component, Prop } from 'vue-property-decorator'
+
+@Component
+export default class extends Vue {
+  @Prop(String) url
+  @Prop(Object) segment
+  @Prop({ type: Boolean, default: false }) isStatic
+
+  isExternal: boolean = this.url.startsWith('http')
```
delapuente, line 25: Be careful, the value of `this` inside the class scope is the class itself and you need to refer to the component instance. Convert these to getters.

components/events/EventCard.vue
```diff
@@ -1,8 +1,8 @@
 <template>
-  <a
-    :href="to"
+  <component
+    :is="to ? 'AppLink' : 'div'"
     class="card-link"
-    target="_blank"
+    v-bind="{ 'url': to }"
```
delapuente, line 5: Same as before:
```suggestion
    :url="to"
```

components/events/EventMenu.vue
```diff
@@ -1,7 +1,12 @@
 <template>
   <nav>
     <section class="ibm-research">
-      <a href="https://qiskit.org" class="text-logo" target="_blank" rel="noopener"><img class="logo" alt="Qiskit logo" src="https://qiskit.camp/img/qiskit-logo.png">&nbsp;Qiskit</a>
+      <AppLink
+        class="text-logo"
+        v-bind="{ url: '/'}"
```
delapuente, line 6: No need for dynamic binding here:
```suggestion
        url="/"
```

components/experiments/ExperimentHeader.vue
```diff
@@ -1,9 +1,9 @@
 <template>
   <div class="experiment-header-container">
     <div>
-      <nuxt-link
+      <AppLink
         class="experiment-header__back-navigation"
-        to="/experiments"
+        v-bind="{ url: '/experiments' }"
```
delapuente, line 6: Nor here.

components/experiments/ExperimentHeader.vue
```diff
@@ -14,23 +14,21 @@
...
           {{ authors }}
         </span>
         |
-        <a
+        <AppLink
           class="experiment-header__source-code-link"
-          :href="source"
-          target="_blank"
-          rel="noopener"
+          v-bind="{ url: source }"
```
delapuente, line 26: Use the property form, please.

components/events/EventCard.vue
```diff
@@ -1,8 +1,8 @@
 <template>
-  <a
-    :href="to"
+  <component
```
delapuente, line 2: An [`<a>` tag without the `href` is semantically correct](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) and means a placeholder for when a may be expected, as is the case with event cards:
> If the a element has no href attribute, then the element represents a placeholder for where a link might otherwise have been placed, if it had been relevant, consisting of just the element's contents.

 We just need to realize that if the `to` property is not provided we should go for the link tag without including any other link-related property (even `href`).

components/elements/ElementPresentation.vue
```diff
@@ -18,20 +18,24 @@
       <Cta
         v-if="to"
         class="element-presentation__get-started-cta"
-        :to="to"
-        @click="segmentAction && $trackClickEvent({
-          action: segmentAction
-        })"
+        :link="{
```
delapuente, line 21: When removing `link`, use the property form for the cases when you are writing the name of the properties explicitly, like this one:
```suggestion
        :url="to"
        :segment="{ action: segmentAction }"
```

components/elements/ElementPresentation.vue
```diff
@@ -18,20 +18,24 @@
...
         Get Started!
       </Cta>
       <Cta
         v-if="sources"
         class="element-presentation__github-cta"
-        :to="sources"
-        @click="segmentAction && $trackClickEvent({
-          action: segmentAction
-        })"
+        :link="{
```
delapuente, line 33: Or this one.

components/elements/ElementPresentation.vue
y4izus: I prefer using `v-bind`

components/experiments/ExperimentHeader.vue
```diff
@@ -14,23 +14,21 @@
...
-          target="_blank"
-          rel="noopener"
+          :url="source"
         >
           Explore the code
-        </a>
+        </AppLink>
       </p>
-      <Cta v-if="launch" :to="launch">
+      <Cta v-if="launch" :link="{ url: launch }">
```
delapuente, line 31: I think you forgot to change this since `link` property no longer exists.

constants/appLinks.ts
```diff
@@ -0,0 +1,170 @@
...
+
+const SOCIAL_MEDIA_COMMUNITY: Array<SocialMediaCommunityLink> = [
+  {
+    name: 'Slack',
+    url: 'http://ibm.co/joinqiskitslack',
+    title: 'Slack community',
+    segment: {
+      action: 'Qiskit Community: Slack'
+    },
+    iconPath: 'M9.879 10.995l1.035 3.085 3.205-1.074-1.035-3.074-3.205 1.08v-.017z',
```
delapuente, line 57: Please, provide the complete SVG.

Splitting the SVG in their components has two main disadvantages:
1. It is non-trivial to re-compose the SVG to figure out what is drawn.
2. It is hard to edit, or accommodate to a new format (in case an icon needs three paths instead of two).

I'm also in favour of referring to an external SVG and embedding with a plugin, in the spirit of [`vue-inline-svg`](https://www.npmjs.com/package/vue-inline-svg) or something better.

y4izus: Per offline conversations, we will leave this point for the moment because svg icons probably change to carbon design icons

components/experiments/ExperimentHeader.vue
y4izus: The ExperimentHeader component has been deleted

components/experiments/ExperimentHeader.vue
```diff
@@ -1,9 +1,9 @@
 <template>
   <div class="experiment-header-container">
     <div>
-      <nuxt-link
+      <AppLink
         class="experiment-header__back-navigation"
-        to="/experiments"
+        :url="'/experiments'"
```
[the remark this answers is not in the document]
y4izus: The ExperimentHeader component has been deleted

components/elements/ElementPresentation.vue
```diff
@@ -18,20 +18,24 @@
       <Cta
         v-if="to"
         class="element-presentation__get-started-cta"
-        :to="to"
-        @click="segmentAction && $trackClickEvent({
-          action: segmentAction
-        })"
+        v-bind="{
```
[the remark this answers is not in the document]
y4izus: I'm using `v-bind` in order to follow the same structure of the rest of the code

PR #641, opened 2020-06-04
Footer new design

### LINKED ISSUE #639: Develop new duo footer with updated copy and logos

y4izus:
Fixes #639


components/layouts/PageFooter/FooterIconsSection.vue
```diff
@@ -0,0 +1,50 @@
+<template>
```
delapuente, line 1: We could try to factor out the common markup and styles for these sections. What about creating three components: `FooterSection`, `FooterLinksSection` and `FooterIconsSection`.

`FooterSection` would use a slot to accept any content where now is the `AppLink` element.

`FooterLinksSection` would use that and add the links as children and `FooterIconsSeciton` would do the same but with icons, in the same way you're already doing here.

components/layouts/PageFooter/index.vue
```diff
@@ -16,31 +35,114 @@
...
+    }
+  }
+
+  &__sections {
+    display: grid;
+    grid-template-columns: repeat(6, 0.5fr);
+    margin: 0 $spacing-07;
+    row-gap: $spacing-07;
+
+    @include mq($until: medium) {
```
delapuente, line 99: Would you mind to write this `@include mq($until: medium)` after the next one? The rationale behind is going from _large_ to _small_ so you have the desktop code, then the modifications for medium, then small.

components/layouts/PageFooter/FooterIconsSection.vue
```diff
@@ -0,0 +1,50 @@
...
+  <div class="footer-icons-section">
```
delapuente, line 2: We can use a more semantic tag like `section` here.

components/layouts/PageFooter/index.vue
```diff
@@ -16,31 +35,114 @@
...
+      grid-template-columns: repeat(3, 0.5fr);
+      grid-template-rows: repeat(2, 0.5fr);
```
delapuente, line 101: Non-blocking: unless you want to enforce 2 rows, you can omit this.

components/layouts/PageFooter/index.vue
```diff
@@ -16,31 +35,114 @@
...
+  }
+
+  &__sections {
+    display: grid;
+    grid-template-columns: repeat(6, 0.5fr);
+    margin: 0 $spacing-07;
+    row-gap: $spacing-07;
+
+    @include mq($until: medium) {
+      grid-template-columns: repeat(3, 0.5fr);
```
delapuente, line 100: It is not very common to have a fractional `fr` and in this case, using `1fr` is semantically the same, could you change it and the other occurrences of this?

components/layouts/PageFooter/index.vue
```diff
@@ -16,31 +35,114 @@
...
+      grid-template-rows: repeat(2, 0.5fr);
+    }
+
+    @include mq($from:medium, $until: large) {
+      grid-template-columns: repeat(5, 0.5fr);
+      grid-template-rows: repeat(2, 0.5fr);
+
+      :nth-child(6) {
```
delapuente, line 108: I'd say it is preferable to use `last-child` here. There would be more scalable solutions but I think using `last-child` improves the intentionality.

components/layouts/PageFooter/FooterIconsSection.vue
y4izus: I unified both components in one called `FooterSection` that decided what to show based on a variable called `showIconLinks`

components/layouts/PageFooter/FooterIconsSection.vue
```diff
@@ -0,0 +1,50 @@
...
+    color: $cool-gray-60;
+    padding-bottom: $spacing-07;
+  }
+
+  &__icon-link {
+    fill: $cool-gray-60;
+    padding-right: $spacing-05;
+  }
+
+  &__icon-link:nth-child(3) {
```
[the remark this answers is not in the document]
y4izus: I used a `section` to do it

components/layouts/PageFooter/FooterSection.vue
```diff
@@ -1,41 +1,71 @@
...
+      {{ title }}
+    </h2>
+    <section :class="showIconLinks && 'footer-section__icons-group'">
+      <AppLink
+        v-for="element in elements"
+        :key="element.label"
+        :class="showIconLinks ? 'footer-section__icon-link' : 'footer-section__link'"
+        v-bind="element"
+      >
+        <component :is="element.icon" v-if="showIconLinks" />
```
delapuente, line 13: From the name of the property, I'd say this means "show the icons of the links _in addition_ to the labels" as if the default behaviour would be not showing them. Can we change this to something like `icons-only` to make it explicit we only want the icons? Then, as the default, we could render both if the icon is available.

PR #645, opened 2020-06-06
Acknowledge our support to Black lives matters movement

delapuente:
This PR introduces the always useful banner component motivated and inspired by the GitHub banner on the same topic. We need our copy, here it is a placeholder.


delapuente:
Let's use this copy and join voices together.

PR #655, opened 2020-06-15
Removes career section

### LINKED ISSUE #649: Move careers from its own section to under community
1. Move Overview section in place of where careers is today
2. Left align 2nd row in mobile design

y4izus:
Fix #649

### DIFF
--- a/constants/menuLinks.ts
+++ b/constants/menuLinks.ts
@@ -154,7 +154,7 @@ const TUTORIALS_LINK: NavLink = {
 
 const QISKIT_CAREERS: NavLink = {
   label: 'Careers',
-  url: ''
+  url: '#'
 }
 
 type FooterSection = {
--- a/constants/menuLinks.ts
+++ b/constants/menuLinks.ts
@@ -152,6 +152,11 @@ const TUTORIALS_LINK: NavLink = {
   }
 }
 
+const QISKIT_CAREERS: NavLink = {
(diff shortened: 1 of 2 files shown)


constants/menuLinks.ts
```diff
@@ -175,7 +175,7 @@ const FOOTER_ELEMENTS: FooterStructure = [
       QISKIT_ELEMENTS.ignis
     ]
   },
-  // Elements in the third footer column
+  // Elements in the second footer column
```
delapuente, line 178: Non-blocking: instead of a comment per entry, you can clarify this before the `const`: `// Each entry represents a column`.

y4izus:
I've created issue #663 to add "careers" link. Now we don't have a target URL, so we decided to remove until we have one

constants/menuLinks.ts
```diff
@@ -184,15 +183,14 @@ const FOOTER_ELEMENTS: FooterStructure = [
       QISKIT_VIDEOS
     ]
   },
-  // Elements in the fourth footer column
   {
     title: 'Community',
     elements: [
       COMMUNITY_SUB_LINKS.events,
       QISKIT_ADVOCATES.advocates
+      // QISKIT_CAREERS
```
delapuente, line 191: Non-blocking: Add a TODO referring to issue #663 like, "enable when #663 is done" or something like this.

PR #662, opened 2020-06-16
New menu implementation

### LINKED ISSUE #659: Implement new menu
The new menu design does not specify what happens with the second-level menu. The integration of the new menu is out of the scope of this task.

delapuente:
Fix #659

### DIFF
--- a/components/layouts/TheMenu/index.vue
+++ b/components/layouts/TheMenu/index.vue
@@ -130,12 +130,12 @@ export default class extends Mixins(MenuMixin) {}
       text-decoration: underline;
     }
 
-    &.nuxt-link-active {
-      color: $purple-70;
-    }
-
     &:last-child {
       margin-right: 0;
+    }
+
+    &_active {
+      color: $purple-70;
     }
   }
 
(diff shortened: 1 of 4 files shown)


components/layouts/TheMenu/index.vue
```diff
@@ -8,41 +8,38 @@
...
-            :key="link.url"
-            class="menu__link"
-            :class="{ 'menu__link_active': isActive(link) }"
-            v-bind="link"
+    <section class="menu__main-level">
+      <nav class="menu__navigation-level">
+        <AppLink
+          class="
+            menu__link
+            menu__link_img-only
```
y4izus, line 16: I'm trying to find the effect of this class `menu__link_img-only` and I didn't find it. I'm missing something or this class is unnecessary?

delapuente: In theory, it removed some extra space added by the line height of the link but that is completely nullified by the fact that links' display is flex in column mode. Should be removed.

delapuente:
I think to make the hover behaviour consistent with the footer is good. I'll do that.

PR #689, opened 2020-06-23
Hero description styles

### LINKED ISSUE #686: Styles of the hero moment description
Go to new redesign home page

Screen Shot 2020-06-23 at 11 53 03

- Color text should be $gray-60
- Block description width should be a little bit smaller

Screen Shot 2020-06-23 at 11 53 19

y4izus:
Fix #686

### DIFF
--- a/components/landing/TheHeroMoment.vue
+++ b/components/landing/TheHeroMoment.vue
@@ -91,7 +91,7 @@ export default class extends Vue {
   &__description {
     @include type-style('body-long-01');
     color: $gray-60;
-    max-width: 4.75 * $column-size-large;
+    max-width: 5 * $column-size-large;
     margin: 0 0 $layout-06;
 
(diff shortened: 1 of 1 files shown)


components/landing/TheHeroMoment.vue
```diff
@@ -90,16 +90,18 @@ export default class extends Vue {
 
   &__description {
     @include type-style('body-long-01');
-    color: $white-text-01;
+    color: $gray-60;
+    max-width: 4.75 * $column-size-large;
```
delapuente, line 94: Regarding `max-width`, I'd prefer if we use some multiple of `$column-size-large`. I'd say it is not 6 but can you confirm what's the correct value with @JRussellHuffman, please?

y4izus: Yes, he thinks it looks better with 5 columns, I will change it

PR #736, opened 2020-07-14
Add the cards for previewing the careful explanations when they appear in addition to other resources

delapuente:
Fix #730

This PR does, mainly, 3 things:
- It factors out the explanations, introducing a `compact` mode and a summary slot for displaying it alone when `compact` is present. That view becomes the card.
- It introduces URL-based filters so the different filter combinations have now a unique URL. For doing it:
  1. The `beforeRouteEnter` and `beforeRouteUpdate` functions parse the URL for the `learnLevel` and the `timeScale` parameters and set them in the store (both functions are needed to be aware of query parameters changes).
  2. Filters no longer change the store, they now navigate to a new URL (this navigation is virtual, using the programmatic version of `<nuxt-link>` which is `this.$router.push()`).
  3. Everything else remains the same.

Regarding the visibility of the explanations:
- The explanations are shown:
  - In compact more when accompanied by more resources.
  - Extended when alone.
  - Not displayed at all when showing **everything**.
- It removes the 1 minute/advance classification for one of the resources, these 1 minute/whatever combinations are restricted to the texts.

Although, it also includes these somehow related fixes:
- As part of #724, makes all the texts `cool-gray-80` to match Figma designs.
- It also distinguishes the new kind of text introduced by the learning page: the content paragraph. This type of paragraph has become the standard paragraph. Outstanding paragraphs are now marked by entity modifiers `_importance_outstanding` while the previous old standard paragraph, which acted as a support text for sections (in contras to content) is now `_importance_support`.

And this unrelated one, but it hurt me seeing it there:
- Fix a bug regarding repeated keys when generating the second-level links of the menu.

PR #739, opened 2020-07-15
Variation on 736 for summary URLs only (not filter URLs)

### LINKED ISSUE #730: Implement the carefully worded description cards in the learn page
Implementation: 
Provide 2 dedicated URLs for carefully worded beginners.

Make the short version to open a full version in a new window.

delapuente:
Fix #730

Same as #736 but with special URLs for the summaries only.


delapuente:
It won't work since URL generation does not take into account query parameters.

PR #802, opened 2020-07-30
Overview header

### LINKED ISSUE #786: Implement header

y4izus:
Fix #786

### DIFF
--- a/components/ui/PageHeader.vue
+++ b/components/ui/PageHeader.vue
@@ -1,0 +1,56 @@
+<template>
+  <header class="page-header">
+    <div class="page-header__container">
+      <h1 class="page-header__title">
+        <slot />
+      </h1>
+    </div>
+  </header>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+import { Component } from 'vue-property-decorator'
+
+@Component
+export default class extends Vue {}
+</script>
+
+<style lang="scss" scoped>
+@import '~/assets/scss/blocks/copy.scss';
+@import '~carbon-components/scss/globals/scss/typography';
+@include carbon--theme($carbon--theme--white);
+
+.page-header {
+  @include responsive-grid-bg-strip('/images/grid/grid-hero-learn.svg', auto, 28rem);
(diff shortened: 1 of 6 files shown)


components/ui/PageHeader.vue
```diff
@@ -0,0 +1,56 @@
...
+import { Component } from 'vue-property-decorator'
+
+@Component
+export default class extends Vue {}
+</script>
+
+<style lang="scss" scoped>
+@import '~/assets/scss/blocks/copy.scss';
+@import '~carbon-components/scss/globals/scss/typography';
+@include carbon--theme($carbon--theme--white);
```
delapuente, line 22: If I don't recall wrong, I think this is the first time I see adding the `carbon--theme` mixing in a component. Why are we doing this?

pages/learn.vue
```diff
@@ -1,16 +1,12 @@
...
-        <h1 class="the-learn-header__title">
-          Start your path towards<br>
-          learning
-          <span class="the-learn-header__relevant-word">
-            Qiskit
-          </span>
-        </h1>
-      </div>
-    </header>
+    <PageHeader>
```
delapuente, line 3: Thank you for factoring out this!

components/ui/PageHeader.vue
```diff
@@ -0,0 +1,56 @@
...
+      height: 28rem * 40 / 64;
+    }
+  }
+
+  &__title {
+    @include type-style('expressive-heading-05', true);
+    color: $white-text-01;
+  }
+
+  &__relevant-word {
```
delapuente, line 49: To preserve the separation of concerns, the header component should not style deep content. Also, we expect the "relevant word" to become a separated component where we implement the typewriter effect. I would prefer if we convert it in a component now (adding the typewriting effect in a follow-up PR) or if we convert it in its own BEM block, with its own identity.

components/ui/PageHeader.vue
y4izus: Per offline conversations, it's a passing experiment. I will remove it

PR #812, opened 2020-08-06
Add overview page

delapuente:
For preview purposes.


components/overview/ContentSection.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+  <article class="content-section">
+    <div class="content-section__overview">
+      <h2 class="copy__title">
+        {{ title }}
+      </h2>
+      <p class="content-section__description">
+        {{ description }}
+      </p>
+      <AppLink class="copy__link" :url="link.url">
```
delapuente, line 10: Let's use `AppCta` instead. It includes the external-link indicator functionality and it is more consistent with other CTAs across the site.

components/overview/ContentSection.vue
```diff
@@ -0,0 +1,77 @@
...
+  @Prop(String) description
+  @Prop(Object) link
+}
+</script>
+
+<style lang="scss" scoped>
+@import '~/assets/scss/blocks/copy.scss';
+
+.content-section {
+  width: 100%;
```
delapuente, line 39: `width` and `margin-bottom` are [external geometry according to BEM](https://en.bem.info/methodology/css/#external-geometry-and-positioning), can we move these to the consumer of the component?

components/overview/TableOfContents.vue
```diff
@@ -0,0 +1,99 @@
...
+<script lang="ts">
+import Vue from 'vue'
+import { Component } from 'vue-property-decorator'
+import AppLink from '~/components/ui/AppLink.vue'
+
+@Component({
+  components: { AppLink }
+})
+export default class extends Vue {
+  elements = [
```
delapuente, line 24: In order to reduce coupling between the presentational and content aspects of the component, let's pass the elements from the consumer. This way, the kind of relationships derived from the `id` of the sections and the `url`s are also in the same place, improving the cohesiveness of the content.

components/ui/PageHeader.vue
```diff
@@ -0,0 +1,47 @@
+<template>
+  <header class="page-header">
```
delapuente, line 2: Please, prefix this component with `The` to follow the [strong recommendation regarding single instance components](https://vuejs.org/v2/style-guide/#Single-instance-component-names-strongly-recommended).

pages/overview.vue
```diff
@@ -0,0 +1,163 @@
...
+
+.overview-page {
+  background-color: white;
+
+  &__content-container {
+    @include contained();
+    display: flex;
+    align-items: flex-start;
+    overflow: visible;
+    margin-bottom: $layout-07 / 2;
```
delapuente, line 132: Please, double-check with @JRussellHuffman if we can stick with Carbon layout spacing rules and convert this into a `layout-06` or `layout-05`.

pages/overview.vue
```diff
@@ -0,0 +1,163 @@
...
+
+    @include mq($until: medium) {
+      margin-bottom: 0;
+    }
+  }
+
+  &__main-nav-options {
+    width: 20%;
+    padding-right: $spacing-07;
+    position: -webkit-sticky; /* Safari */
```
delapuente, line 142: Maybe I'm wrong but I think there is no need for the vendor-prefixed version. It should be handled by the PostCSS `autoprefixer` plugin,  [enabled by default in Vue.js](https://cli.vuejs.org/guide/css.html#postcss).

components/overview/ContentSection.vue
y4izus: I didn't use `AppCta` because on the Figma design there is no external/internal-link. We should ask @JRussellHuffman before chage it. @JRussellHuffman , what do you think?

components/overview/ContentSection.vue
y4izus: Good point! We should review all the code because we are not doing it properly everywhere

components/overview/ContentSection.vue
delapuente: Solved during the sprint review, we move forward with changing to the `AppCta` for consistency.

pages/overview.vue
delapuente: During sprint review, we decided this would be `layout-05`.

PR #832, opened 2020-08-12
Research apps section

### LINKED ISSUE #800: Implement "Research Applications"

y4izus:
Fix #800 

I've also implemented to feedbacks from Russell:
- Change the accordion tab hover color to `$cool-gray-30`
- Accordion behavior: We should see always at least one tab open


components/overview/ContentAccordion.vue
```diff
@@ -27,10 +29,13 @@ import AccordionLayout from '~/components/overview/AccordionLayout.vue'
 export default class extends Vue {
   @Prop(Array) elements
 
-  expandedItem = 0
+  open = Array.from({ length: this.elements.length }, (_, i) => i === 0)
```
delapuente, line 32: May be `itemStates` or similar is better, something that reminds us this is the list of the state of each item. Annotating the property with the type `boolean[]` also helps figuring out what this is about.

components/overview/ContentAccordion.vue
```diff
@@ -27,10 +29,13 @@ import AccordionLayout from '~/components/overview/AccordionLayout.vue'
...
 
   actionChange (ev: any) : void {
```
delapuente, line 34: Non-blocking: the preferred abbreviation in HTML for the event is `evt`.

You can also annotate `ev` with type `{changedItem: number, state: boolean[]}`. You could even unpack both values here:
```suggestion
  actionChange ({ changedItem, state }: {changedItem: number, state: boolean[]}) : void {
```

components/overview/ContentAccordion.vue
```diff
@@ -27,10 +29,13 @@ import AccordionLayout from '~/components/overview/AccordionLayout.vue'
...
-    this.expandedItem = ev.changedIndex
+    const accordionComponent = this.$refs.acc as any
+
+    this.open = accordionComponent.state.map((_, index) => index === ev.changedIndex)
```
delapuente, line 37: Instead of using this state, use the one that comes with the event `ev.state`.

PR #883, opened 2020-09-01
Remove !important

### LINKED ISSUE #764: nuxt-lazy-load component requires using CSS `!important` for overriding the behaviour of `background-picture` in responsive design
Notice how in 01c540c3f2ce389664eb54b14d11816aa3b8991a and #763 we need to use `!important` to override the behaviour of the `lazy-background` directive. We need to come up with a better solution not involving `!important`. This refactor task should be prioritized.

y4izus:
Fix #764 

With the new design, the images don't change between screen sizes, so it's not necessary to use `!important` anymore


delapuente (approved):
Nice! Thank you!

PR #908, opened 2020-09-11
Swipe background effect

### LINKED ISSUE #609: Add animation of the CTA icon
Human Design presented an animation of the CTA where the arrow moved when hover. Could be a good addition to recover it from oblivion and implement it.

korgan00:
Fix #609 

Background in buttons now makes a swipe transition on hover.

### DIFF
--- a/components/landing/LandingCta.vue
+++ b/components/landing/LandingCta.vue
@@ -32,6 +32,10 @@ export default class extends Vue {
   align-items: center;
   padding: 0 $spacing-07;
 
+  background-size: 200% 100%;
+  background-position-x: 100%;
+  transition: background-position-x 0.3s ease-out, color 0.3s ease-out;
+
   @include mq($from: large) {
     width: 15rem;
   }
@@ -42,13 +46,13 @@ export default class extends Vue {
     width: 12rem;
   }
 
+  &:hover {
+    background-position-x: 0%;
+  }
+
   &_primary {
     color: $white;
-    background-color: $purple-70;
-
-    &:hover {
-      background-color: $purple-90;
-    }
+    background-image: linear-gradient(90deg, $purple-90 0%, $purple-90 50%, $purple-70 50%, $purple-70 100%);
 
     @include mq($from: large) {
       padding-top: $spacing-07;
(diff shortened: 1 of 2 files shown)


components/landing/LandingCta.vue
```diff
@@ -32,6 +32,10 @@ export default class extends Vue {
   align-items: center;
   padding: 0 $spacing-07;
 
+  background-size: 200% 100%;
```
techtolentino, line 35: this approach is very clean @korgan00 !

techtolentino (commented):
@korgan00 - this looks great. I pulled down the code and tested.

Question for you or @JRussellHuffman
- looks like we decided on Option 3 from Russell's figma yeah? (if so, I like it)
- should `active` states have this same animation effect?

Great work, I can approve once we hear from Russell.

🙏

components/landing/LandingCta.vue
korgan00: Thank you :)

delapuente:
Thank you @techtolentino for your review. Do not merge yet though, let's wait for @JRussellHuffman comments and another review.

delapuente (approved):
Thank you @korgan00, this looks nice to me.

JRussellHuffman:
It looks great.

For future reference, would it be easy to make it animate from right to left (instead of left to right)? I like it as is because the animation goes the same direction as the arrow, but just curious how flexible the component is.

delapuente:
Should be easy indeed, regarding the questions open by Randy, what's your take [on those](https://github.com/Qiskit/qiskit.org/pull/908#pullrequestreview-486757161)?

JRussellHuffman:
@techtolentino to answer your questions.

- yep, looks like Option 3 (which looks good to me)
- I don't think the active state should have this same animation effect. I would imagine the animation would only get about halfway done before the state is no longer active (assuming active = mouse down).

delapuente:
The active situation is a very good question, not for desktop, where it is mouse down indeed but thinking of mobile where you don't have mouse over, using active serves as feedback for pressing the button so the animation will play more or less depending on how long you press the button and that's provide a very, very good feeling. Let's give it a try!

components/landing/LandingCta.vue
```diff
@@ -42,13 +46,13 @@ export default class extends Vue {
     width: 12rem;
   }
 
+  &:hover {
```
delapuente, line 49: @korgan00 would you mind to give it a try and implement the same effect for the `active` state?

delapuente (commented):
Let's take a look at implementing this for the `active` state so we get some feedback in mobile, following [general discussion](https://github.com/Qiskit/qiskit.org/pull/908#issuecomment-692932242).

PR #913, opened 2020-09-14
Add type effect (#804)

### LINKED ISSUE #804: Implement typewritting effect on RelevantWord component
Following comment at https://github.com/Qiskit/qiskit.org/pull/802#discussion_r463436947, this task is about to rename the component and encapsulate a typewritting effect:

Typewriter effect

techtolentino:
Work in Progress

**Issue**
Resolves #804

**Description**
This PR adds a `TypedWord` component that will replace the `RelevantWord` component, on the Learn Page


components/ui/TypedWord.vue
```diff
@@ -0,0 +1,91 @@
+<template>
+  <div class="typed-text__container" :class="{ 'typed-text__cursor': showCursor }">
```
delapuente, line 2: We follow a convention in which the main container has a class named after the name of the component in kebab-case so please, could you name this `div` class `typed-word`?

components/ui/TypedWord.vue
```diff
@@ -0,0 +1,91 @@
...
+    <span class="typed-text__content">{{typeValue}}</span>
+  </div>
+</template>
+
+<script>
+export default {
```
delapuente, line 8: In qiskit.org we follow TypeScript class-based components as shown in [Nuxt.js documentation](https://typescript.nuxtjs.org/cookbook/components/#template) (see "Class API" tab). Could you rewrite this component to follow this style, please?

components/ui/TypedWord.vue
```diff
@@ -0,0 +1,91 @@
...
+      typeValue: "",
+      typeStatus: false,
+      typeArray: ["Qiskit", "Python", "Quantum"],
+      typingSpeed: 200,
+      erasingSpeed: 100,
+      newTextDelay: 3000,
+      eraseTextDelay: 5000,
+      typeArrayIndex: 0,
+      charIndex: 0,
+      showCursor: true,
```
delapuente, line 20: We should identify the public API of the component (those values that customize the behaviour of the component) and convert them into component properties so it is the consumer of the component who's in charge of customizing the behaviour to their needs.

The way it is right now implies a strong coupling with the learn page.

delapuente (changes requested):
This is looking nice @techtolentino. I would recommend starting with a word already written so it does not seem like broken on first load.

Please, review the [build logs](https://travis-ci.com/github/Qiskit/qiskit.org/builds/184232066) and solve the issues regarding the linter.

components/learn/TheLearnHeader.vue
```diff
@@ -1,16 +1,20 @@
 <template>
   <ThePageHeader class="the-learn-header">
-    Start your path towards<br>
-    learning <RelevantWord>Qiskit</RelevantWord>
+    Start your path towards
+    <br />learning
+    <TypedWord
+      :initialWord="'Qiskit'"
```
delapuente, line 6: Use kebab-case here, it is automatically converted into camelCase by Nuxt.js.

delapuente: Also, you don't need this to be dynamic so it is better to use:
```suggestion
      :initial-word="Qiskit"
```

components/ui/TypedWord.vue
```diff
@@ -0,0 +1,113 @@
...
+
+<style lang="scss" scoped>
+.typed-word {
+  &__content {
+    font-style: italic;
+    color: white;
+    padding: $spacing-01 $spacing-03 $spacing-01 0;
+    margin-left: $spacing-02;
+  }
+  &__container {
```
delapuente, line 90: Add blank lines between selectors.

components/ui/TypedWord.vue
```diff
@@ -0,0 +1,113 @@
...
+</template>
+
+<script>
+// Implementation based on https://github.com/webnoobcodes/vuejs-typeeffect
+import Vue from 'vue'
+import { Component, Prop } from 'vue-property-decorator'
+
+@Component
+export default class extends Vue {
+  @Prop({ type: String, default: '' }) initialWord
```
delapuente, line 14: Add TypeScript annotations such as:
```suggestion
  @Prop({ type: String, default: '' }) initialWord: string = ''
```

components/ui/TypedWord.vue
```diff
@@ -0,0 +1,113 @@
...
+  @Prop({ type: Array, default: [] }) typeArray
+  @Prop({ type: Number, default: 200 }) typingSpeed
+  @Prop({ type: Number, default: 100 }) erasingSpeed
+  @Prop({ type: Number, default: 3000 }) eraseTextDelay
+  @Prop({ type: Number, default: 3000 }) newTextDelay
+
+  // Initial data
+  value = ''
```
delapuente, line 22: Same here:
```suggestion
  value: string = ''
```
We try to leverage TypeScript type inference although it is considered a good practice to annotate declarations and signatures.

components/ui/TypedWord.vue
```diff
@@ -0,0 +1,113 @@
+<template>
+  <div class="typed-word__container" :class="{ 'typed-word__cursor': showCursor }">
```
delapuente, line 2: The top-level element should include a BEM block class named after the component so this element should include the class `typed-word`. I argue that there is no need for adding the container too although you can try to convince me. ;)

delapuente: Correct me if I'm wrong, please: the class `typed-word__cursor` is there for signalling the cursor is visible, is it right? If that's the case, consider converting it into a [boolean BEM block modifier](https://en.bem.info/methodology/quick-start/#boolean) like `typed-word_cursor-visible`.

components/ui/TypedWord.vue
```diff
@@ -0,0 +1,113 @@
...
+  }
+  &__container {
+    background-color: $purple-70;
+    display: inline-block;
+  }
+
+  &__cursor {
+    animation: 0.75s flash step-end infinite;
+
+    & .typed-word__content {
```
delapuente, line 98: You don't need to prepend `&` for making a descendant selector. It is enough with:
```suggestion
    .typed-word__content {
```

delapuente (changes requested):
Getting there! Thank you @techtolentino! Please, implement the changes I ask for here.

components/ui/TypedWord.vue
techtolentino: @delapuente - I'm running into another prop mutation issue, when adding the Ts annotation here.

Screen Shot 2020-09-17 at 7 19 16 AM


and the only place it's being referred to is here:

```js
created () {
  this.value = this.initialWord
  setTimeout(this.eraseText, this.newTextDelay + 200)
}
```

any thoughts?

techtolentino: (actually, I think I fixed it)

Issue #909, opened 2020-09-11
Animate links arrow


korgan00:
Per offline conversations, we talked about animating link arrows for both internal and external links.
I found a way to animate external links arrow only (without square), so we can animate both (internal and external).

delapuente:
Closed by #914

PR #914, opened 2020-09-14
Link arrow animation

korgan00:
#909 
Arrows in links now move slightly

### DIFF
--- a/components/landing/LandingCta.vue
+++ b/components/landing/LandingCta.vue
@@ -65,7 +65,8 @@ export default class extends Vue {
     color: $purple-70;
     background-image: linear-gradient(90deg, $purple-70 0%, $purple-70 50%, $cool-gray-10 50%, $cool-gray-10 100%);
 
-    &:hover {
+    &:hover,
+    &:active {
       color: $cool-gray-10;
     }
 
@@ -78,17 +79,21 @@ export default class extends Vue {
   &__icon {
     margin-left: $spacing-06;
     fill: currentColor;
+    overflow: visible;
   }
 
-  svg {
-      overflow: visible;
-  }
-  path:nth-child(1) {
+  $arrow-path: "path:nth-child(1)";
+
+  #{$arrow-path} {
     transform: translate(0, 0);
     transition: transform 0.2s ease-in-out;
   }
-  &:hover path:nth-child(1) {
-    transform: translate(3px, 0);
+
+  &:hover,
+  &:active {
+    #{$arrow-path} {
+      transform: translate(4px, 0);
+    }
   }
 }
 </style>
--- a/components/landing/LandingCta.vue
+++ b/components/landing/LandingCta.vue
@@ -79,5 +79,16 @@ export default class extends Vue {
     margin-left: $spacing-06;
     fill: currentColor;
   }
+
+  svg {
+      overflow: visible;
+  }
+  path:nth-child(1) {
+    transform: translate(0, 0);
+    transition: transform 0.2s ease-in-out;
+  }
+  &:hover path:nth-child(1) {
+    transform: translate(3px, 0);
+  }
 }
 </style>
--- a/components/ui/AppCta.vue
+++ b/components/ui/AppCta.vue
@@ -2,10 +2,19 @@
   <AppLink
     ref="link"
     class="app-cta copy__link"
+    :class="{
+      'app-cta_is-external': isExternal,
+      'app-cta_is-internal': !isExternal
+    }"
     v-bind="$attrs"
   >
-    <div><slot /></div>
-    <component :is="isExternal ? 'Launch16' : 'ArrowRight16'" />
+    <div class="app-cta__content">
+      <slot />
+    </div>
+    <component
+      :is="isExternal ? 'Launch16' : 'ArrowRight16'"
+      class="app-cta__icon"
+    />
   </AppLink>
 </template>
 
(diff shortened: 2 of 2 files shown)


components/landing/LandingCta.vue
```diff
@@ -75,5 +78,16 @@ export default class extends Vue {
...
+
+  svg {
+      overflow: visible;
+  }
+  path:nth-child(1) {
+    transform: translate(0, 0);
+    transition: transform 0.2s ease-in-out;
+  }
+  &:hover path:nth-child(1) {
+    transform: translate(3px, 0);
```
techtolentino, line 90: @korgan00 - I _think_ this arrow shift was going to be `8px`, if I remember correctly... 

@JRussellHuffman can verify.

techtolentino (commented):
Very nice work, @korgan00 

I think we still need to check with @JRussellHuffman to ensure we have all the micro-interactions that we needed.

Items where we could use Russell's guidance:
- [ ] the left-to-right shifting of the arrow on the Landing CTAs
- [ ] the diagonal-up-out interaction on the `LaunchOut` icons (Events page)
- [ ] the left-to-right shift on Arrow icon (Events page, last event)


Once we get Russell's feedback, we'll be good. From the code side, I like what you've done here, thank you!

components/landing/LandingCta.vue
```diff
@@ -75,5 +78,16 @@ export default class extends Vue {
...
   }
+
+  svg {
+      overflow: visible;
+  }
+  path:nth-child(1) {
+    transform: translate(0, 0);
+    transition: transform 0.2s ease-in-out;
+  }
+  &:hover path:nth-child(1) {
```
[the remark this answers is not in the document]
korgan00: Yes, both are applying.

You can check it on Preview link that the Vercel bot commented:  https://qiskitorg-git-fork-korgan00-feature-gi-i909-link-ar-b7dec5.qiskit.vercel.app

JRussellHuffman:
In general, it looks good. Its nice and subtle, but effective. I like it. I see some questions about pixel distance. Will answer those above.

components/landing/LandingCta.vue
JRussellHuffman: 8px may be a bit much compared to 3px. Can we do 4px?
Its tiny, but generally the IBM Design standards push for everything to be in intervals of 8 (which is where that first number came from). In this case, we can do 1/2 of 8, so 4px is fine.

For the "launch" icon (eg. on the learn page), is it translating the same distance?

components/landing/LandingCta.vue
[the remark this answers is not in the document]
JRussellHuffman: I think they make sense for `hover` and `focus`, but not active. Looks like we currently aren't doing anything for Active? Maybe we should make a separate issue to think about what should happen on `active`. It could just be a color change to purple 90.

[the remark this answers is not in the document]
JRussellHuffman: Actually, it looks like Salva has some thoughts on the active state over on https://github.com/Qiskit/qiskit.org/pull/908#issuecomment-692932242 regarding active for mobile. I trust his thoughts on this one.

components/ui/AppCta.vue
```diff
@@ -37,5 +37,26 @@ export default class extends Vue {
   & :first-child {
     margin-right: $layout-02;
   }
+  svg {
```
delapuente, line 40: Name the icon explicitly as a BEM Element of `app-cta`. Something like `app-cta__icon` should be enough.

components/ui/AppCta.vue
```diff
@@ -37,5 +37,26 @@ export default class extends Vue {
...
+      overflow: visible;
+  }
+  &[target=_blank] {
```
delapuente, line 43: Now, let's rely on the `isExternal` computed property to dynamically add BEM Block Modifiers depending on the property. Find the root of the template and use the object syntax of the `class` property to add modifiers depending of it is external or internal. For instance:

```
:class="{
  'app-cta_is-external': isExternal,
  'app-cta_is-internal': !isExternal
}"
```

delapuente: Once you have those modifiers, structure the SASS code in two blocks, one for when the block is internal, and other for when it is external:

```
.app-cta {
  ...
  &_is-internal {
  }

  &_is-external {
  }
}
```

components/ui/AppCta.vue
```diff
@@ -37,5 +37,26 @@ export default class extends Vue {
...
+    path:nth-child(2) {
+      transform: translate(0, 0);
+      transition: transform 0.3s ease-in-out;
+    }
+    &:hover path:nth-child(2) {
+      transform: translate(2px, -2px);
+    }
+  }
+  &:not([target=_blank]) {
```
delapuente, line 52: This will become then `&_is-internal`.

components/ui/AppCta.vue
```diff
@@ -37,5 +37,26 @@ export default class extends Vue {
   & :first-child {
     margin-right: $layout-02;
   }
+  svg {
+      overflow: visible;
+  }
+  &[target=_blank] {
+    path:nth-child(2) {
```
delapuente, line 44: Provide a meaningful SASS name to this selector, inside the corresponding modifier, such as:

```
&_is-external {
  $arrow-path: "path:nth-child(2)";
}
```

components/ui/AppCta.vue
```diff
@@ -37,5 +37,26 @@ export default class extends Vue {
...
+      transform: translate(0, 0);
+      transition: transform 0.3s ease-in-out;
+    }
+    &:hover path:nth-child(2) {
+      transform: translate(2px, -2px);
+    }
+  }
+  &:not([target=_blank]) {
+    path:nth-child(1) {
```
delapuente, line 53: Do the same with this one, you can reuse the SASS name:
```
&_is-internal {
  $arrow-path: "path:nth-child(1)";
}
```

components/ui/AppCta.vue
```diff
@@ -37,5 +37,26 @@ export default class extends Vue {
   & :first-child {
     margin-right: $layout-02;
   }
+  svg {
+      overflow: visible;
+  }
+  &[target=_blank] {
+    path:nth-child(2) {
+      transform: translate(0, 0);
```
delapuente, line 45: Finally, move the styles inside each modifier block so the final arrangement should look like:
```
&_is-external {
  $arrow-path: "path:nth-child(2)";

  #{$arrow-path} {
    ...
  }

  &:hover #{$arrow-path} {
    ...
  }
}

&_is-internal {
  $arrow-path: "path:nth-child(1)";

  #{$arrow-path} {
    ...
  }

  &:hover #{$arrow-path} {
    ...
  }
}
```
This way, we follow BEM naming conventions. And we make a [legit use of nested selectors](https://en.bem.info/methodology/css/#nested-selectors) for the code we have no control over it. The new structure has the advantage of being more regular (revealing a further change for factoring out more common parts) and keeps all the behaviour related to internal/external animation inside the internal/external block modifiers.

components/landing/LandingCta.vue
```diff
@@ -75,5 +78,16 @@ export default class extends Vue {
     margin-left: $spacing-06;
     fill: currentColor;
   }
+
+  svg {
```
delapuente, line 82: Use the same strategy here. Even better, consider addressing issue #916 first, in a separated PR, before continuing with this one, and you should be able to reuse everything in the other component for free.

delapuente (changes requested):
Thank you @korgan00, this is looking great. Let's apply some [BEM](https://en.bem.info/methodology/quick-start/#introduction) to the component, read my comments and ask for a new review when you finish.

Issue #916, opened 2020-09-15
Factor out common behaviour for LandingCta and AppCta components


delapuente:
Part of #973 

There is an opportunity for implementing the `LandingCta` component in terms of the `AppCta` component and reuse all the functionality in `AppCta` related to styles and animations.

delapuente:
As said in #917, it seems the behaviour is not exactly the same (the spacing between the text and the arrow changes for different sizes) and composition is not as trivial as it seems. Keeping this open and renaming to reflect the opportunity of factoring out some common behaviour.

korgan00:
I think we can consider this a duplicate of #973.
In fact, this is older, but #973 has a wider approach. We can move the discussion there.

y4izus:
I think this is done [here](https://github.com/Qiskit/qiskit.org/pull/1145). Do you agree @korgan00 @delapuente ?? if yes, I will close it

PR #917, opened 2020-09-17
Implemented landing CTA as an app CTA wrapper

korgan00:
#916 
Changed LandingCTA to keep functionality and reuse AppCTA.


delapuente:
Unfortunately, it seems the implementations are not compatible, nor even suitable for composition and a better rethink in not repeating ourselves is needed. Closing.

PR #949, opened 2020-09-22
Implement events header

### LINKED ISSUE #933: Events page update - Implement Header

y4izus:
Fix #933 

This PR also set the size for the title in the header instead of using `<br>` to calculate it

Note: Waiting for the list of words for the typewriting effect but the code review could start

### DIFF
--- a/components/events/TheEventsHeader.vue
+++ b/components/events/TheEventsHeader.vue
@@ -1,6 +1,6 @@
 <template>
   <ThePageHeader class="the-learn-header">
-    Join events from the worlds
+    Join events from the world's
     <TypewriterEffect
       :values="[
         'largest',
--- a/components/events/TheEventsHeader.vue
+++ b/components/events/TheEventsHeader.vue
@@ -1,0 +1,21 @@
+<template>
+  <ThePageHeader class="the-learn-header">
+    Join events from the worlds
+    <TypewriterEffect
+      :values="[
+        'largest',
+      ]"
+    />
+    quantum computing community
+  </ThePageHeader>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+import { Component } from 'vue-property-decorator'
+import ThePageHeader from '~/components/ui/ThePageHeader.vue'
+import TypewriterEffect from '~/components/ui/TypewriterEffect.vue'
+
+@Component({ components: { ThePageHeader, TypewriterEffect } })
+export default class extends Vue { }
+</script>
--- a/components/learn/TheLearnHeader.vue
+++ b/components/learn/TheLearnHeader.vue
@@ -1,7 +1,6 @@
 <template>
   <ThePageHeader class="the-learn-header">
-    Start your path towards
-    <br>learning
+    Start your path towards learning
     <TypewriterEffect
       :values="[
         'Qiskit',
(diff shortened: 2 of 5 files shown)


components/events/TheEventsHeader.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+  <ThePageHeader class="the-learn-header">
+    Join events from the worlds
```
techtolentino, line 3: ooh, I also think this should be `world‘s`, yeah?

techtolentino (commented):
Everything here looks good to me, @y4izus 

My only comment was about the updates to the Header styling, if we should wait on that... 

Great work!

pages/events/index.vue
```diff
@@ -1,35 +1,13 @@
 <template>
   <div class="event-page">
-    <header class="header-video">
+    <TheEventsHeader />
+    <div class="wrapper">
```
delapuente, line 4: A good moment for changing this turning this class name into BEM. ;)

delapuente (approved):
Looks great to me too. Thanks, @y4izus.

components/events/TheEventsHeader.vue
y4izus: I don't know, I just take the text from the Figma file 😅

components/learn/TheLearnHeader.vue
```diff
@@ -1,7 +1,6 @@
 <template>
   <ThePageHeader class="the-learn-header">
-    Start your path towards
-    <br>learning
+    Start your path towards learning
```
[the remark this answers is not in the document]
y4izus: I changed it because for the events page header the `<br>` is not always after the same word, so I think having the size with columns is a good solution. Moreover, I think we can still do the issue you're mentioning. What do you think?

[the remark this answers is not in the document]
techtolentino: @y4izus - I see what you mean. 

Yeah the typewriter effect, used on other pages, is doing some things to the surrounding text, that _maybe_ we didn't anticipate. 

I'm good with allowing this to go in.

Out of curiosity - where'd you get the calculation for the [max-width](https://github.com/Qiskit/qiskit.org/pull/949/files#diff-b40c93ca21d967330c592b9d48faabf7R45) from?

components/events/TheEventsHeader.vue
techtolentino: (checked with Russell, and it should read `world‘s` :pray:

components/learn/TheLearnHeader.vue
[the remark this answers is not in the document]
y4izus: As you already know, we are working with the Carbon Design System. It has a [2x Grid](https://www.carbondesignsystem.com/guidelines/2x-grid/overview/) that is the geometric foundation of all the visual elements of IBM Design. If you take a look at the Figma file, you can see these two grids:

Screen Shot 2020-09-23 at 09 55 44

We count the columns for each size and multiply them for the column's size on this screen size. For example, in the image above, you can count eight columns (white column + red column = 1) for the title, so:
`max-width: 8 * $column-size-large;`

BUT sometimes, some elements do not fit the same way on "carbon design figma" than "carbon design web" 🤷‍♀️. That's why we use a [look & feel](https://ishadeed.com/article/pixel-perfection/) approach, i.e., if you count and code 8 columns, but "your look & feel" is better with 10, go for the 10 ;)

PR #953, opened 2020-09-23
Update empty state for events list

### LINKED ISSUE #945: Create empty state card
**Description**
On the Events page, the design calls for having an empty state card. This card should appear when there are no events that match the filtered criteria.

techtolentino:
Resolves #945 

This pull request makes small adjustments to the Events template, and adds an EmptyCard component, so that we can provide an interim empty state. This work is moving in the direction of the card refactor in general.

Changes proposed:
- update Events page to show an EmptyCard when there are no results
- pass title and description props that only show if there are no results

(work in progress)


y4izus (changes requested):
Although this would be great to have for the dark theme, now we are implementing the light theme, so you should follow the Figma design file, i.e., instead of doing this one:

Screen Shot 2020-09-25 at 16 00 37

you should implement this one:

Screen Shot 2020-09-25 at 16 01 15

You have the design in the Figma file, if you don't find it, let me know :)

components/events/EmptyCard.vue
```diff
@@ -0,0 +1,71 @@
...
+
+    @include mq($until: medium) {
+      height: carbon--mini-units(22);
+      width: auto;
+    }
+  }
+
+  &__content {
+    flex: 1;
+    margin: $spacing-05;
```
y4izus, line 66: If you see content margins, not all are the same, the left margin is bigger than the top one:

Screen Shot 2020-09-28 at 19 03 50

components/events/EmptyCard.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+  <article class="empty-card">
+    <img
+      class="empty-card__image"
+      src="data:image/svg+xml,%3Csvg width='224' height='200' viewBox='0 0 224 200' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Crect width='224' height='200' fill='%23C1C7CD'/%3E%3Cpath d='M112 111.346V81.7308M156 138H68L112 61L156 138Z' stroke='white' stroke-width='2' stroke-linejoin='round'/%3E%3Cpath d='M112 126.154C113.62 126.154 114.933 124.828 114.933 123.192C114.933 121.557 113.62 120.231 112 120.231C110.38 120.231 109.067 121.557 109.067 123.192C109.067 124.828 110.38 126.154 112 126.154Z' fill='white'/%3E%3C/svg%3E%0A"
+      alt=""
+    >
+    <div class="empty-card__content">
+      <h3 class="copy__subtitle">
```
y4izus, line 9: You're using `copy_subtitle` that change from `'productive-heading-02'` on large screens to `'productive-heading-01'` on medium and small screens, but on the Figma design the typography doesn't change from one size to another

components/events/EmptyCard.vue
```diff
@@ -0,0 +1,71 @@
...
+        {{ title }}
+      </h3>
+      <p class="copy__paragraph copy__paragraph_importance_support">
```
y4izus, line 12: Just curious, why are you using `copy__paragraph_importance_support`?

components/events/EmptyCard.vue
```diff
@@ -0,0 +1,71 @@
...
+      </p>
+    </div>
+  </article>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+import { Component, Prop } from 'vue-property-decorator'
+
+@Component({})
```
y4izus, line 23: If the component decorator is empty, you can write `@Component`

components/events/EmptyCard.vue
```diff
@@ -0,0 +1,71 @@
...
+
```
y4izus, line 24: Please, remove this space, it's easy to read without it :)

components/events/EmptyCard.vue
```diff
@@ -0,0 +1,71 @@
...
+  @Prop(String) title
+  @Prop(String) description
+}
+</script>
+
+<style lang="scss" scoped>
+@import '~carbon-components/scss/globals/scss/typography';
+
+.empty-card {
+  height: carbon--mini-units(26);
```
y4izus, line 35: First time I see this!! we could open a discussion to see if we use this or rems as we are already doing. Just to have consistency on the code :)

y4izus (changes requested):
Thank Randy! I have some comments, could you take a look at them? :)

components/events/EmptyCard.vue
techtolentino: I'm totally fine with using rems values... I carried this over from previous habit of using everything Carbon :grim:

components/events/EmptyCard.vue
techtolentino: (bad copy paste from another paragraph, eslewhere)

components/events/EmptyCard.vue
y4izus: no no! it's a great idea to use as much as possible from Carbon, but we need to be all aligned to have a consistent code :) (and of course, a refactor ;) )

components/events/EmptyCard.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+  <article class="empty-card">
+    <img
+      class="empty-card__image"
+      src="data:image/svg+xml,%3Csvg width='224' height='200' viewBox='0 0 224 200' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Crect width='224' height='200' fill='%23C1C7CD'/%3E%3Cpath d='M112 111.346V81.7308M156 138H68L112 61L156 138Z' stroke='white' stroke-width='2' stroke-linejoin='round'/%3E%3Cpath d='M112 126.154C113.62 126.154 114.933 124.828 114.933 123.192C114.933 121.557 113.62 120.231 112 120.231C110.38 120.231 109.067 121.557 109.067 123.192C109.067 124.828 110.38 126.154 112 126.154Z' fill='white'/%3E%3C/svg%3E%0A"
```
y4izus, line 5: Could you add this img as a file and follow the structure we already have on the `LearningResourceCard` component?

y4izus (changes requested):
Thanks for the changes! Just one more :)

components/events/EmptyCard.vue
techtolentino: @y4izus - please have a look at the updates.

I added the img (as an svg to keep the clean lines). I was thinking that `image` doesn't necessarily have to be a prop, since it looks like this will be the image for all empty cards?

Thanks!

y4izus (approved):
Thank you Randy!! I will merge it :D

components/events/EmptyCard.vue
y4izus: we are going to reuse the card component, so the img is going to change :)

components/events/EmptyCard.vue
delapuente: Can you link the follow-up, please?

PR #954, opened 2020-09-25
Implement event card

### LINKED ISSUE #942: Implement event card

y4izus:
Fixes #942 

Note: I've used `width: calc(100% - #{$img-medium-width});` to size the `content` block because I didn't find a better way, but if you do, please, let me know :)

### DIFF
--- a/components/events/EventCard.vue
+++ b/components/events/EventCard.vue
@@ -10,13 +10,11 @@
         alt=""
       >
       <div class="event-card__content">
-        <header>
-          <span class="event-card__subtitle">
-            {{ type }}
-          </span>
+        <header class="event-card__header">
           <h3 class="event-card__title">
             {{ title }}
           </h3>
+          <cv-tag :label="type" kind="purple" />
         </header>
 
         <footer>
@@ -102,13 +100,24 @@ export default class extends Vue {
     justify-content: space-between;
   }
 
+  &__header {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+
+    @include mq($until: large) {
+      flex-direction: column;
+    }
+
+    .bx--tag--purple {
+      background-color: $purple-70;
+      color: $white;
+    }
+  }
+
   &__title {
     @include type-style('productive-heading-03');
     margin-top: 0.31rem;
-  }
-
-  &__subtitle {
-    @include type-style('productive-heading-02');
   }
 
   &__location, &__date {
--- a/components/events/EventCard.vue
+++ b/components/events/EventCard.vue
@@ -82,8 +82,9 @@ $img-medium-width: 13rem;
 
   &__image {
     flex: 0 0 14rem;
+    background-color: $cool-gray-80;
     background-repeat: no-repeat;
-    background-size: cover;
+    background-size: 100%;
     background-position: center;
     overflow: hidden;
 
--- a/components/events/EventCard.vue
+++ b/components/events/EventCard.vue
@@ -1,11 +1,9 @@
 <template>
   <article class="event-card">
-    <img
+    <div
       class="event-card__image"
-      alt=""
-      src="data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw=="
       :lazy-background="image"
-    >
+    />
     <div class="event-card__content">
       <header class="event-card__header">
         <h3 class="event-card__title">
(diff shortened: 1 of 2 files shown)


components/events/EventCard.vue
```diff
@@ -65,96 +65,108 @@ export default class extends Vue {
...
-  }
-
-  &__location, &__date {
-    @include type-style('body-short-01');
-  }
+  &__image {
+    flex: 0 0 14rem;
+    background-repeat: no-repeat;
+    background-size: cover;
+    background-position: center;
```
techtolentino, line 87: I'm wondering if we leave background-position _not_ center... and I only ask because it seems like for the Event images, sometimes they have the text top-left aligned...

Screen Shot 2020-09-25 at 12 01 56 PM


If we removed this center value, the images render better it seems
Screen Shot 2020-09-25 at 12 01 19 PM



What do you think? Or are these images that come with these events too unpredictable?

components/events/EventCard.vue
```diff
@@ -1,57 +1,57 @@
...
-import AppLink from '~/components/ui/AppLink.vue'
+import AppCta from '~/components/ui/AppCta.vue'
 
 @Component({
-  components: { AppLink }
+  components: { AppCta }
 })
 export default class extends Vue {
-  @Prop(String) type
+  @Prop(Array) types
```
techtolentino, line 54: this is great, making room for types! 👏

components/events/EventCard.vue
```diff
@@ -65,96 +65,108 @@ export default class extends Vue {
...
-    width: 1.25rem;
-    height: 1.25rem;
+  &__content {
+    @include type-style('body-long-01');
+    padding: $spacing-05 $spacing-05 $spacing-05 $spacing-07;
 
-    &--purple {
-      color: $icon-01;
+    @include mq($from: medium, $until: large) {
+      width: calc(100% - #{$img-medium-width});
```
techtolentino, line 105: I'm good with this since it's derived from a calculation that we can understand...

techtolentino (changes requested):
This looks awesome, @y4izus 

Just left 1 comment regarding the background-position

components/events/EventCard.vue
y4izus: Good catch! but I think we have to include @JRussellHuffman on the conversation :). Look, I did what you propose with the images I have:

Screen Shot 2020-09-28 at 18 51 51

We cut the right part of the img...  

The "problem" I see is that now we have an img almost square compared with the previous one:

- Previous one
Screen Shot 2020-09-28 at 19 12 50

- New one
Screen Shot 2020-09-28 at 19 13 34

So.. to be able to see all the img I propose 3 options:
- 1 => make the img wider
Screen Shot 2020-09-28 at 19 14 57

-2 => center the img to see 100% of the width
Screen Shot 2020-09-28 at 19 16 13

-3 => show the img to cover 100% width and 100% height (I don't like this one)
Screen Shot 2020-09-28 at 19 17 02

@JRussellHuffman , @techtolentino if you have any other option, let me know. @JRussellHuffman what should I chose?

JRussellHuffman: When someone uploads an event picture, are they given particular constraints, or can it be any size? Option 1 would make sense if the image is always the same size. Option 2 makes the most sense for flexibility, but now it looks like there is a margin on the image on 3 sides, which is less than ideal. could we do a background color behind the image in these scenarios? I'll mock something up.

Agreed, option 3 isn't the way we should go.

JRussellHuffman: Could we add border above and below on option 2? 
borders

JRussellHuffman: Its still less than idea though. In the future, would it be possible to do any of the following when they first upload the image?

- give a size recommendation
- force a particular size when they are submitting an event
- give them a preview so they can see if it looks cropped
- allow them to crop the image to fit our ratio
 
Or does all that happen on the airtable side?

y4izus: All that happen on the airtable side... maybe we can talk with the event team to find an img size that fits for all of us

techtolentino: I agree, this is a result of having varying dimensions when images are uploaded. From a branding and asset standpoint, it is best to _try_ and ensure we can have the same asset treatment, all around, with these images.

I understand that the Airtable is owned by another team, still I agree that if we can help other teams know which are the best dimensions, maybe that can become an adopted practice?

y4izus: Until we talk with the events team, I will implement the second option with the background-color as @JRussellHuffman suggest to be able to merge and not block, are you both agree?

techtolentino: sounds good!

techtolentino (approved):
Thanks @y4izus 

this matches what we agreed on in this thread...

PR #957, opened 2020-09-28
Fix mobile filters to light theme

### LINKED ISSUE #955: Convert filters to ligth theme for small screen on events page
We didn't catch this when we did the review of issue #934 

Go to https://qiskitorg-git-feature-update-events-page.qiskit.vercel.app/events and resize the screen to small size

Filters look like this:

Screen Shot 2020-09-25 at 18 57 28

Filters should look like this:

Screen Shot 2020-09-25 at 18 58 07

techtolentino:
Resolves #955

Updated the colors on the dropdown tabs for mobile.

Note: in order to not recreate another v-deep selector, I had to remove the scoped attribute in the component's style tag, which I think should be fine since this is an event-page.

### DIFF
--- a/pages/events/index.vue
+++ b/pages/events/index.vue
@@ -196,50 +196,54 @@ export default class extends QiskitPage {
 
   &__filters-time {
     margin-top: $layout-03;
-    & .bx--tabs__nav-link {
+    .bx--tabs__nav-link {
       color: $black-100;
       border-bottom-color: $gray-20;
     }
 
-    & .bx--tabs__nav-item:not(.bx--tabs__nav-item--disabled) .bx--tabs__nav-link,
+    .bx--tabs__nav-item:not(.bx--tabs__nav-item--disabled) .bx--tabs__nav-link,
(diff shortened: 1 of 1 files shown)


y4izus (changes requested):
Thanks Randy! I'm going to start the review with the visual elements:

1.- This is very subtle: text and arrow should have the same color
- Yours (Arrow = #343a3f, text = #161616)

Screen Shot 2020-09-29 at 15 35 32

- Figma (both = gray-100)

Screen Shot 2020-09-29 at 15 37 50

2.- Background color is different from Figma when the dropdown is opened. Also, on Fogma, texts have the same color:
- Yours
Screen Shot 2020-09-29 at 15 54 27

- Figma
Screen Shot 2020-09-29 at 15 55 27

I think we need @JRussellHuffman again :). @JRussellHuffman are Figma designs updated? are you agree with my comments?

JRussellHuffman:
Yes, designs are updated. And personally I like the idea of having the active drop down being the cool-gray-10 instead of white, if it isn't too much reinvention or too custom. But open to doing something different.

y4izus (approved):
Thank you, Randy!

Issue #959, opened 2020-10-01
Update Events page header to use apostrophe


techtolentino:
For proper typography, we need to replace `'` with `‘`, in the Events page title.

![Screen Shot 2020-10-01 at 6 59 11 AM (2)](https://user-images.githubusercontent.com/6276074/94812627-369a2380-03b4-11eb-8289-3f891381b34c.png)

PR #960, opened 2020-10-01
Replace single quote with apostrophe

techtolentino:
Small fix for #959


y4izus (approved):
Good catch Randy!! Thank you!

PR #961, opened 2020-10-01
Extract card component and use it in Learn and Event pages

### LINKED ISSUE #943: Extract card component and use it on learn and events pages

y4izus:
Fixes #943 and #963


pages/events/index.vue
```diff
@@ -62,7 +62,13 @@
           />
         </div>
         <div v-else class="event-page__results">
-          <EmptyCard :title="emptyCardTitle" :description="emptyCardDescription" />
+          <AppBasicCard
+            class="event-page__no-events-card"
+            :image="emptyCard.img"
+            :title="emptyCard.title"
+          >
+            {{ emptyCard.description }}
```
techtolentino, line 70: For this description (no events description), the design in Figma looks like it shouldn't go across the full width of the card - my initial guess would be a max-width of 26rem (calculated off design in Figma)

**Figma**
Screen Shot 2020-10-01 at 8 10 07 AM

**Preview in Vercel**
Screen Shot 2020-10-01 at 8 16 00 AM


I also wish we could remove the darker gray borders above/below the no-events image. I'm fine if we save that for later since we're still working to improve event images altogether, with the Events team.

Is there a way for the the `__no-events-card` to just be that solid `$cool-gray-30`?

techtolentino (changes requested):
This is awesome, @y4izus! 

I left a few comments.

I also had a couple of questions:

1) Should [this component](https://github.com/Qiskit/qiskit.org/pull/961/files#diff-f5bccdcfec2d036fb031c17293a33db1R53) have changed from EventCard to AppBasicCard? Or is that going to happen in a different refactor?

2) On Event card links, can we move the link down another rem? from 2 rem to 3 rem?
Screen Shot 2020-10-01 at 8 38 01 AM


Thank you!

components/ui/AppBasicCard.vue
```diff
@@ -1,17 +1,17 @@
 <template>
-  <article class="learning-resource-card">
-    <img
-      class="learning-resource-card__image"
-      alt=""
-      src="data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw=="
+  <article class="app-card">
+    <div
+      class="app-card__image"
```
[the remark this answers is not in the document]
y4izus: The image is the background of a `div` not of an `img`, and `div` tags don't have `alt` attribute, that's why I remove it. Anyways, we could try one of [these solutions](https://stackoverflow.com/questions/38380115/is-there-an-equivalent-to-the-alt-attribute-for-div-elements), what do you think?

pages/events/index.vue
y4izus: After [this conversation](https://github.com/Qiskit/qiskit.org/pull/954#discussion_r495149641) we decided to have these darker gray borders when the img don't fit all the space, I think we should have a consistent design for all cards. Aren't you agree?

y4izus:
Answering your questions:

1.- On this refactor, I have extracted the basic card component (card without tags) and use it in Learn page and for Empty Card. EventCard still remains because the card with tags is only used in one place. If you want, I can create a card for the basic layout (image and content), another for the basic card (title, description and link) based on the basic layout and another for the complex layout (card with tags) based also in the basic layout. But this is not what the issue said, so we need to create another one. What do you think?

2.- This is related to the event card, not with this PR. Would you mind to create another issue?

If you have any other question, just let me know!! :)

pages/events/index.vue
techtolentino: That's true... we did have that conversation. I'm okay to let this in and fix later if needed.

components/ui/AppBasicCard.vue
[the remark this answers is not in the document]
techtolentino: You're right, this is a `div` and not an `img`... 

I take that back now, I think since these are purely _decorative_ images, and since it's not also an image element, we don't need that here.

pages/events/index.vue
techtolentino: Agreed on the consistency.

techtolentino (approved):
Thanks, @y4izus!

pages/events/index.vue
delapuente: The dark background act as bands for filling images that do not accommodate to the container because we have no control on them. Fortunately, we can control everything regarding the empty card so I think these bands should not exist.

@JRussellHuffman what do you think?

delapuente (changes requested):
Nice refactor team! Please, rename `AppBasicCard.vue` with `AppCard.vue` and make it match the root component class. I prefer this over making the root class to be `app-basic-card` since it is simpler and already conveys the meaning of bare/basic.

pages/events/index.vue
techtolentino: This has been resolved with the latest commits, thank you Yaiza.

static/images/events/no-events.svg
```diff
@@ -1,5 +1,5 @@
-<svg width="224" height="200" viewBox="0 0 224 200" fill="none" xmlns="http://www.w3.org/2000/svg">
-<rect width="224" height="200" fill="#C1C7CD"/>
+<svg width="224" height="224" viewBox="0 0 224 224" fill="none" xmlns="http://www.w3.org/2000/svg">
```
techtolentino, line 1: this works nicely, thank you...

techtolentino (approved):
This looks good to me.

PR #962, opened 2020-10-01
Update Events page header with list of words

### LINKED ISSUE #952: Provide the list of words for the events page header
The typewriter effect shows several words, but now we only have one for the events page header. The complete list of words is needed

techtolentino:
Resolves #952 

Adding list of words.


y4izus (approved):
Great!! thank you Randy!

Issue #489, opened 2020-04-07
Meta: Apply BEM to components


delapuente:
From #271:

> Let's try BEM without removing scoped when possible and being flexible with the inner hierarchies, trying not to over-componentize.

y4izus:
There are several name conventions to write BEM. We should be aligned with one of them in order to be consistent everywhere in the code.
[Here](https://en.bem.info/methodology/naming-convention/) you can read more about that. 
For me, using the classic convention could be a good choice. What do you think @delapuente ?

delapuente:
I think it is a good starting point although I prefer the double hyphen style we are already following (in the cases we are using BEM). In my opinion, it makes it more clear the separation between modifiers and elements.

Is there any reason we should avoid the double hyphen form, @y4izus ?

y4izus:
[Here](https://en.bem.info/methodology/naming-convention/#two-dashes-style) we can read:
> Important: A double hyphen inside a comment (--) may cause an error during validation of an HTML document.

I know the probability to have this is small, but I prefer to avoid this "may cause an error" having (_) instead of (--)

lerongil:
Influenced by issue #473 and to be split up into a smaller issue for this sprint.

delapuente:
Right, although only in the cases when it can fail are those where the class contains [characters such as `!`, `>` or `<`](https://html.spec.whatwg.org/multipage/syntax.html#comments) which is very, very unlikely. Considering this, I'll be fine with whatever you decide.

y4izus:
Issues involved:
- [ ] Apply BEM to CompactFeature component #542 
- [ ] Apply BEM to AdvocateCard component #543
- [ ] Apply BEM to LegacyPresentation component #544
- [ ] Apply BEM to ExperimentCard component #545
- [ ] Apply BEM to ExperimentCarousel component #546
- [ ] Apply BEM to ExperimentHeader component #547
- [ ] Apply BEM to QiskitOrgMenu component #548
- [ ] Apply BEM to LinkList component #549
- [ ] Apply BEM to LegacySection component #550
- [ ] Apply BEM to PageSection component #551
- [ ] Apply BEM to Media component #552
- [ ] Apply BEM to SoftwareStack component #553
- [ ] Apply BEM to SyntaxHighlight component #554

delapuente:
Thank you a lot for taking the time of splitting this issue.

PR #985, opened 2020-10-07
Layout for dark page

### LINKED ISSUE #976: Implement the layout of the dark page.

y4izus:
Fix #976

### DIFF
--- a/components/layouts/ThePageFooter/FooterSection.vue
+++ b/components/layouts/ThePageFooter/FooterSection.vue
@@ -1,13 +1,21 @@
 <template>
   <section class="footer-section">
-    <h2 class="footer-section__title">
+    <h2
+      class="footer-section__title"
+      :class="{'footer-section__title_dark-mode': darkMode }"
+    >
       {{ title }}
     </h2>
     <nav :class="{ 'footer-section__icons-group': iconsOnly }">
       <AppLink
         v-for="element in elements"
         :key="element.url"
-        :class="iconsOnly ? 'footer-section__icon-link' : 'footer-section__link'"
+        :class="{
+          'footer-section__icon-link': iconsOnly,
+          'footer-section__icon-link_dark-mode': darkMode,
+          'footer-section__link': !iconsOnly,
+          'footer-section__link_dark-mode': darkMode
+        }"
         v-bind="element"
       >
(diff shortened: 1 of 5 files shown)


components/layouts/ThePageFooter/FooterSection.vue
```diff
@@ -1,13 +1,21 @@
 <template>
   <section class="footer-section">
-    <h2 class="footer-section__title">
+    <h2
+      class="footer-section__title"
+      :class="{'footer-section__title_dark-mode': darkMode }"
```
techtolentino, line 5: I think we can just use `footer-section__title--dark`, since it indicates the modifier anyway.

techtolentino, line 5: can we combine these class attributes to be combined into one attribute?

techtolentino (changes requested):
This is great @y4izus - just a couple of small comments.

Thanks

components/layouts/ThePageFooter/FooterSection.vue
y4izus: We [decided](https://github.com/Qiskit/qiskit.org/issues/489) some months ago to use `_modifier` instead of `--modifier`, so we have the `_modifier` along all the code. We should continue doing the same to have a consistent code.

Also, only having `dark` I think is not appropiate because, for example, in this case, it makes me think that the `title` is dark, but the title is white on dark-mode. Do you know what I mean? What do you think? We can look for another name if you don't like `_dark-mode`, but I think it explains what it does

korgan00 (approved):
As per offline conversations, the link on TheHeroMoment should be improved, but it is just a placeholder.

components/layouts/ThePageFooter/FooterSection.vue
techtolentino: sounds good... I thought I saw a `--purple` modifier at once, but I realize that might be from an override.

techtolentino: I thought there was another way of having the conditional in one line... but if this is the convention we're going with, that's good for me to know.

techtolentino (approved):
Thanks @y4izus !

PR #1000, opened 2020-10-09
Fix mobile "keep in touch" section

### LINKED ISSUE #999: Fix mobile "keep in touch" section
The content is breaking out in mobile view, this needs to be fixed.

techtolentino:
Fixes #999


y4izus (approved):
Good catch! Thank you!

Issue #660, opened 2020-06-16
Decide the nature of the menu and footer components and extend it when required


delapuente:
Decide if the current `AppFooter` and `TheMenu` are [Single Instance components](https://vuejs.org/v2/style-guide/#Single-instance-component-names-strongly-recommended) or [Base components](https://vuejs.org/v2/style-guide/#Base-component-names-strongly-recommended) and rename accordingly. Extend this pattern where it is needed.

delapuente:
From the [definition given in Vue docs](https://vuejs.org/v2/guide/components-registration.html#Automatic-Global-Registration-of-Base-Components), I follow that base components are relatively simple, like small wrappers. I don't think the site footer and menu are simple components at all.

However, they do fall under the [_once per page_ category](https://vuejs.org/v2/style-guide/#Single-instance-component-names-strongly-recommended) so I think they should be singleton components prefixed with `The`.

Thoughts @y4izus?

y4izus:
Agree

PR #1007, opened 2020-10-13
Enhance dark page

y4izus:
This PR does several things:
- Implements some of the requested changes requested [here](https://github.com/Qiskit/qiskit.org/pull/987) (the ones where I've added 👍 or a comment saying that I will do it)
- Refactors the page footer
- Creates a mixin for the scroll between sections
- On the button ctas adds the link to request access


components/layouts/PageFooter/index.vue
```diff
@@ -2,32 +2,32 @@
   <footer class="the-page-footer">
```
delapuente, line 2: Now I'm confused, I thought this would fall under your definition of:
> A component that is used across all the web, generally included as part of the main layout

Also, you agreed in prefixing this very component (discussed here #660 and implemented here #842).

In any case, please remember to open the discussion to clarify the rule and, if sticking with the new name, rename the root class to be `page-footer`.

mixins/scrollBetweenSections.ts
```diff
@@ -0,0 +1,62 @@
+import Vue from 'vue'
```
delapuente, line 1: Non-blocking: the trivial factorization is OK. However, making this really reusable would imply separating the concerns of detecting-active-section/scrolling (there is another factorization opportunity here considering the behaviour of the learn page), invoke callbacks (for instance, upon parsing the URL, instead of always changing the section, you invoke `onParseUrl`; or upon a section being marked as active, you invoke `onActiveSectionChange`) so you decouple the detection behaviour from its consequences.

Instead of a mixin, the "scroll with active section detection" could be a component admitting parameters for customizing the detection window and emitting events when some things happen.

Ultimately, there is room for improvement so I would open a follow-up (you can use this comment as a starting point) or retire the factorization and keep track of it in an issue.

delapuente (changes requested):
Almost there. Thank you.

components/metal/BuildingSection.vue
```diff
@@ -4,38 +4,43 @@
...
             Qiskit Metal is a visionary work in progress in its early development stages. We‘re working toward a vision—a quantum design & analysis platform built from the ground up specifically for quantum hardware. It is crucial that we keep this product open source in order to grow Qiskit Metal into the software that will be most useful for quantum hardware engineers around the world and for the innovation in and growth of the field as a whole. If Qiskit Metal doesn't have what you want, then you extend Metal to build it!
           </p>
           <h3 class="copy__subtitle">
             Education
           </h3>
-          <!-- eslint-disable vue/multiline-html-element-content-newline -->
-          <p class="copy__paragraph">
-            A few sentences here about educational material. The first link is to “<AppLink
-              class="building-section__copy-link copy__link "
+          <p class="building-section__text">
```
techtolentino, line 27: The one reason why I had the `eslint-disable` was to prevent this spacing issue, happening in the `AppLink` component: 

Is there a better way of handling this?

components/metal/EarlyAccessSection.vue
```diff
@@ -25,9 +25,8 @@ import AppCtaBtn from '~/components/metal/AppCtaBtn.vue'
 })
 
 export default class extends Vue {
-  // TODO: update with url
   stayInTouchLink = {
-    url: 'https://qiskit.org',
+    url: 'http://qisk.it/metal',
```
techtolentino, line 29: 👍

techtolentino (commented):
Just leaving some comments, Yaiza -- great work here!

components/layouts/PageFooter/FooterSection.vue
```diff
@@ -2,20 +2,19 @@
...
-          'footer-section__icon-link': iconsOnly,
-          'footer-section__icon-link_dark-mode': darkMode,
-          'footer-section__link': !iconsOnly,
-          'footer-section__link_dark-mode': darkMode
-        }"
+        :class="
+          iconsOnly
+            ? `footer-section__icon-link_${mode}-mode`
+            : `footer-section__link footer-section__link_${mode}-mode`
+        "
```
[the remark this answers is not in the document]
y4izus: I think is more clear now: if it's icons only use the styles for icons, if not, use the styles for links. Let me ask the rest of the team, and if they agree with you I will change it, ok?

@techtolentino , @korgan00 , thoughts?

components/layouts/PageFooter/FooterSection.vue
```diff
@@ -2,20 +2,19 @@
...
         :key="element.url"
-        :class="{
-          'footer-section__icon-link': iconsOnly,
-          'footer-section__icon-link_dark-mode': darkMode,
-          'footer-section__link': !iconsOnly,
-          'footer-section__link_dark-mode': darkMode
-        }"
+        :class="
+          iconsOnly
+            ? `footer-section__icon-link_${mode}-mode`
```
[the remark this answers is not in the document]
y4izus: The element class is empty, that is why I removed it

components/layouts/PageFooter/index.vue
y4izus: When I agreed to the discussion we only had one instance of the footer. Now we have two and we create them by using props, so according to the [style guide](https://vuejs.org/v2/style-guide/#Single-instance-component-names-strongly-recommended), it's no longer a single-instance component and we should remove the `The` prefix

I will remove the `the-` from the classes because I forgot to do it

components/layouts/PageFooter/FooterSection.vue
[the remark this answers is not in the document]
delapuente: I'm fine with a discussion and we have discussion issues for it. The comment is intended to be non-blocking, so let's not block on it.

components/layouts/PageFooter/FooterSection.vue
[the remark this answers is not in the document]
delapuente: According to BEM, [a modifier cannot be used alone](https://en.bem.info/methodology/quick-start/#a-modifier-cant-be-used-alone). According to what I read in the BEM specification, this is for highlighting the modifier is "modifying" something and that something is the Element and a ward against future changes in the Element since the Element should factor-out the comon behavior.

components/layouts/PageFooter/index.vue
delapuente: Got it. :+1:

components/metal/BuildingSection.vue
delapuente: Right! That's why I asked and did not require the change. The correct way to do it is [to add `AppLink` to the ignore list of the rule](https://eslint.vuejs.org/rules/multiline-html-element-content-newline.html#options) because it should be considered an inline element. You can modify the rule or simply restore the previous version, perhaps adding a clarifying comment a-la:
```html
<!-- Disabling linter to keep the quotations and phrasing content together. -->
```

y4izus: Could you add the screenshot again? I can't see it and I'm not sure about the spacing issue you are talking about

y4izus: I've reverted my changes and added the comment you suggest

PR #1009, opened 2020-10-13
Implement dark header

### LINKED ISSUE #983: Implement header
Create the header with the title and quantum computer. The title underline should point to the computer processor.

techtolentino:
Resolves #983 

**Notes** 
- given the nature of the specific positioning of elements (responsively), there are many instances of what feel like _magic numbers_, but I'm open to figuring out a better way
- I feel like the page will need some sort of transition that will allow elements to fade in, instead of just flash-appearing when loaded


techtolentino:
@korgan00 - please re-review when you have a chance.

:pray:

components/metal/TheDarkHeader.vue
```diff
@@ -0,0 +1,282 @@
...
+    @include contained();
+
+    position: relative;
+    display: flex;
+    align-items: center;
+    padding-top: 23rem;
+
+    @include mq($from: medium, $until: large) {
+      padding-top: 12.75rem;
+    }
```
korgan00, line 151: I think it could work better this way:
```suggestion
    @include mq($from: medium, $until: large) {
      padding-top: 12.75rem;
      max-width: 740px;
    }
```

components/metal/TheDarkHeader.vue
```diff
@@ -0,0 +1,282 @@
...
+
+    @include mq($until: medium) {
+      padding-top: 14.15rem;
+    }
```
korgan00, line 155: Same here with another width:
```suggestion
    @include mq($until: medium) {
      padding-top: 14.15rem;
      max-width: 450px;
    }
```

components/metal/TheDarkHeader.vue
techtolentino: @korgan00 - For now, I'm going to intentionally skip the `max-width` so that the header title can still align left with the rest of the page content.

components/metal/TheDarkHeader.vue
techtolentino: Same here, I think we can move forward without the `max-width`, so that the page title can align left with the rest of the content.

delapuente (approved):
I'm merging this for us to see all the pieces together. Tomorrow, Yaiza will make us the last review so, let's merge this!

PR #1017, opened 2020-10-15
Added video support in scrolling section

### LINKED ISSUE #1016: Add video to capabilities section

korgan00:
fix #1016 
I think that the video/lazy-image selector can be extracted to a component.


components/metal/CapabilitiesSection.vue
```diff
@@ -42,6 +55,11 @@ export default class extends Mixins(ScrollSectionsMixin) {
   isActiveImage (item: MetalCapability, index: number): boolean {
     return item.title === this.activeSection || (this.activeSection === '' && index === 0)
   }
+
+  isVideo (url: string): boolean {
+    const extension = url.substring(url.length - 4)
```
techtolentino, line 60: 👍

components/metal/CapabilityCard.vue
```diff
@@ -8,9 +8,22 @@
...
       </div>
     </div>
+    <video
+      v-if="isVideo()"
+      class="capability-card__visual-resource"
+      loop
+      autoplay
+      muted
+      playsinline
+    >
```
techtolentino, line 18: @korgan00 - this is something we might have not discussed during design handoff, but for accessibility reasons, I think we need to enable the `controls` attribute, at minimum.

Realistically, and we can do this in a separate issue (which I'll file after), we could be better with our accessible approach by incorporating [custom controls](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/Multimedia).

I just didn't want to block this PR cause what you've built is still within scope.

techtolentino (approved):
Looks great, @korgan00. 

My only comments are regarding `a11y` practices with the video, and I'll be opening an issue to enhance that. For now, I think this is ready.

Tagging @JRussellHuffman just as an FYI.

components/metal/CapabilityCard.vue
korgan00: Ok, thank you, that is interesting.

components/metal/CapabilitiesSection.vue
```diff
@@ -9,19 +9,32 @@
...
       <div class="capabilities-section__scrolling-ui">
         <div
           v-for="(item, index) in capabilities"
-          :key="item.image"
-          class="capabilities-section__image"
+          :key="item.visualResource"
+          class="capabilities-section__visual-resource-container"
           :class="{
-            'capabilities-section__active-image' : isActiveImage(item, index)
+            'capabilities-section__visual-resource-container_active' : isActiveImage(item, index)
```
techtolentino, line 21: one last thing -- should this be `__active`?

PR #1038, opened 2020-10-22
Replaced SVG with a better one for mobile versions

### LINKED ISSUE #1034: Center icon on Empty Events card (on mobile)
We need to fix the centering position of the icon on mobile, for the empty Events state

Screen Shot 2020-10-20 at 12 38 30 PM

korgan00:
Fix #1034


y4izus (approved):
Thank you!

PR #1040, opened 2020-10-22
Enforce no implicit any

### LINKED ISSUE #1028: Enable automated checks for type annotations
Set TypeScript linter to check:
- All property definitions are annotated.
- All signatures are annotated.

delapuente:
Fix #1028

The pull request [enforces non implicit `any` annotations](https://www.typescriptlang.org/tsconfig#noImplicitAny). We now have enabled all the strict check of TypeScript.

This patch also adss some missing types for modules we were using, and declare placeholder modules for those that does not provide type annotations.

In addition, the PR explicitly add `any` type annotations. A follow-up should address the problem of replacing `any` with more especific types.


y4izus (approved):
Thank you!

Issue #1026, opened 2020-10-19
Add 2 additional multi-select dropdown for Location and Type


lerongil:
Size: M

PR #1045, opened 2020-10-23
Add mobile multiselect events filters

techtolentino:
Fixes https://github.com/Qiskit/qiskit.org/issues/1026

**Notes**
- Carbon Vue package had to be updated to prevent an wanted `prop` warning when using [cv-multi-select](http://vue.carbondesignsystem.com/?path=/story/components-cvmultiselect--default) component


pages/events/index.vue
```diff
@@ -161,6 +187,38 @@ export default class extends QiskitPage {
...
+
+  updateTypeFilters (selectedRegions) {
+    const { commit } = this.$store
+    const payload = {
+      filter: 'typeFilters',
+      filterValues: selectedRegions
+    }
+
+    commit('updateFilterSet', payload)
+  }
```
korgan00, line 220: I think that both can be merged into a single function with a new parameter `filter` as the `updateFilter` function does.
Also, if you replace the `selectedXXXX` variable name with `filterValues`, you can follow the `updateFilter` pattern in order to simplify reading.
I don't know if there is any way to merge them with `updateFilter`. I think that there is not a simple solution.

korgan00 (changes requested):
Good job @techtolentino 
Just a few comments.

pages/events/index.vue
techtolentino: Thanks for the guidance here, @korgan00

pages/events/index.vue
```diff
@@ -161,6 +187,38 @@ export default class extends QiskitPage {
...
+    return (listType: string) => {
+      return this[listType].map((item) => {
+        return {
+          label: item,
+          value: item,
+          name: item
+        }
+      })
+    }
+  }
```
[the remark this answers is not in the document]
techtolentino: updated

korgan00:
@JRussellHuffman I have some non-technical thinking here.
When a filter is applied on mobile, it still shows `All locations` with a pill indicating the number of applied filters. I understand it, but I think it is a bit confusing.

pages/events/index.vue
```diff
@@ -161,6 +187,26 @@ export default class extends QiskitPage {
...
+
+  updateFilters (filter: string, filterValues: object) {
+    const { commit } = this.$store
+    const payload = {
+      filter,
+      filterValues
+    }
+
+    commit('updateFilterSet', payload)
+  }
```
korgan00, line 208: I usually avoid naming two functions so similar (only adding an "s"). In this case `updateFilter` and `updateFilters`. I like to complete with a word like `updateFilterValues` or `updateWholeFilter`.

I would like to know @delapuente and @y4izus opinion here.

pages/events/index.vue
```diff
@@ -277,6 +320,82 @@ export default class extends QiskitPage {
...
+
+  &__filters-region,
+  &__filters-type {
+    @include mq($from: medium) {
+      display: none;
+    }
+
+    .bx--list-box--light,
+    .bx--list-box__menu {
+      background-color: $white;
```
korgan00, line 338: We are restyling the multiselect here because it has not the [Carbon Design](http://vue.carbondesignsystem.com/?path=/story/components-cvmultiselect--default) colors of the light theme.
I don't know if there is something inherited, and it doesn't work properly.

korgan00 (approved):

pages/events/index.vue
y4izus: I like the `updateWholeFilter` suggestion because the name expresses perfectly what it's doing

pages/events/index.vue
```diff
@@ -10,11 +10,37 @@
...
       </div>
+      <div class="event-page__filters-region">
+        <client-only>
+          <cv-multi-select
+            class="event-age__filters-region_dropdown"
+            :theme="'light'"
+            :label="'All locations'"
+            :inline="false"
+            :options="getOptions('regions')"
+            :selection-feedback="'fixed'"
```
y4izus, line 21: Same as before, please, use: `selection-feedback="fixed"`

pages/events/index.vue
```diff
@@ -10,11 +10,37 @@
...
+            @change="options => updateFilters('regionFilters', options)"
```
y4izus, line 22: I think `selectedOptions` is more semantic than just `options`. With `options` I think in all the possible options

pages/events/index.vue
```diff
@@ -10,11 +10,37 @@
...
+          />
+        </client-only>
+      </div>
+      <div class="event-page__filters-type">
+        <client-only>
+          <cv-multi-select
```
y4izus, line 28: Please replace the string properties the same way that I mention for the previous multi-select

pages/events/index.vue
```diff
@@ -161,6 +187,26 @@ export default class extends QiskitPage {
     return (this as any).filteredEvents.length === 0
   }
 
+  getOptions (type) {
```
y4izus, line 190: If you use a function to set the options, it will execute several times. Please, create a variable with the values as we did on line 176 and use it

pages/events/index.vue
```diff
@@ -10,11 +10,37 @@
           </cv-tabs>
         </client-only>
       </div>
+      <div class="event-page__filters-region">
+        <client-only>
+          <cv-multi-select
+            class="event-age__filters-region_dropdown"
```
y4izus, line 16: I think this is a typo, it's `event-page`, not `event-age`

y4izus (changes requested):
Thank you Randy! Please, take a look at my requested changes and ask me for a second review when you finish :)

pages/events/index.vue
```diff
@@ -157,10 +183,39 @@ export default class extends QiskitPage {
...
+    return optionsList.map((item) => {
+      return {
+        label: item,
+        value: item,
+        name: item
+      }
+    })
+  }
+
+  updateWholeFilter (filter: string, filterValues: object) {
```
y4izus, line 209: Be more specific and use `Array` instead of `object` and add the return type, please

y4izus (changes requested):
Let's go with the second review. Ask me for a new one when you finish, please :)

pages/events/index.vue
```diff
@@ -157,10 +183,39 @@ export default class extends QiskitPage {
     img: '/images/events/no-events.svg'
   }
 
+  // multiselect
+  regionOptions = this.getOptions(this.regions)
+  typeOptions = this.getOptions(this.types)
+  theme: string = 'light'
+  labelRegions: string = 'All locations'
+  labelTypes: string = 'All types'
```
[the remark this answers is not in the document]
korgan00: I think that has more sense.
As I mentioned [above](https://github.com/Qiskit/qiskit.org/pull/1045#issuecomment-717260049) when we apply a filter, it still has the "All locations" word and it doesn't make sense to me.
Other (but a bit more complicated) is to change the label for a concatenation of the filters:
 * Empty: "All Locations"
 * Not empty: Europe, America...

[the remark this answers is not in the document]
techtolentino: I see two things happening here:

1) I could see Event locations and Event types as an improvement

2) To @korgan00's point - I think this is why the component is designed with the tiny pill component that gives the _checked items count_ (that purple pill). We should decide if we want to pursue the dynamic placeholder update that you're illustrating above - I think it's an interesting pattern.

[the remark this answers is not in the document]
korgan00: I think @JRussellHuffman should decide that.

pages/events/index.vue
```diff
@@ -10,11 +10,37 @@
...
+            class="event-page__filters-region_dropdown"
+            :theme="theme"
+            :label="labelRegions"
+            :inline="inline"
+            :options="regionOptions"
+            :selection-feedback="feedback"
+            @change="selectedOptions => updateWholeFilter('regionFilters', selectedOptions)"
+          />
+        </client-only>
+      </div>
```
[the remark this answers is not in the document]
techtolentino: @y4izus - would you be okay with saving this particular item for a refactor story? Or do you feel it should be addressed prior to merge.

I'd also be willing to take this refactor story because I think it's a good learning opportunity

techtolentino:
@y4izus - regarding the checkbox binding (checked boxes on desktop should show on mobile as well), that's fixed

pages/events/index.vue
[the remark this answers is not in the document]
y4izus: I'm ok with adding it as a refactor issue, but please, create it :)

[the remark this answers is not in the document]
techtolentino: Issue created here - https://github.com/Qiskit/qiskit.org/issues/1051

pages/events/index.vue
```diff
@@ -157,10 +181,33 @@ export default class extends QiskitPage {
...
+  theme: string = 'light'
+  labelRegions: string = 'All locations'
+  labelTypes: string = 'All types'
+  feedback: string = 'fixed'
+
   get noEvents (): boolean {
     return (this as any).filteredEvents.length === 0
   }
 
+  getOptions (optionsList): Array<object> {
```
y4izus, line 196: `object` is very generic, create a type for this `object`. See an example [here](https://github.com/Qiskit/qiskit.org/blob/master/components/landing/TheQuickStart/StartLocally.vue#L78) and [here](https://github.com/Qiskit/qiskit.org/blob/master/components/landing/TheQuickStart/StartLocally.vue#L84)

y4izus (changes requested):
We are almost there!! Just one more change, wait for Russell opinion and we are done!

pages/events/index.vue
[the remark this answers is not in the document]
JRussellHuffman: I think I see the issue. On the larger sizes that don't have the  drop downs, having nothing selected is the same as having everything selected. In other words, no filter is applied so by default its "all". This isn't as clear on mobile. We should make it the labels should just say "Locations" and "Types" and drop the word "All". When something is select from the drop down, the little purple pill populates at the top of the drop down next to "Locations" or "Types". I think this best option and should alleviate confusion around if all locations are selected. And its consistent with the other sizes. Does that make sense?

TL;DR: remove the word "All"

pages/events/index.vue
techtolentino: Updated - any feedback would be appreciated :pray:

y4izus (approved):

PR #1049, opened 2020-10-26
fix issue 1039

### LINKED ISSUE #1039: Too much space between the header and content on the metal page
Go to https://qiskit.org/metal

Too much space between the header section and the section that says "Early access for quantum..."

Close the space between the two section. I recommend 72 px spacing. 

fyi:
@delapuente @y4izus @techtolentino

korgan00:
fix #1039

### DIFF
--- a/components/metal/IntroSection.vue
+++ b/components/metal/IntroSection.vue
@@ -64,7 +64,7 @@ export default class extends Vue {
 
   &__title {
     @include type-style('expressive-heading-05');
-    max-width: 500px;
+    max-width: 32rem;
     margin-bottom: $layout-03;
     font-weight: 300;
 
--- a/components/metal/IntroSection.vue
+++ b/components/metal/IntroSection.vue
@@ -55,7 +55,7 @@ export default class extends Vue {
   @include mq($until: medium) {
     margin-top: -5rem;
   }
-  
+
   &__container {
     @include contained();
     padding-top: $layout-05;
--- a/components/metal/IntroSection.vue
+++ b/components/metal/IntroSection.vue
@@ -46,6 +46,16 @@ export default class extends Vue {
 @import '~carbon-components/scss/globals/scss/typography';
 
 .intro-section {
+  position: relative;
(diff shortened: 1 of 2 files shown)


components/metal/IntroSection.vue
```diff
@@ -54,6 +64,7 @@ export default class extends Vue {
 
   &__title {
     @include type-style('expressive-heading-05');
+    max-width: 500px;
```
techtolentino, line 67: For the sake of sticking with Carbon mini-units (8px), can we use a value here that's a multiple of 8?

`512px`, for example, or `32rem`?

techtolentino: I realize that we have values around the code base that don't stick to this mini-unit principle all the time, but it'd be nice to use it when we can.


thoughts? 

or does this not make sense to do?

techtolentino (approved):
This is looks great, @korgan00 - just had a question for discussion but I approve what I see here.


Thanks

components/metal/IntroSection.vue
y4izus: If a value doesn't stick to the mini-unit principle is because it was done before we started using the carbon design system. Now we try to use them

components/metal/IntroSection.vue
```diff
@@ -54,6 +64,7 @@ export default class extends Vue {
 
   &__title {
     @include type-style('expressive-heading-05');
+    max-width: 32rem;
```
y4izus, line 67: For `max-width` property we are using the columns size variable for each screen. Look the example on line 87: `max-width: 6 * $column-size-large;`. Could you change this `32rem` with the proper value using `$column-size-large;`?

y4izus (changes requested):
Thank you! Just two things:
- The comment I added 👇 
- The build is failing and we can't merge until it's fixed. Please, run `npm run lint` to see the error. I think you can fix it by running: `npm run fix-lint`

y4izus (approved):
Thank you! :D

PR #1055, opened 2020-10-29
Align filter arrows on event page

### LINKED ISSUE #1052: Dropdown arrows in mobile size are not completely aligned
Go visit integration branch for light-themed event page:
https://qiskitorg-git-feature-update-events-page.qiskit.now.sh/events

image

The three dropdown arrows should be aligned.

y4izus:
Fix #1052

### DIFF
--- a/pages/events/index.vue
+++ b/pages/events/index.vue
@@ -353,6 +353,10 @@ export default class extends QiskitPage {
       color: $white-text-01;
     }
 
+    .bx--list-box__menu-icon {
+      right: 0.8rem;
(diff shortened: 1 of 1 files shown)


techtolentino (approved):
this looks good, @y4izus - thanks for fixing.

(I wonder how this happened in the first place 🤔)

y4izus:
@techtolentino this happened because although they look the same, we are working with two different components, the multi select and the tabs, and they have different alignment 🤷‍♀️

Issue #1022, opened 2020-10-19
Discuss name conventions


y4izus:
After the discussion [here](https://github.com/Qiskit/qiskit.org/pull/987#discussion_r503244231), we decided to create an issue to decide which name convention to follow to be all on the same page

PR #1078, opened 2020-11-02
Add Advocates page header

### LINKED ISSUE #1077: Advocates Header
Size: S

techtolentino:
Fixes #1077


y4izus (changes requested):
Thank you! I've added a requested change and I wanted to know the opinion of the rest of the team on this: do you prefer to remove all the page at once as it's done here or do you prefer to replace the components step by step when we do them?? I prefer the replacements step by step, but I would like to read your opinions :) @techtolentino , @korgan00 , @delapuente

korgan00:
I have no preference for this.

techtolentino:
@y4izus - yeah, I thought about this yesterday... I'm honestly fine either way.

I took this approach because I thought it'd be easier from the implementing/reviewing standpoint, since you'd only be looking at the code that's a part of the new design.

But I've also worked in both approaches, so if you prefer, I can change that here and _only_ get the Header setup.

techtolentino:
@y4izus @korgan00 -  I reverted some changes so that this PR is updating the Header only. This might be the better approach..

y4izus (approved):

pages/advocates/index.vue
```diff
@@ -1,22 +1,6 @@
...
-            <CompactFeature
-              v-for="benefit in advocateBenefits"
-              :key="`benefit-${benefit.icon}`"
-              v-bind="benefit"
-            />
-          </div>
-        </template>
-      </CommunityHeader>
-    </header>
+    <TheAdvocatesHeader />
```
y4izus, line 3: Not a blocker, it is just in case you don't know it: We have [this issue](https://github.com/Qiskit/qiskit.org/issues/1022) to talk about how we should name the components. From my point of view, this should be `AdvocatesHeader`, but we should discuss it together :)

y4izus (approved):
It's perfect Randy! 🎉  thank you very much!

korgan00 (approved):
Good job

PR #1084, opened 2020-11-04
Adds redesigned Advocates page

y4izus:
PR created for review purpose before merging with master


components/ui/CompactFeature.vue
```diff
@@ -1,51 +1,63 @@
...
 import Vue from 'vue'
 import { Component, Prop } from 'vue-property-decorator'
 
 @Component
-export default class CompactFeature extends Vue {
-  @Prop(String) icon: string | undefined
-  @Prop(String) title: string | undefined
-  @Prop(String) description: string | undefined
+export default class extends Vue {
+  @Prop(String) icon!: string
```
delapuente, line 23: Be careful, if the property is not required, it **can be `undefined`** and the type should reflect that.

components/ui/AppMultiSelect.vue
```diff
@@ -1,27 +1,41 @@
...
 
+type multiSelectOption = {
+  label: string,
+  value: string,
+  name: string
+}
+
 @Component
 export default class AppMultiSelect extends Vue {
   @Prop(Array) options!: string[]
```
delapuente, line 27: Same here. We can also provide a default so the `!` would make sense then.

components/ui/InfiniteScroll.vue
```diff
@@ -0,0 +1,48 @@
...
+  </div>
+</template>
+
+<script lang="ts">
+import { Component, Mixins, Prop, Watch } from 'vue-property-decorator'
+import ScrollSectionsMixin from '~/mixins/scrollBetweenSections'
+
+@Component
+export default class extends Mixins(ScrollSectionsMixin) {
+  @Prop(Array) items!: Array<any>
```
delapuente, line 20: Be careful here too.

components/ui/InfiniteScroll.vue
```diff
@@ -0,0 +1,48 @@
...
+    </div>
+  </div>
+</template>
+
+<script lang="ts">
+import { Component, Mixins, Prop, Watch } from 'vue-property-decorator'
+import ScrollSectionsMixin from '~/mixins/scrollBetweenSections'
+
+@Component
+export default class extends Mixins(ScrollSectionsMixin) {
```
delapuente, line 19: Missing name of the class for better debugging.

components/ui/InfiniteScroll.vue
```diff
@@ -0,0 +1,48 @@
...
+  @Prop(Array) items!: Array<any>
+  @Prop(Function) keyGenerator!: Function
+  @Prop({ type: Number, default: 5 }) minItems!: number
```
delapuente, line 22: Here, the hint `!` makes sense!

delapuente (changes requested):
It seems the rebase of this branch introduced some regressions regarding class names and some of them are missing.

Main problems have to do with following BEM conventions, computed properties and incorrect type annotations. I'll take care of the problems although I wanted to make the review for the record.

delapuente (approved):
Very nice done, folks! Let's merge.

Issue #944, opened 2020-09-22
👍👍 Reuse filters and results layout on learn and events pages


y4izus:

y4izus:
I've created a layout while doing the advocates page, so I rename this issue to use this component on learn and events page

PR #1087, opened 2020-11-05
Implements meet advocates layout

### LINKED ISSUE #1074: Implement layout for "meet advocate" section
Size:

y4izus:
Fix #1074 

This PR:
- Creates a component for the filters-results layouts we have on learn's, events', and advocates' page but it's only used on the advocates' page. We have an [issue to extend this layout](https://github.com/Qiskit/qiskit.org/issues/944) to the rest of the pages
- Creates an AppCheckbox component that should be finished and reused on the events' page when [implementing the behavior of the filter](https://github.com/Qiskit/qiskit.org/issues/1086)

I did my best naming the components, but I'm not sure if it's easy to follow, so please, if you have suggestions, let me know

### DIFF
--- a/components/advocates/MeetTheAdvocates.vue
+++ b/components/advocates/MeetTheAdvocates.vue
@@ -57,7 +57,7 @@ export default class extends Vue {
 
   filter = {
     label: 'Locations',
-    options: ['Europe', 'Asia', 'Africa', 'America'],
+    options: ['Americas', 'Asia Pacific', 'Europe', 'Africa'],
     filterType: 'regionFilters'
   }
 }
--- a/components/advocates/MeetTheAdvocates.vue
+++ b/components/advocates/MeetTheAdvocates.vue
@@ -1,50 +1,42 @@
 <template>
   <div class="meet-the-advocates">
-    <div class="meet-the-advocates__container">
-      <h2 class="copy__title">
-        Meet the Advocates
-      </h2>
-      <p class="copy__paragraph copy__paragraph_importance_support">
-        Qiskit advocates are some of the finest minds in quantum computing,
-        all over the world. If you are interested in getting involved with the
-        quantum computing community, reach out to an advocate local to your area.
-      </p>
-      <div
-        class="meet-the-advocates__extra-filters meet-the-advocates__extra-filters_on-small-screen"
-      >
-        <AppMultiSelect
-          :label="extraFilter.label"
-          :options="extraFilter.options"
+    <h2 class="copy__title">
+      Meet the Advocates
+    </h2>
+    <p class="copy__paragraph copy__paragraph_importance_support">
+      Qiskit advocates are some of the finest minds in quantum computing,
+      all over the world. If you are interested in getting involved with the
+      quantum computing community, reach out to an advocate local to your area.
+    </p>
+    <AppFiltersResultsLayout>
+      <template slot="filters-on-m-l-screen">
+        <AppFieldset :label="filter.label">
+          <client-only>
+            <AppCheckbox
+              v-for="option in filter.options"
+              :key="option.label"
+              v-bind="option"
+            />
+          </client-only>
+        </AppFieldset>
+      </template>
+      <template slot="filters-on-s-screen">
+        <AppMultiSelect v-bind="filter.label" />
+      </template>
+      <template slot="results">
+        <AdvocateCard
+          v-for="advocate in advocates"
+          :key="`advocate-${advocate.attributes.name}`"
+          v-bind="advocate.attributes"
         />
-      </div>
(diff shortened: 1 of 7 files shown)


korgan00 (changes requested):

components/ui/AppFiltersResultsLayout.vue
```diff
@@ -0,0 +1,58 @@
...
+  margin-top: $layout-05;
+
+  @include mq($until: medium) {
+    flex-direction: column;
+  }
+
+  &__filters {
+    color: $gray-100;
+
+    &_on-large-screen {
```
techtolentino, line 36: Sometimes, I have the tendency to combine the `display: none` for both `_on-large-screen` and `_on-small-screen` - here I see you have them separated.

Is it better to keep them separate?

techtolentino (changes requested):
This looks great, @y4izus 

I have one request, and one question...

components/ui/AppMultiSelect.vue
```diff
@@ -1,27 +1,41 @@
...
-  />
+  <client-only>
+    <cv-multi-select
+      class="app-multi-select"
+      :theme="theme"
+      :label="label"
+      :options="formatedOptions"
+      :value="value"
+      :selection-feedback="feedback"
+      @change="$emit('change-on-multi-select', $event)"
```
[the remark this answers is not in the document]
y4izus: Agree, I like the `change-selection` option :)

components/ui/AppCheckbox.vue
```diff
@@ -0,0 +1,41 @@
...
+export default class extends Vue {
+  @Prop(String) option!: string
+  // TODO: Add checked, aria-checked and @change when doing the filter behaviour. Also
+  // reuse this component on events page
+
+  formatedOption = this.formatOption(this.option)
+
+  formatOption (option: string): ckeckboxOption {
+    return { label: option, value: option }
+  }
```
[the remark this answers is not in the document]
y4izus: The `option` is declared as [`prop`](https://vuejs.org/v2/api/#props), i.e, an attribute that accepts data from the parent. By default, it's not reactive. If you want to modify a value on the parent and see how the child catches the modification, use [@ProvideReactive](https://github.com/kaorun343/vue-property-decorator#-providereactivekey-string--symbol--injectreactiveoptions--from-injectkey-default-any---injectkey-decorator).

On the other hand, as you can read [here](https://class-component.vuejs.org/guide/class-component.html#data), we are creating the `formattedOption` variable as component data. And that means that it will be reactive unless its value is undefined.

So, in this code, `option` is not reactive, but `formattedOption` is.

Does this answer your question?

components/ui/AppFiltersResultsLayout.vue
y4izus: I keep them separate because the media queries are different. What do you have in mind to combine them? Maybe you have a better solution! :)

techtolentino: Disregard - this makes sense...

components/advocates/MeetTheAdvocates.vue
```diff
@@ -0,0 +1,64 @@
...
+    AppCheckbox,
+    AppFiltersResultsLayout
+  }
+})
+export default class extends Vue {
+  @Prop(Array) advocates!: any
+
+  filter = {
+    label: 'Locations',
+    options: ['Europe', 'Asia', 'Africa', 'America'],
```
techtolentino, line 60: I think this should be `Americas`

y4izus: Yes! And `Asia` should be `Asia Pacific` to match the filters on the events' page. Thank you Randy!

korgan00: Maybe we should extract that filter configuration to avoid incongruences between filters.

y4izus: Yes, but I think it could be better to do it when implementing the filter. This issue is related just with the layout. Do you agree?

components/advocates/MeetTheAdvocates.vue
```diff
@@ -0,0 +1,64 @@
...
+              v-for="option in filter.options"
+              :key="option"
+              :option="option"
+            />
+          </client-only>
+        </AppFieldset>
+      </template>
+      <template slot="filters-on-s-screen">
+        <AppMultiSelect v-bind="filter" />
+      </template>
```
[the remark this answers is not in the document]
y4izus: Agree, but let's finish [this issue](https://github.com/Qiskit/qiskit.org/issues/944) first to easily extract the component

techtolentino (approved):

korgan00 (approved):

PR #1089, opened 2020-11-06
Add join section

### LINKED ISSUE #1075: Implement join program section
Size: M

techtolentino:
Fixes #1075


components/advocates/AdvocatesJoinSection.vue
```diff
@@ -0,0 +1,117 @@
...
+
+      &-cta {
+        width: 100% !important;
+        justify-content: space-between;
+      }
+    }
+
+    &__feature {
+      width: 4 * $column-size-large;
+      margin-left: $spacing-05;
```
korgan00, line 102: The design has double space, you can set just `margin: $spacing-05;` because flex does not overlap margins.
To keep it in sync with `__access` you can put the same there.

components/advocates/AdvocatesJoinSection.vue
```diff
@@ -0,0 +1,117 @@
...
+    url: '#meet-the-advocates',
+    label: 'Find an advocate'
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+  @import '~carbon-components/scss/globals/scss/typography';
+  .join-section {
+    @include contained();
```
korgan00, line 59: In very large, padding is not even on both sides, I think contained is the problem. If you change the `__feature` margin, this should have `padding: 0 $spacing-05;`  to match the design.

components/advocates/AdvocatesJoinSection.vue
```diff
@@ -0,0 +1,117 @@
...
+      padding-left: $spacing-05;
+    }
+
+    &__access {
+      @include type-style('body-long-01');
+      color: $white-text-01;
+      display: flex;
+      flex-direction: column;
+      justify-content: space-between;
+      width: 4 * $column-size-large;
```
korgan00, line 75: You are using flex. I think it is better to use `flex` combined property or `flex-basis`. I suggest maybe something like `flex: 1 1 4 * $column-size-large;`
[flexbox guide](https://css-tricks.com/snippets/css/a-guide-to-flexbox/)

components/advocates/AdvocatesJoinSection.vue
```diff
@@ -0,0 +1,117 @@
...
+
+      @include mq($until: large) {
+        width: 45%;
```
korgan00, line 78: Same flex things here but `flex: 1 1 45%;`

components/advocates/AdvocatesJoinSection.vue
```diff
@@ -0,0 +1,117 @@
...
+      }
+
+      &-cta {
+        width: 100% !important;
+        justify-content: space-between;
+      }
+    }
+
+    &__feature {
+      width: 4 * $column-size-large;
```
korgan00, line 101: Same flex things here

components/advocates/AdvocatesJoinSection.vue
```diff
@@ -0,0 +1,117 @@
...
+      margin-left: $spacing-05;
+      padding-bottom: $spacing-09;
+
+      @include mq($until: large) {
+        width: 45%;
```
korgan00, line 106: Same flex things here

components/advocates/TheAdvocatesHeader.vue
```diff
@@ -22,3 +22,11 @@ import TypewriterEffect from '~/components/ui/TypewriterEffect.vue'
...
 export default class extends Vue { }
 </script>
+
+<style lang="scss">
+.advocates-header {
+  .page-header__title {
+    max-width: initial;
+  }
+}
+</style>
```
korgan00, line 32: This is not a join section change, so I think it is better to keep these changes in a separate PR.

korgan00 (changes requested):

korgan00:
Good work @techtolentino

components/advocates/AdvocatesJoinSection.vue
techtolentino: @korgan00 - I think I'm missing what this looks like, can you please attach a screenshot :pray:

korgan00: 2020-11-09 10_08_23-Window

techtolentino: @korgan00 - I know that this `contained()` mixin was created long before I began on this project, and it seems like it's supposed to be that way?

@y4izus - can you confirm?

y4izus: You both are right. On the one hand, yes, we use the `contained()`, but on the other hand, if I have understood Goyo properly, this `contained()` is making your code more complicated when managing the margins/paddings. If you add the `padding: 0 $spacing-05;` right before the `contained()`, you can easily align the items.

@korgan00 , that's what you were trying to explain, right?

techtolentino: If I'm understanding @korgan00 correctly... the design shows `1rem` of left and right padding, but here, our container is showing `padding-left: 2.5rem; padding-right: 1rem`... and this is the problem, is that right?

The way I'm using the `contained` mixin on this page, I believe (correct me if I'm mistaken), is consistent with how it's used on other pages.

korgan00: The design shows 2rem on each side. In your implementation, you have 1.5 from the container padding and 1 from the margin on the right side, and 2.5 from the container on the left.
That forces to make a difference between `__feature`  and `__access` margins and makes the CSS more complex.
If you apply this little change. You can merge the margins, flex, and media-queries from the `__feature`  and `__access`
```suggestion
    @include contained();
    padding: 0 1rem;
```

korgan00 (changes requested):

eddybrando (approved):
LGMT!

But since I'm new, please wait for a _more authoritative_ approval before merging 😄

korgan00 (approved):
It is ok for me. 
Good job

y4izus (approved):
Great work and great reviews!! Thanks!!

PR #1097, opened 2020-11-10
Remove header title width constraint

techtolentino:
No issue related.

This updates the Advocates header to match the designs, on full-screen

### DIFF
--- a/components/advocates/TheAdvocatesHeader.vue
+++ b/components/advocates/TheAdvocatesHeader.vue
@@ -26,7 +26,9 @@ export default class extends Vue { }
 <style lang="scss">
 .advocates-header {
   .page-header__title {
-    max-width: none;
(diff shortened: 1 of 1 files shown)


korgan00 (approved):

delapuente:
Non-blocking, what is the effect of removing the constrains for all the titles by putting this indie the `ThePageHeader` component? Is it good enough to become the norm or there are cases in which we prefer the old style and keep this specific as _the exception_.

korgan00:
Yaiza suggested normalizing this, and then I talked to Russell to knows design opinion. I will create an Issue to change `ThePageHeader` and check all the affected pages.

y4izus:
I will close this PR because we are going to do it for all [here](https://github.com/Qiskit/qiskit.org/issues/1104)

PR #1099, opened 2020-11-11
Simple infinite scrolling implementation used on MeetTheAdvocates component

### LINKED ISSUE #1073: Implement Infinite scroll
Size: M

korgan00:
fix #1073

### DIFF
--- a/components/advocates/MeetTheAdvocates.vue
+++ b/components/advocates/MeetTheAdvocates.vue
@@ -24,11 +24,14 @@
         <AppMultiSelect v-bind="filter" />
       </template>
       <template slot="results">
-        <AdvocateCard
-          v-for="advocate in advocates"
-          :key="`advocate-${advocate.attributes.name}`"
-          v-bind="advocate.attributes"
-        />
+        <InfiniteScroll
+          v-slot="advocate"
+          :items="advocates"
+          :min-items="5"
+          :key-generator="(advocate) => advocate.attributes.name"
+        >
+          <AdvocateCard v-bind="advocate.attributes" />
+        </InfiniteScroll>
       </template>
     </AppFiltersResultsLayout>
   </div>
@@ -42,6 +45,7 @@ import AppMultiSelect from '~/components/ui/AppMultiSelect.vue'
 import AppFieldset from '~/components/ui/AppFieldset.vue'
 import AppCheckbox from '~/components/ui/AppCheckbox.vue'
 import AppFiltersResultsLayout from '~/components/ui/AppFiltersResultsLayout.vue'
+import InfiniteScroll from '~/components/ui/InfiniteScroll.vue'
 
 @Component({
   components: {
@@ -49,7 +53,8 @@ import AppFiltersResultsLayout from '~/components/ui/AppFiltersResultsLayout.vue
     AppMultiSelect,
     AppFieldset,
     AppCheckbox,
-    AppFiltersResultsLayout
+    AppFiltersResultsLayout,
+    InfiniteScroll
   }
 })
 export default class extends Vue {
--- a/components/ui/InfiniteScroll.vue
+++ b/components/ui/InfiniteScroll.vue
@@ -1,0 +1,38 @@
+<template>
+  <section>
+    <div
+      v-for="(item, index) in items.slice(0, lastShown)"
+      :id="index + '-' + keyGenerator(item)"
+      :key="keyGenerator(item)"
+      class="scrollable"
+    >
+      <slot v-bind="item" />
+    </div>
+  </section>
+</template>
+
+<script lang="ts">
+import { Component, Mixins, Prop } from 'vue-property-decorator'
+import ScrollSectionsMixin from '~/mixins/scrollBetweenSections'
+
+@Component
+export default class extends Mixins(ScrollSectionsMixin) {
+  @Prop(Array) items!: Array<any>
+  @Prop(Function) keyGenerator!: Function
+  @Prop({ type: Number, default: 5 }) minItems!: number
+
+  lastShown: number = this.minItems
+
+  activeSectionChanged () : void {
+    const current: number = this.activeSection ? parseInt(this.activeSection.split('-', 2)[0]) : 0
+    const oldLastShown = this.lastShown
+    this.lastShown = Math.max(Math.min(current + this.minItems, this.items.length), this.lastShown)
+    if (oldLastShown < this.lastShown) {
+      this.updateObserved()
+    }
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+</style>
--- a/components/ui/InfiniteScroll.vue
+++ b/components/ui/InfiniteScroll.vue
@@ -21,7 +21,7 @@ export default class extends Mixins(ScrollSectionsMixin) {
   @Prop(Function) keyGenerator!: Function
   @Prop({ type: Number, default: 5 }) minItems!: number
 
-  lastShown: number = this.minItems
+  lastShown: number = Math.min(this.minItems, this.items.length)
 
   activeSectionChanged () : void {
     const current: number = this.activeSection ? parseInt(this.activeSection.split('-', 2)[0]) : 0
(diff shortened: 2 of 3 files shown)


components/ui/InfiniteScroll.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <section>
```
eddybrando, line 2: I'd argue that we should get rid of the `section` wrapper here since the `InfiniteScroll` component looks more like a reusable UI element / factory, than a [**standalone section**](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section).

As per [spec](https://html.spec.whatwg.org/multipage/sections.html#the-section-element):

> The section element represents a generic section of a document or application. A section, in this context, is a thematic grouping of content, typically with a heading.

and

> The section element is not a generic container element. When an element is needed only for styling purposes or as a convenience for scripting, authors are encouraged to use the div element instead. A general rule is that the section element is appropriate only if the element's contents would be listed explicitly in the document's outline.

In this context, I don't believe that this block should be semantically considered a "section". The section would rather be the parent element that includes the cards with infinite scrolling, but also the filter and the "Meet the Advocates" headline.

techtolentino (approved):
This is awesome @korgan00, especially for the simple infinite scroll... 

At some point, or in future iterations, I wonder if we need some sort of load more button... 

Nice work!

mixins/scrollBetweenSections.ts
```diff
@@ -50,12 +58,17 @@ export default class extends Vue {
...
+        if (this.activeSection !== target.id) {
+          this.activeSection = target.id
+          this.activeSectionChanged()
+        }
         highestTopValue = targetTop
       }
     })
   }
 
+  activeSectionChanged () {}
```
eddybrando, line 70: Probably adding a small JSDoc here would be beneficial to avoid ambiguity at first glance.

From the name of the method, it could also be interpreted as a function that performs a check to see if the active section has changed and returns the evaluation of that check.

To be fair, after checking the `void` return type and the implementation in the extendable class, it's clear what it does, but still, some cognitive load could be saved with something like this:

```suggestion
  /**
   * This methods gets called when the active section changes.
   */
  activeSectionChanged () {}
```

components/advocates/MeetTheAdvocates.vue
```diff
@@ -24,11 +24,14 @@
...
-        <AdvocateCard
-          v-for="advocate in advocates"
-          :key="`advocate-${advocate.attributes.name}`"
-          v-bind="advocate.attributes"
-        />
+        <InfiniteScroll
+          v-slot="advocate"
+          :items="advocates"
+          :min-items="5"
+          :key-generator="(advocate) => advocate.attributes.name"
```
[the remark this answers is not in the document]
korgan00: Yes, you are right. I wanted a first quick simple implementation and make it easy to change later.

components/ui/InfiniteScroll.vue
```diff
@@ -0,0 +1,38 @@
...
+    const oldLastShown = this.lastShown
+    this.lastShown = Math.max(Math.min(current + this.minItems, this.items.length), this.lastShown)
+    if (oldLastShown < this.lastShown) {
+      this.updateObserved()
+    }
+  }
+}
+</script>
+
+<style lang="scss" scoped>
```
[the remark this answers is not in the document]
korgan00: I don't really know. I use to put the 3 tags for code completion and facilitate modification tasks.
I think we can set a rule for this if it is not set. 

What do you think @y4izus @delapuente @eddybrando

components/ui/InfiniteScroll.vue
```diff
@@ -0,0 +1,38 @@
...
+    </div>
+  </section>
+</template>
+
+<script lang="ts">
+import { Component, Mixins, Prop } from 'vue-property-decorator'
+import ScrollSectionsMixin from '~/mixins/scrollBetweenSections'
+
+@Component
+export default class extends Mixins(ScrollSectionsMixin) {
```
eddybrando, line 19: Naming component classes can help with DX by making them easier to distinguish with [Vue.js devtools](https://github.com/vuejs/vue-devtools):

```suggestion
export default class InfiniteScroll extends Mixins(ScrollSectionsMixin) {
```

This would result in:

Screenshot 2020-11-11 at 16 46 13

components/ui/InfiniteScroll.vue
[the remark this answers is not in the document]
eddybrando: The only real difference I'm aware of is that **Vue Loader** will still add the `data-v-<hash>` attributes to the component DOM element and its children, even if the `style` tag is empty, because of the `scoped` attribute. Like this: https://vue-loader.vuejs.org/guide/scoped-css.html#scoped-css

I'd personally prefer to remove the `style` tag form the component file if it's not being used.

components/ui/InfiniteScroll.vue
korgan00: This is a project convention. If that should change, we have to change the whole project.

components/ui/InfiniteScroll.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <section>
+    <div
+      v-for="(item, index) in items.slice(0, lastShown)"
+      :id="index + '-' + keyGenerator(item)"
+      :key="keyGenerator(item)"
```
eddybrando, line 6: I believe we can simplify this like so:

```suggestion
      :id="`${_uid}-${index}`"
      :key="index"
```

Since the elements in the array used by the `v-for` don't change order, I think that the `index` of the list should a good **key** here.

In order to avoid duplicate IDs, we could use Vue's component `_uid` and then append the index.

components/advocates/MeetTheAdvocates.vue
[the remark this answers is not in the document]
eddybrando: I added an comment that may tackle this issue: https://github.com/Qiskit/qiskit.org/pull/1099/files#r521472241

components/ui/InfiniteScroll.vue
korgan00: The elements will be filtered at some point. I think that invalidates the key.

components/ui/InfiniteScroll.vue
eddybrando: Is there a discussion issue related to this? I'm afraid I couldn't find one if that's the case.

Otherwise I'd open an issue to have a discussion about pros and cons that this change could convey.

mixins/scrollBetweenSections.ts
korgan00: I was thinking of multiple implementations such as an event or a variable callback. I thought that is the most simple implementation but I can change it.
I think that writing some documentation will be good anyway.

components/ui/InfiniteScroll.vue
korgan00: The elements will be filtered at some point. I think that invalidates the key, but I'm not sure

components/ui/InfiniteScroll.vue
korgan00: Ok, open it.

mixins/scrollBetweenSections.ts
korgan00: I added a comment on InfinteScroll also to make it easier to understand.

components/ui/InfiniteScroll.vue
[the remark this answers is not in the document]
y4izus: Agree with @eddybrando

mixins/scrollBetweenSections.ts
eddybrando: As an alternative renaming the method to something like `activeSectionChangedCallback` or `activeSectionChangedCb` could also remove said ambiguity IMO.

y4izus (approved):
Awesome job! thank you!

components/ui/InfiniteScroll.vue
eddybrando: Opened here: https://github.com/Qiskit/qiskit.org/issues/1103

PR #1101, opened 2020-11-11
Update advocate card

### LINKED ISSUE #1088: Create advocate card

korgan00:
Fix #1088 

This PR affects AppCard, so the events card should be checked too.

### DIFF
--- a/components/advocates/AdvocateCard.vue
+++ b/components/advocates/AdvocateCard.vue
@@ -1,45 +1,37 @@
 <template>
-  <article class="advocate-card">
-    <div
-      class="advocate-card__picture"
-      :style="`background-image: ${decorate(image)};`"
-    />
-    <div class="advocate-card__copy">
-      <h3 class="advocate-card__name">
-        {{ name }}
-      </h3>
+  <AppCard
+    class="advocate-card"
+    :image="image"
+    :title="name"
+    :tags="[region]"
+  >
+    <div>
       <p class="advocate-card__location">
+        <Map20 class="advocate-card__icon" />
         {{ location }}
       </p>
-      <h3 class="advocate-card__focus-areas-title">
-        Focus areas
-      </h3>
-      <p class="advocate-card__focus-areas">
-        {{ areas }}
+      <p class="advocate-card__contact">
+        <Calendar20 class="advocate-card__icon" />
+        <time>{{ slackId }}</time>
       </p>
     </div>
-  </article>
+  </AppCard>
 </template>
 
 <script lang="ts">
 import Vue from 'vue'
 import { Component, Prop } from 'vue-property-decorator'
+import AppCard from '~/components/ui/AppCard.vue'
 
-@Component
+@Component({
+  components: { AppCard }
+})
 export default class extends Vue {
   @Prop(String) name!: any
   @Prop(String) image!: any
   @Prop(String) location!: any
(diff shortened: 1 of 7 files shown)


components/advocates/AdvocateCard.vue
```diff
@@ -1,79 +1,66 @@
...
-      <h3 class="advocate-card__name">
-        {{ name }}
-      </h3>
+  <AppCard
+    class="advocate-card"
+    :image="image"
+    :title="name"
+    :tags="formattedRegion"
+  >
+    <div>
```
techtolentino, line 8: I don't think we need this `div` anymore

content/advocates/index/hassi-norlen.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: Hassi Norlen
-image: hassi-norlen.png
+image: /images/advocates/hassi-norlen.png
```
techtolentino, line 3: I'm curious about the updated path... 

is this to match Yaiza's updates?

techtolentino (commented):
This looks great, @korgan00 

Just a couple of comments...

content/advocates/index/hassi-norlen.md
korgan00: I didn't changed that. So, i supose yaiza knows it.
This PR has most commits from Yaiza an some css commits from me

components/advocates/AdvocateCard.vue
korgan00: I will check it tomorrow. Good catch.

content/advocates/index/hassi-norlen.md
techtolentino: This makes sense...

y4izus: Yes, I did it. The previous advocate's card manually added the `/images/advocates/`, but when reusing the `AppCard`, I needed to add all the path to the mock data. Anyways, these .md files are going to disappear because we are going to recover all the data from airtable :)

components/advocates/AdvocateCard.vue
```diff
@@ -1,79 +1,66 @@
...
       <p class="advocate-card__location">
+        <Map20 class="advocate-card__icon" />
         {{ location }}
       </p>
-      <h3 class="advocate-card__focus-areas-title">
-        Focus areas
-      </h3>
-      <p class="advocate-card__focus-areas">
-        {{ areas }}
+      <p class="advocate-card__contact">
```
y4izus, line 13: For some advocates, we don't have the slack contact, could you add a `v-if="slackId"` to only show this block when we have the slack id?

components/advocates/AdvocateCard.vue
```diff
@@ -1,79 +1,66 @@
...
+        <LogoSlack20 class="advocate-card__icon" />
+        {{ slackId }}
```
y4izus, line 15: As the first approach, we can use [this method](https://slack.com/intl/en-es/help/articles/360003827751-Create-a-link-to-a-members-profile-#free,-standard,-and-plus-plans-1) to open a conversation with the advocates. Could you prepare the `AppLink` here for that instead of having just the `{{ slackId }}`?

y4izus (changes requested):
Great! thank you! just two comments :)

y4izus (approved):

PR #1111, opened 2020-11-12
Add slack link

### LINKED ISSUE #1110: Add Slack link to advocate card

techtolentino:
Fixes #1110 

- Updates logic to only show advocates that have provided consent and their SlackId
- Updates mock data to render Slack links on Advocates cards
- Update intro copy that has the join Slack link


components/advocates/AdvocateCard.vue
```diff
@@ -37,6 +37,7 @@ export default class extends Vue {
   @Prop(String) country!: any
   @Prop(String) region!: any
   @Prop(String) slackId!: any
+  @Prop(String) slackUsername!: any
```
eddybrando, line 40: If it's known that the Slack username is a string, we should probably also set the TS type accordingly:

```suggestion
  @Prop(String) slackUsername!: string
```

hooks/advocate-conversion-utils.ts
```diff
@@ -28,7 +29,12 @@ async function fetchAdvocates (apiKey: string): Promise<Advocate[]> {
   }).eachPage((records, nextPage) => {
     for (const record of records) {
       const advocate = convertToAdvocate(record)
-      advocates.push(advocate)
+      // only commit to store, advocates
+      // who have filled out the approval form
+      // and provided their Slack ID
```
eddybrando: As per spec, a multi-line comment should be written like this:

```suggestion
      /**
       * only commit to store, advocates who have filled out the approval form
       * and provided their Slack ID
       */
```

https://www.ecma-international.org/ecma-262/5.1/#sec-7.4

components/advocates/MeetTheAdvocates.vue
```diff
@@ -8,6 +8,11 @@
       all over the world. If you are interested in getting involved with the
       quantum computing community, reach out to an advocate local to your area.
     </p>
+    <p class="copy__paragraph copy__paragraph_importance_support">
+      <strong>Note:</strong> you must have a <AppLink class="copy__link" :url="'https://ibm.co/joinqiskitslack'">
```
y4izus, line 12: You're using a fixed url, so use `url` instead of `:url`. Remeber that [`:` is the shorthand of `v-bind`](https://vuejs.org/v2/api/#v-bind) and [`v-bind` is used to bind HTML attribute to a JS expression](https://masteringjs.io/tutorials/vue/bind)

components/advocates/MeetTheAdvocates.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="meet-the-advocates">
+  <section id="meet-the-advocates" class="meet-the-advocates">
```
y4izus, line 2: Good replacement! div -> section 👏

content/advocates/advocates.json
```diff
@@ -3,13 +3,21 @@
     "name": "Rana Prathap Simh Mukthavaram",
     "image": "https://dl.airtable.com/.attachmentThumbnails/568134fea19c1e34a98c06cfd77a86ec/949489e7",
     "region": "Asia",
-    "location": "India"
+    "location": "India",
```
y4izus, line 6: We are not getting the location from the airtable, we are building it by joining the city and the country. It will be better to have mock data that fits the real one

tests/hooks/advocate-conversion-utils.spec.ts
```diff
@@ -15,22 +15,24 @@ type RecordFields = {
   city?: string,
   country?: string,
   region?: AdvocatesWorldRegion
-  // slackId: string
+  slackId: string,
+  slackUsername: string
```
y4izus, line 19: These properties are optional, right? Correct me if I'm wrong

tests/hooks/advocate-conversion-utils.spec.ts
```diff
@@ -15,22 +15,24 @@ type RecordFields = {
...
 }
 
-// TODO: Add tests for slackId
+// TODO: Add tests for slackId, slackUsername
```
y4izus, line 22: Please, add the test for `slackId` and `SlackUsername`

y4izus (changes requested):
Thank you Randy! Please take a look at my comments and ask me for a new review when you finish :)

korgan00 (changes requested):
Good job

components/advocates/AdvocateCard.vue
```diff
@@ -9,10 +9,10 @@
       <Map20 class="advocate-card__icon" />
       {{ location }}
     </p>
-    <p v-if="slackId" class="advocate-card__contact">
+    <p class="advocate-card__contact">
```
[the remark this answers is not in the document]
techtolentino: if we don't have the `slackId`, we're not going to show the entire card... that was the last decision

it _may_ change later, but this is the way we're doing it now.

tests/hooks/advocate-conversion-utils.spec.ts
techtolentino: I think it depends on what we mean as optional.

These properties aren't required in the Airtable, but we also filter out any entries that don't have a `slackId`... so while not _required_ by any hard coded rule, I do think we require them since all advocate objects will have a `slackId` and `slackUsername`

By definition, does that make them required? or should this still be optional?

y4izus: For now, let's leave them as a "check" of only getting the advocates with `slackId`

y4izus (approved):

PR #1127, opened 2020-11-17
migrate from travis ci to github actions

### LINKED ISSUE #1090: Stop using travis CI
With travis's recent billing changes that went into effect this month: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing we're rapidly approach the qiskit project quota for the free trial accounts. This is problematic as both retworkx (for non-x86 jobs) and ibmq provider (for networking reasons) both rely on travis for their CI. We don't want to hit that quota and block development/testing on those projects. So we should try to remove our travis usage for things that dwe can run elsewhere to prioritize access to projects that really need it.

Github actions provides a good alternative that has all the necessary capabilities and native integration with github so we should migrate our CI jobs to use that instead.

vabarbosa:
this PR:

- replaces the Travis CI yaml with GitHub Actions yaml
- configures a (lint, test, and build) workflow job on Node.js 12.x and 14.x
    - runs when action is manually triggered
    - runs when a pull request is made
- configures a (build and deploy) workflow job on Node.js 12.x
    - runs when action is manually triggered
    - runs on pushes to master branch
    - runs daily at 08:00am UTC (on the default/master branch)
- updates build status badge in README

additional change required before this PR should be merged:

- [x] [create secrets](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-an-organization) for the [environment variables](https://github.com/Qiskit/qiskit.org/blob/e0afb4dbca80c7aa128cbdce79962bbcc2686086/.github/workflows/deploy.workflow.yml#L41-L49) defined in the [Travis CI settings](https://www.travis-ci.com/github/Qiskit/qiskit.org/settings)
- [ ] update branch protection rules in Settings for GitHub actions (instead of Travis CI)

Resolves #1090


.github/workflows/deploy.workflow.yml
```diff
@@ -0,0 +1,49 @@
...
+        run: cd deploy && ./deploy_cos.sh
+        shell: bash
+        env:
+          AIRTABLE_API_KEY: ${{ secrets.AIRTABLE_API_KEY }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          IBM_CLOUD_API_KEY: ${{ secrets.IBM_CLOUD_API_KEY }}
+          IBM_CLOUD_ORGANIZATION: ${{ secrets.IBM_CLOUD_ORGANIZATION }}
+          IBM_CLOUD_SPACE: ${{ secrets.IBM_CLOUD_SPACE }}
+          QISKIT_SEGMENT_APP_KEY: ${{ secrets.QISKIT_SEGMENT_APP_KEY }}
+          QISKIT_SEGMENT_SCRIPT: ${{ secrets.QISKIT_SEGMENT_SCRIPT }}
```
delapuente, line 47: I think the codebase does not use these values. Could you double-check before merging? If we are not using them, let's remove them.

delapuente (commented):
Nice done. Contact @y4izus and @mtreinish for setting up the secrets you're lacking.

.github/workflows/deploy.workflow.yml
vabarbosa: removed unused variables

vabarbosa:
- [x]  airtable key has been added (thanks @y4izus)
- [x]  rclone parameters have been added (thanks @mtreinish)

.github/workflows/deploy.workflow.yml
```diff
@@ -0,0 +1,43 @@
...
+      - run: npm ci
+      - run: npm run generate
+
+      - name: Deploy
+        run: cd deploy && ./deploy_cos.sh
+        shell: bash
+        env:
+          AIRTABLE_API_KEY: ${{ secrets.AIRTABLE_API_KEY }}
+          encrypted_rclone_iv: ${{ secrets.encrypted_rclone_iv }}
+          encrypted_rclone_key: ${{ secrets.encrypted_rclone_key }}
```
delapuente, line 43: Secrets names were registered in all-uppercase, although in the script, it uses snake_case names. Let's match each case:
```suggestion
          encrypted_rclone_iv: ${{ secrets.ENCRYPTED_RCLONE_IV }}
          encrypted_rclone_key: ${{ secrets.ENCRYPTED_RCLONE_KEY }}
```

delapuente (changes requested):
Very nice. Thank you for updating the README too.

.github/workflows/deploy.workflow.yml
mtreinish: It doesn't matter they're case insensitive matches. The all caps is forced by the github ui (I didn't actually enter that in). This will work fine as is.

eddybrando (approved):
Awesome, thanks a lot! 😄

vabarbosa:
@delapuente @y4izus one more thing that needs to be checked (which i do not have access to) is the Branch Protection Rules (Settings > Branches). the rules will need to be updated to require the GitHub actions instead of the Travis CI.

delapuente (approved):
Thank you.

delapuente:
> the rules will need to be updated to require the GitHub actions instead of the Travis CI.

This is done, @vabarbosa!

PR #1128, opened 2020-11-18
Global registration of components

### LINKED ISSUE #820: 👍👍👍👍 Consider global registration of the base components
There is a set of components (like `AppLink` or `AppCta`) so widely used that it would be convenient to register them globally instead of importing each time. A good indicator of when a component should be registered globally is that it's a [base component](https://vuejs.org/v2/guide/components-registration.html#Automatic-Global-Registration-of-Base-Components). Our base components are prefixed by `App`.

korgan00:
Fix #820 

This module is used to avoid `import` declaration in the script section and component registration in the `@Component { components: [ ... ] }`

I will sum up what I consider the most interesting features of the `@nuxt/components` package

- **[Dynamic Imports/Lazy Load](https://github.com/nuxt/components#dynamic-imports)**: Lazy load any component. I don't know the tech details about this.
- **[Directory based features](https://github.com/nuxt/components#directories)**: Can be configurated in the nuxt.config. 
  - We can choose a list of files/directories to include ([path](https://github.com/nuxt/components#path), [file extensions](https://github.com/nuxt/components#extensions), [pattern](https://github.com/nuxt/components#pattern)). Default is `components` folder.
  - [Ignore](https://github.com/nuxt/components#ignore) some files in the specified path using a pattern.
  - Establish a [prefix](https://github.com/nuxt/components#prefix) to the components in a specific path. For example: `components/app/button.vue` could be prefixed with `app` to be instanced with `<AppButton>`.
- A new [hook](https://github.com/nuxt/components#library-authors) for creating/importing component libraries.

### DIFF
--- a/components/events/EventCard.vue
+++ b/components/events/EventCard.vue
@@ -23,11 +23,8 @@
 <script lang="ts">
 import Vue from 'vue'
 import { Component, Prop } from 'vue-property-decorator'
-import AppCard from '~/components/ui/AppCard.vue'
 
-@Component({
-  components: { AppCard }
-})
+@Component
 export default class EventCard extends Vue {
   @Prop(Array) types!: any
   @Prop(String) title!: any
--- a/components/events/EventMenu.vue
+++ b/components/events/EventMenu.vue
@@ -23,11 +23,8 @@
 <script lang="ts">
 import Vue from 'vue'
 import { Component } from 'vue-property-decorator'
-import AppLink from '~/components/ui/AppLink.vue'
 
-@Component({
-  components: { AppLink }
-})
+@Component
 export default class EventMenu extends Vue {}
 </script>
 
--- a/components/events/TheEventsHeader.vue
+++ b/components/events/TheEventsHeader.vue
@@ -16,9 +16,7 @@
 <script lang="ts">
 import Vue from 'vue'
 import { Component } from 'vue-property-decorator'
-import ThePageHeader from '~/components/ui/ThePageHeader.vue'
-import TypewriterEffect from '~/components/ui/TypewriterEffect.vue'
 
-@Component({ components: { ThePageHeader, TypewriterEffect } })
+@Component
 export default class TheEventsHeader extends Vue { }
 </script>
--- a/components/landing/TheFeatures/index.vue
+++ b/components/landing/TheFeatures/index.vue
@@ -35,12 +35,8 @@
 import Vue from 'vue'
 
 import { Component } from 'vue-property-decorator'
-import TheFeatureMosaic from '~/components/landing/TheFeatures/TheFeatureMosaic.vue'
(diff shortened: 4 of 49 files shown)


y4izus (changes requested):
In my opinion, this PR should include the registration of the base components and the deletion of the `imports` that are no needed

eddybrando (approved):
Awesome job, thanks a lot @korgan00 🥳 

I only had a concern regarding possible conflicts with non-unique component names, like for example:

- `components/foo/card.vue` for a **Card** component
- `components/bar/card.vue` for another **Card** component

Including a card in a wrapper component would then present an ambiguity issue when including `<Card />` in the template.

But if I understand it correctly, we would then have `<FooCard />` and `<BarCard />` available right? If so, I have now 3 questions:

1. Does this happen automatically or do we have to adjust a configuration every time this ambiguity presents itself?
2. What happens if I then (just to create chaos, I guess) create a third **Card** component  `components/FooCard.vue`. How could I distinguish between `components/FooCard.vue` and `components/foo/Card.vue` in a wrapper component's template?
3. Would any of this benefit from a new discussion regarding our components naming strategy?

korgan00:
As you said, we can create a [prefix](https://github.com/nuxt/components#prefix). The prefix is not natively automatic, but I think we can create that automation.

To ensure we are on the same page, the [prefix](https://github.com/nuxt/components#prefix) is not necessary to be related to the path. We can create a [prefix](https://github.com/nuxt/components#prefix) for the whole project (as carbon uses cv) or even call `BarCard` to `components/foo/card.vue`.

With the PR, we should prevent that chaos, but if we MUST create that component, the default behavior is explained [here](https://github.com/nuxt/components#overwriting-components). We have some possibles workarounds:
 - We could change the prefix to the other or create a new prefix for that new component.
 - We could [Ignore](https://github.com/nuxt/components#ignore) that file and use the classic import/registration.

Probably there are other possible solutions, but I expect not to need them.

korgan00:
Per offline conversations. We decided to extend global registration to all components, not only the base components. I will annotate the pros and cons that we exposed:

**Global registration (pros):**

- The `components` list in the `@Component` decorator will disappear so that most components will have the decorator empty (cleaner files)
- Clean & relevant imports
- Better development experience, avoiding some Git conflicts and unnecessary code

**Global registration (cons):**

- Possible name conflicts as presented in the [comment](https://github.com/Qiskit/qiskit.org/pull/1128#pullrequestreview-537482367) above 

**Cherry-Picked registration (pros):**

- Possibly avoiding name conflicts
- Possibly avoiding bigger than a necessary final bundle. (Seems like the package has [automatic tree-shaking](https://github.com/nuxt/components#library-authors))

**Cherry-Picked registration (cons):**

- More configuration, maintainability, and possible technical debt when upgrading.

y4izus (approved):

PR #1137, opened 2020-11-23
WIP Setup Storybook

### LINKED ISSUE #884: Document components in a storybook
Components number in qiskit.org is getting bigger and it would be convenient to start documenting or cataloguing those in a [Storybook](https://storybook.js.org/docs/vue/get-started/introduction) or alike.

eddybrando:
Setup of [Storybook](https://storybook.js.org/), including 3 example stories.

The [official Storybook integration for Nuxt](https://storybook.nuxtjs.org/) is being used.

`npm run storybook` will now start the Storybook.

These 3 example stories we created to showcase the basic functionality of Storybook in our project. They can and should be extended in the future.

---

### Note for reviewers

If you're testing this locally, don't forget to `rm -rf node_modules && npm install`, since there are some new additions to the dependencies.

---

#### Nuxt < 2.14.0 support
The official **nuxt/storybook** integration has many benefits in simplifying the configuration but also has some issues with Nuxt versions < 2.14.0, like ours.

There is a workaround in `nuxt.config.ts`, as suggested in https://storybook.nuxtjs.org/setup/. We could be able to remove this after upgrading Nuxt.

#### Module resolution of `@storybook/vue`

There is an issue with module resolution of the `@storybook/vue/types-6-0` type definition files, which causes the console to log error messages due to `nuxt/typescript` not being able to resolve those modules.

This may be a Webpack configuration issue related to aliases or just an issue that may be solved with an upgrade of Nuxt (I spend some time looking into it but couldn't solve it yet), but it didn't cause any major interference besides the console messages.

To avoid those, the aforementioned `.d.ts` files are being referenced with relative paths in the stories.

#### Docs add-on not resolving mixins

The documentation add-on causes the console to log error messages regarding not resolvable mixin files. I also looked into this issue but it didn't seem trivial.

Since the documentation generation isn't directly in scope of the Storybook setup issue, I suggest to handle it later in case it causes any issue and it doesn't get solved with an upgrade of our Nuxt version.

There is a comment regarding this in `.storybook/main.js`.

---

Closes #884


eddybrando:
Replaced by https://github.com/Qiskit/qiskit.org/pull/1143

Issue #1150, opened 2020-11-25
Setup official GitHub Dependabot


abdonrd:
Setup official GitHub Dependabot.

https://docs.github.com/en/github/administering-a-repository/about-github-dependabot-version-updates

> GitHub Dependabot takes the effort out of maintaining your dependencies. You can use it to ensure that your repository automatically keeps up with the latest releases of the packages and applications it depends on.

Then we will have a new tab `Dependabot` in the [`Insights > Dependency graph`](https://github.com/Qiskit/qiskit.org/network/dependencies) here:

https://github.com/Qiskit/qiskit.org/network/updates

abdonrd:
First I will try to update the dependencies manually so as not to start with too many PRs from the bot.

PR #1153, opened 2020-11-25
Update ESLint dependencies

abdonrd:
To simplify #1150


constants/appLinks.ts
```diff
@@ -5,6 +5,7 @@ type SegmentData = {
 /**
  * Represent a navigation link for the social media community links of the site
  */
+// eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
```
y4izus, line 8: We are not using this interface anymore, so please, remove it instead of adding the es-lint comment

y4izus (changes requested):
Thank you!

Just one small change :)

constants/appLinks.ts
abdonrd: Perfect! As I am not very familiar with the project, I preferred not to delete anything.

y4izus: No worries! For me it is the same and I'm familiar with it, but...

Screen Shot 2020-11-26 at 16 59 37

y4izus (approved):

PR #1173, opened 2020-11-30
Parallelize jobs in GitHub Actions

### LINKED ISSUE #1172: Parallelize jobs in GitHub Actions
I think we could parallelize the GitHub Actions jobs; we would save a little time and the logs would be somewhat clearer.

abdonrd:
Fix #1172

If we merge it, we should [update the **`Required`** checks](https://github.com/Qiskit/qiskit.org/settings/branches):

Screenshot 2020-12-01 at 12 25 21

### DIFF
--- a/.github/workflows/test.workflow.yml
+++ b/.github/workflows/test.workflow.yml
@@ -1,14 +1,26 @@
-name: lint, test, and build
+name: CI
 
 # run when action is manually triggered
 # or when a pull request is made
 on: [workflow_dispatch, pull_request]
 
 jobs:
-  # job id/key
-  lint-test-and-build:
-    # job name
-    name: Lint, Test, and Build
+  lint:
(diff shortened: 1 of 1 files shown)


korgan00 (approved):
Sounds good ;)

delapuente:
@abdonrd I'm assuming we are removing the cache mechanism cause we do not longer need to reuse the same `npm_folder`, right? Also, why do not we lint for node 12?

abdonrd:
> @abdonrd I'm assuming we are removing the cache mechanism cause we do not longer need to reuse the same `npm_folder`, right? Also, why do not we lint for node 12?

Yes!

About the linting, I have assumed doing it on a single node version is enough.

delapuente (commented):
Nice done. Thank you.

PR #1201, opened 2020-12-03
Create Storybook showcase story

### LINKED ISSUE #1155: Create Storybook story to showcase functionalities
Creation of a Storybook story for at least one component where we can showcase all (or at least most of) the functionalities

---

Part of https://github.com/Qiskit/qiskit.org/issues/884#issuecomment-733034345

eddybrando:
**Note:** This is only one part of https://github.com/Qiskit/qiskit.org/issues/884#issuecomment-733034345

---

Make the necessary adjustments to the **AppCard** Storybook story to showcase the various Storybook functionalities.

Only the card's image is currently not showing due to _lazy-background_ not loading the image. I tried to dig deeper into this but couldn't make it work. But I personally believe that this should not be an impediment for the scope of this PR.

---

Closes #1155

### DIFF
--- a/components/ui/AppCard.stories.ts
+++ b/components/ui/AppCard.stories.ts
@@ -15,7 +15,10 @@ export default {
 const Template: Story = (args, { argTypes }) => ({
   components: { AppCard },
   props: Object.keys(argTypes),
-  template: '<AppCard :cta-label="ctaLabel" :image="image" :tags="tags" :title="title" :to="to">{{ vSlot }}</AppCard>'
+  template: `
+    <AppCard :cta-label="ctaLabel" :image="image" :tags="tags" :title="title" :to="to">
+      {{slotDefault}}
+    </AppCard>`
 })
 
 export const Default = Template.bind({})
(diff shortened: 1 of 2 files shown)


techtolentino (approved):
This looks great, @eddybrando 

I agree with your note on the lazy-load issue.

components/ui/AppCard.stories.ts
```diff
@@ -15,15 +15,18 @@ export default {
 const Template: Story = (args, { argTypes }) => ({
   components: { AppCard },
   props: Object.keys(argTypes),
-  template: '<AppCard :cta-label="ctaLabel" :image="image" :tags="tags" :title="title" :to="to">{{ vSlot }}</AppCard>'
+  template: `
+    <AppCard :cta-label="ctaLabel" :image="image" :tags="tags" :title="title" :to="to">
+      {{slotDefault}}
```
techtolentino, line 20: just curious where `slotDefault` change came from... is it for readability?

y4izus: I think so, @techtolentino

y4izus (approved):

components/ui/AppCard.stories.ts
eddybrando: @techtolentino , it's an "arg" that gets registered in the component so that we can later change it in the Storybook story dynamically.

This way we can the easily test how the component reacts with no text, some text and a lot of text for example. 🙂

PR #1203, opened 2020-12-03
Updating items on Infinite scroll now update observed data for ScrollSectionsMixin

korgan00:


eddybrando (changes requested):
Thanks for the update @korgan00, but I’m afraid I could still see some issues when integrating this change to the advocates filter.

### Current behaviour:

- Checking “Australia” and “Africa” displays 2 advocates.

### Expected behaviour:

- Checking “Australia” and “Africa” should display 8 advocates (after some navigation).

### Steps to reproduce:

1. Check “Australia”
2. Navigate to the “events” page
3. Navigate to the “advocates” page
4. Check also “Africa”

### Additional information:

The Vuex store is calculating the 8 filtered advocates and passing these to the child component to render the cards properly.

korgan00:
> Thanks for the update @korgan00, but I’m afraid I could still see some issues when integrating this change to the advocates filter.
> 
> ### Current behaviour:
> * Checking “Australia” and “Africa” displays 2 advocates.
> 
> ### Expected behaviour:
> * Checking “Australia” and “Africa” should display 8 advocates (after some navigation).
> 
> ### Steps to reproduce:
> 1. Check “Australia”
> 2. Navigate to the “events” page
> 3. Navigate to the “advocates” page
> 4. Check also “Africa”
> 
> ### Additional information:
> The Vuex store is calculating the 8 filtered advocates and passing these to the child component to render the cards properly.

Good catch! I made some other variables update. You can try it again.

eddybrando (approved):
Thank you @korgan00! 🙂 

I just gave it a try integrating it with the new advocates filter and the issue that I reported before seems to be gone.

Nice work 👍

y4izus (approved):
Thank you very much!

PR #1244, opened 2020-12-15
Register "pre-requisite" dynamic components for homepage

### LINKED ISSUE #1219: homepage Start locally toggle
- Navigate to qiskit.org > Quick Start section > Start locally
- select `Unstable`
- see the part that says _'Building Qiskit requires a C++ compiler and development headers'_

The copy is within a accordion/toggle component

Maybe this copy should stand on its own? Unless there really needs to be a toggling experience?

Screen Shot 2020-12-04 at 7 02 32 AM

eddybrando:
After the introduction of the global registration of components in #820 with PR #1128, some homepage components weren't registered.

This happened becase the `@nuxt/components` module doesn't automatically import dynamic components.

In order to be able to lazy load these components, we need to move them to the `components/global` directory, as pointed out by their documentation: https://github.com/nuxt/components#dynamic-components.

---

### Before:

ezgif-6-6beae3642267

image

### After:

ezgif-6-e6c3ba2a8346

---

Closes #1219


eddybrando:
After discussing with @korgan00 offline, we agreed that it would be better to avoid moving the components to the `components/global` directory and instead import and register them manually in the component that uses them.

I'm going to change this PR to comply with the aforementioned and I'll continue the reviewing process after that.

korgan00:
I reviewed this, and I think that modifying the current folder structure is a global change for a local problem.
I will suggest registering those items using manual import/registration into `StartLocally`. If we delete this section, this patch will disappear.

```
import Vue from 'vue'
import PrerequisitesForLinux from '~/components/landing/TheQuickStart/PrerequisitesForLinux.vue'
import PrerequisitesForMac from '~/components/landing/TheQuickStart/PrerequisitesForMac.vue'
import PrerequisitesForWindows from '~/components/landing/TheQuickStart/PrerequisitesForWindows.vue'

type ChoicesGroup = {
  title: string,
  id: string,
  options: Array<string>
}
type InstallChoices = Array<ChoicesGroup>

@Component({
  components: {
    PrerequisitesForLinux,
    PrerequisitesForMac,
    PrerequisitesForWindows
  }
})
```

eddybrando:
I inspected the registered components and made some interesting findings in our various scenarios and ways to handle this issue.

### Manual import in components

When registering the components manually in the `StartLocally` component, we have them all available from the very beginning. This means that they are included in the initial `app.js` JS bundle.

### Global registration via `components/global/`

The same is true for this strategy. The components are included in the initial bundle.

### Lazy loading with `lazy-` prefix

The JS for the components in loaded on demand, reducing the size of the initial JS bundle.

## "Real world" repercussions

Ideally the website should initially load only the JS required so that the time until the user can see and interact with the website is reduced as much as possible. This is, of course, quite a high goal and may not always be worth other development tradeoffs.

The fact that loading components that may never be used happens here on the homepage for instance could be worth analysing. But this analysis should be broader and maybe even include rethinking the strategy in which the footer loads for instance.

As of this particular case, it may be too much trouble for too little potential improvements.

eddybrando:
@korgan00 I updated the PR with the [lazy imports](https://github.com/nuxt/components#lazy-imports) approach.

I believe it may be even cleaner and more straight forward, but please take another look at it and let me know if you prefer the manual imports and registration over this. 🙂

techtolentino (approved):
This looks great, @eddybrando -- and thank you for including the reference to the docs... Learned something new!

delapuente:
Thank you @eddybrando and nice catch @techtolentino! I'm merging this.

PR #1271, opened 2021-01-05
feat: use Vuex modules

### LINKED ISSUE #1058: Use module store model instead of classic store model
We are using [classic store model](https://nuxtjs.org/guide/vuex-store/#classic-mode), but it's deprecated and will disappear with Nuxtjs 3. We have to change to the [module store model ](https://nuxtjs.org/guide/vuex-store/#modules-mode) to avoid problems when we move to the new Nuxtjs version

eddybrando:
This PR replaces the classic Vuex store model (which is deprecated) with Vuex modules.

There is [some magic in the way NuxtJS configures Vuex](https://nuxtjs.org/docs/2.x/directory-structure/store#modules), so the main part of this task revolved around moving the files in the `/store` directory, and then also adding some namespaces.

In addition to this, some changes to the store tests where needed to create mock store with this new structure.

---

Closes #1058


y4izus (approved):
Awesome! Thank you!

PR #1301, opened 2021-01-12
feat: create "what is this event" section for Seminar Series

### LINKED ISSUE #1278: Implement section of "What is the Quantum Information Science Seminar Series?" for single event page
This section could be found [on this Figma file](https://www.figma.com/file/nqFBhDOUq6Dknbw9pM92AE/1263-individual-event-pages?node-id=0%3A1). You can see the detail on this image:
Screen Shot 2021-01-08 at 10 30 20

As you can see, it has a title and the mosaic component that we're already using on the [home page](https://qiskit.org/)

Objectives:
- [x] Once #1279 is finished, add a section for "What is the Quantum Information Science Seminar Series?"
- [x] Once the refactoring is done (issue #628), create a reusable component for the mosaic feature
- [x] Reuse the component for both the home page and individual event page

For now, use the texts and images that are on Figma but have in mind that they could change

eddybrando:
This PR creates the "What is this event" section for the Seminar Series page.

Some content needs to change and some images may also change as well, but that will be done later, once we have those assets and information.

To preview the page in the Vercel deployment, navigate to `/events/seminar-series`.

Screenshot 2021-01-12 at 14 25 52

---

Closes #1278


eddybrando:
As discussed offline, I'll extract the section to its own component and change the naming of the CSS classes from "events-page" to "seminar-series-page" to be more precise.

components/events/single-events/WhatIsThisEvent.vue
```diff
@@ -0,0 +1,22 @@
...
+  </section>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+import { Component, Prop } from 'vue-property-decorator'
+import { MosaicElement } from '~/components/ui/AppMosaic.vue'
+
+@Component
+export default class WhatIsThisEvent extends Vue {
```
techtolentino, line 14: I'm curious why we don't use `EventTitle` or something of that nature?

components/ui/AppMosaic.vue
```diff
@@ -159,12 +159,8 @@ export default class AppMosaic extends Vue {
 
       &_fourth {
         background-position: right bottom;
-        background-size: 12rem auto;
+        background-size: contain;
```
techtolentino, line 162: I like this update, thanks!

techtolentino (commented):
Thanks for this @eddybrando - it looks great! 

I just have a question around the component name...

components/events/single-events/WhatIsThisEvent.vue
eddybrando: Sure, it's because this component isn't only the title but also the mosaic.

So, everything inside the red square:

104320063-1a913c80-54e2-11eb-916c-ba228cfd339e

eddybrando:
As discussed offline, I'll refactor the "What is this event" component a little bit. My original idea was to make it generic and reusable by other "single event" pages, but since we don't have the final designs for those, the correct level of abstraction would only be a guess. Maybe this component will not be reused at all, even.

So, I'll make it more specific to the "seminar series" page, and also adjust the name a little bit to make it more descriptive by appending "Section" to it.

y4izus (approved):
Great Eddy!! Thank you very much!!

PR #1330, opened 2021-01-19
fix: use factory functions for default value of Arrays

eddybrando:


y4izus (approved):
Good point Eddy!! thank you very much for doing this!

PR #1371, opened 2021-02-02
refactor: next event from events json

eddybrando:
With this PR, the **next Seminar Series event** will be extracted from the list of **upcoming events** on the **Header** component directly, instead of being stored once as a JSON file with a hook event.


y4izus (approved):
Much more readable now! Thank you!

PR #1372, opened 2021-02-02
fix: missing helpful resources links

eddybrando:
This PR adds 2 missing links for the **Helpful Resources** section of the Seminar Series page.

It also includes a minor copy fix in the **Stay informed** block.


y4izus (approved):
Good catch on the copy review! Thank you!

Issue #1183, opened 2020-12-01
Implement mobile navigation menu


techtolentino:
This is the implementation story for the **mobile version** of the nav updates: https://github.com/Qiskit/qiskit.org/issues/862

Acceptance criteria
- [ ] Navigation looks like the Figma file.
- [ ] Navigation displays Events and Advocates links.

Figma file: https://www.figma.com/file/PDbN8wcg5WteWVvinjAEUv/qiskit.org-main--file?node-id=1%3A9533

PR #1373, opened 2021-02-02
Fix mobile-menu footer breakpoint

techtolentino:
This update didn't make the final push, as a part of #1183, but fixes the breakpoint for when the mobile-menu footer should appear.

Adding screen capture to show fix in place:

https://user-images.githubusercontent.com/6276074/106607418-85052d80-6520-11eb-9401-6a6884934ebd.mov


y4izus (approved):
Much better now Randy!! thank you for the improvement!

Issue #1443, opened 2021-02-17
Link to new demo Textbook page


eddybrando:
In order to simplify the manual testing process of the new demo Textbook page, we need to link that page in the main navbar.

**Task:** Create a link in the navbar that points to the demo Textbook page.

PR #1444, opened 2021-02-17
feat: add link to "Textbook demo"

eddybrando:
This PR adds a link in the navbar that points to the demo Textbook page.

This is only needed during development for testing purposes.

**Note:** There is a small design bug in the _big screen_ view (the margin between "Learn" and "Textbook" is small, and the margin between "Textbook" and "Community" is big) due to a CSS rule targeting a _nth:child_. But for the purpose of this issue (just having the link to test the page easily), it can be ignored.

Screenshot 2021-02-17 at 13 23 25

---

Closes https://github.com/Qiskit/qiskit.org/issues/1443

### DIFF
--- a/constants/menuLinks.ts
+++ b/constants/menuLinks.ts
@@ -122,6 +122,14 @@ const LEARN_LINK: NavLink = {
   }
 }
 
+const TEXTBOOK_DEMO_LINK: NavLink = {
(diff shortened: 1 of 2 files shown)


y4izus (approved):
Thank you!

PR #1490, opened 2021-03-02
Improves basic link

y4izus:
Check if the `url` exists before using it to avoid problems when the `url` is not present


eddybrando (approved):
Thanks @y4izus! This are very useful checks 🙂

PR #1491, opened 2021-03-02
feat: check Seminar Series event's values

eddybrando:
Check the Airtable Seminar Series Events entries to ensure that only those with all the values are included in the content JSON files.

This is needed to avoid trying to perform invalid operations on unexpected data structures (i.e. with missing data) that can cause runtime issues.


y4izus (approved):
Thank you!

PR #1512, opened 2021-03-03
refactor: create constants for segment actions

### LINKED ISSUE #1511: Create reusable definitions of segments actions' names for tracking
The tracking segment actions consist of long strings which can be broken down into reusable pieces in the form of constants.

This will help with future maintenance as well as providing a general overview of the labels we're using across the project.

eddybrando:
Closes #1511


eddybrando:
We'll go with the approach suggested in https://github.com/Qiskit/qiskit.org/issues/1511#issuecomment-862284188, which will require different code changes.

Issue #1453, opened 2021-02-18
Update content of the "Start Learning" section in the new Textbook landing page


eddybrando:
Once we have the content for the "Start Learning" section in the new Textbook landing page, we need to update it.

### Tasks:

- [x] Wait for #1435 to be done
- [x] Define final content (copy and links)
- [x] Update content
- [x] Add "segments" for link events tracking

Issue #1448, opened 2021-02-17
Update content of the "Helpful Resources" section in the new Textbook landing page


eddybrando:
Once we have the content for the "Helpful Resources" section in the new Textbook landing page, we need to update it.

### Tasks:

- [x] Wait for #1447 to be done
- [x] Define final content (copy and links)
- [x] Update content
- [x] Add "segments" for link events tracking

PR #1521, opened 2021-03-11
Updates copy and images on textbook demo home

y4izus:
Start fixing #1453 and #1448. We don't have all the info, so the issue is not complete 100%. Anyways, we need merge this


korgan00 (approved):
Good job!

PR #1525, opened 2021-03-11
feat: emit "mouseenter" event

eddybrando:
Propagate emission of the _mouseenter_ event to allow for more mouse interactions while still leveraging the tracking functionalities of this component.


y4izus (approved):
Thanks

PR #1561, opened 2021-03-17
feat: add segment to Introduction Course header

### LINKED ISSUE #1549: Learning Path LP: Add "segment" to header CTA

eddybrando:
Closes #1549

### DIFF
--- a/components/textbook-demo/IntroductionCourseHeader.vue
+++ b/components/textbook-demo/IntroductionCourseHeader.vue
@@ -36,10 +36,7 @@
             </p>
           </div>
         </div>
-        <AppCta
-          label="Start learning"
-          url="https://project-platypus-default.apis-dev-5dd5718798d097eccc65fac4e78a33ce-0000.us-south.containers.appdomain.cloud/course/learning-states-and-qubits/introduction"
-        />
+        <AppCta v-bind="startLearningCTA" />
       </div>
 
       <div class="introduction-course-header__aside">
@@ -55,9 +52,12 @@
 <script lang="ts">
 import Vue from 'vue'
 import { Component } from 'vue-property-decorator'
+import { TEXTBOOK_DEMO_INTRODUCTION_COURSE_START_LEARNING } from '~/constants/appLinks'
 
 @Component
-export default class AppPageHeader extends Vue {}
+export default class AppPageHeader extends Vue {
+  startLearningCTA = TEXTBOOK_DEMO_INTRODUCTION_COURSE_START_LEARNING
+}
 </script>
 
 <style lang="scss" scoped>
--- a/components/textbook-demo/IntroductionCourseHeader.vue
+++ b/components/textbook-demo/IntroductionCourseHeader.vue
@@ -55,7 +55,7 @@ import { Component } from 'vue-property-decorator'
 import { TEXTBOOK_DEMO_INTRODUCTION_COURSE_START_LEARNING } from '~/constants/appLinks'
 
 @Component
-export default class AppPageHeader extends Vue {
+export default class IntroductionCourseHeader extends Vue {
   startLearningCTA = TEXTBOOK_DEMO_INTRODUCTION_COURSE_START_LEARNING
 }
 </script>
--- a/components/textbook-demo/IntroductionCourseHeader.vue
+++ b/components/textbook-demo/IntroductionCourseHeader.vue
@@ -51,13 +51,15 @@
 
 <script lang="ts">
 import Vue from 'vue'
-import { Component } from 'vue-property-decorator'
-import { GeneralLink, textbookDemoIntroductionCourseStartLearningUrl } from '~/constants/appLinks'
+import { Component, Prop } from 'vue-property-decorator'
+import { GeneralLink } from '~/constants/appLinks'
 
 @Component
 export default class IntroductionCourseHeader extends Vue {
+  @Prop({ type: String, required: true }) courseStartUrl!: string
+
   startLearningCTA: GeneralLink = {
-    url: textbookDemoIntroductionCourseStartLearningUrl,
+    url: this.courseStartUrl,
     label: 'Start learning',
     segment: {
       action: 'introduction-course > header > introduction'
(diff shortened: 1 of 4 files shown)


constants/appLinks.ts
```diff
@@ -77,6 +77,14 @@ const TEXTBOOK_DEMO_START_LEARNING: GeneralLink = {
   }
 }
 
+const TEXTBOOK_DEMO_INTRODUCTION_COURSE_START_LEARNING: GeneralLink = {
+  url: 'https://project-platypus-default.apis-dev-5dd5718798d097eccc65fac4e78a33ce-0000.us-south.containers.appdomain.cloud/course/learning-states-and-qubits/introduction',
+  label: 'Start learning',
+  segment: {
+    action: 'introduction-course > header > start-learning'
```
y4izus, line 84: This CTA point to the same page than the first link on the learning paths section, so they should have the same text at the end of the action. Replace `introduction-course > header > start-learning` with `introduction-course > header > introduction`

constants/appLinks.ts
```diff
@@ -92,6 +100,7 @@ export {
   REQUEST_AN_EVENT_CTA,
   SEMINAR_SERIES_ALL_EPISODES_CTA,
   SEMINAR_SERIES_FULL_ARCHIVE_CTA,
+  TEXTBOOK_DEMO_INTRODUCTION_COURSE_START_LEARNING,
```
y4izus, line 103: We use this link in two places but changing the label, please, find a way to reuse the rest instead of creating a new one

y4izus (changes requested):
Thank you! Please, take a look at my comments

constants/appLinks.ts
eddybrando: As discussed offline, I created this related refactoring issue: https://github.com/Qiskit/qiskit.org/issues/1566

constants/appLinks.ts
```diff
@@ -77,6 +79,16 @@ const TEXTBOOK_DEMO_START_LEARNING: GeneralLink = {
   }
 }
 
+const textbookDemoIntroductionCourseStartLearningUrl = 'https://project-platypus-default.apis-dev-5dd5718798d097eccc65fac4e78a33ce-0000.us-south.containers.appdomain.cloud/course/learning-states-and-qubits/introduction'
+
+const TEXTBOOK_DEMO_INTRODUCTION_COURSE_START_LEARNING: GeneralLink = {
```
y4izus, line 84: Please, move this link to the component you use it

constants/appLinks.ts
```diff
@@ -77,6 +79,16 @@ const TEXTBOOK_DEMO_START_LEARNING: GeneralLink = {
   }
 }
 
+const textbookDemoIntroductionCourseStartLearningUrl = 'https://project-platypus-default.apis-dev-5dd5718798d097eccc65fac4e78a33ce-0000.us-south.containers.appdomain.cloud/course/learning-states-and-qubits/introduction'
```
y4izus, line 82: We should move this url to a better place to be close to where it's used, ideas? In addition, you can change the name to `introductionLearningPathIntroductionSectionURL` or something similar to be more clear

y4izus (changes requested):
Thank you again :). Please, take a look to my comments

constants/appLinks.ts
eddybrando: Lots of ideas 😄  But as discussed, I thought we were going to agree on the proper way to handle this going forward with https://github.com/Qiskit/qiskit.org/issues/1566.

constants/appLinks.ts
eddybrando: Are we starting to do this for all cases? I thought we were delaying this until https://github.com/Qiskit/qiskit.org/issues/1566. 😕

components/textbook-demo/IntroductionCourseHeader.vue
```diff
@@ -55,9 +52,12 @@
 <script lang="ts">
 import Vue from 'vue'
 import { Component } from 'vue-property-decorator'
+import { TEXTBOOK_DEMO_INTRODUCTION_COURSE_START_LEARNING } from '~/constants/appLinks'
 
 @Component
-export default class AppPageHeader extends Vue {}
+export default class AppPageHeader extends Vue {
```
y4izus, line 58: The name of the class should be `IntroductionCourseHeader`

components/textbook-demo/IntroductionCourseHeader.vue
```diff
@@ -55,9 +52,12 @@
 <script lang="ts">
 import Vue from 'vue'
 import { Component } from 'vue-property-decorator'
+import { TEXTBOOK_DEMO_INTRODUCTION_COURSE_START_LEARNING } from '~/constants/appLinks'
```
y4izus, line 55: Move this link to this component

components/textbook-demo/LearningPathPagesSection.vue
```diff
@@ -52,7 +52,7 @@
 <script lang="ts">
 import Vue from 'vue'
 import { Component } from 'vue-property-decorator'
-import { SegmentData } from '~/constants/appLinks'
+import { SegmentData, textbookDemoIntroductionCourseStartLearningUrl } from '~/constants/appLinks'
```
y4izus, line 55: How do you think if we create a constants file on a textbook-demo folder to add the urls we are going to reuse? Right now we only have one, but if we continue developing learning path we will have more

y4izus (changes requested):
Ideas for the links and one more change :)

components/textbook-demo/LearningPathPagesSection.vue
eddybrando: I like the idea in general, but I have some concerns about creating a new pattern for one single use case (as for now), because there is potential of forgetting that pattern later.

After some more thinking, I wanted to suggest defining the URL in the parent component (`pages/textbook-demo/learning-paths/introduction-course.vue`) and passing it as a prop named something similar to `courseStartUrl`, which may be generic enough to make sense for both components, and also for future path pages that may reuse this template.

What do you think? 🙂

eddybrando: As discussed offline, we'll move the URL to appLinks until we find a global way of handling the links with https://github.com/Qiskit/qiskit.org/issues/1566.

y4izus (approved):
Great! Thank you Eddy!

PR #1579, opened 2021-03-23
Remove @abdonrd as code owner

abdonrd:
I still watch the repository (so I get GitHub notifications from each PR).
But since I'm not working daily on qiskit.org, I avoid receiving emails from each PR.


eddybrando (approved):
LGTM!

Issue #965, opened 2020-10-02
❤️👍👍👍   Use Carbon themes and customize it to match visual identity


techtolentino:
Since all of the pages have moved to the light theme, we need to modify the `community-theme.scss` file to match our visual branding.

This work is the follow-up for the suggestion mentioned [here](https://github.com/Qiskit/qiskit.org/pull/951#discussion_r495666584).

korgan00:
I think this may be related to #703

eddybrando:
We are currently importing some styles in the layouts to overwrite the dark styles, which are no longer the default. So we should reorder this.

Also, this is a good opportunity to look deeper into the copy / typography styles that need to be imported individually in various components to avoid FOUT, as suggested offline by @y4izus.

delapuente:
>> Also, this is a good opportunity to look deeper into the copy / typography styles that need to be imported individually in various components to avoid FOUT, as suggested offline by @y4izus.

Is there an issue for this specific part?

PR #1588, opened 2021-03-23
fix: display correct layout styles

### LINKED ISSUE #1585: Textbook demo: styles break when accessing page directly

Go to https://qiskitorg-git-feature-textbook-demo-qiskit.vercel.app/textbook-demo directly.

The page has some broken styles:

Screenshot 2021-03-23 at 15 07 21

The page should look the way it looks when loading it through navigation (starting in https://qiskitorg-git-feature-textbook-demo-qiskit.vercel.app and clicking on the menu link that redirects to the Textbook Demo page): 

Screenshot 2021-03-23 at 15 07 43

eddybrando:
Closes #1585

### DIFF
--- a/layouts/default-textbook.vue
+++ b/layouts/default-textbook.vue
@@ -18,3 +18,7 @@ export default class DefaultTextbookLayout extends Vue {
   isMenuShown: boolean = false
 }
 </script>
+
+<style lang="scss">
+@import '~/assets/scss/layout.scss';
+</style>


y4izus (approved):
Thank you Eddy! I like this as "quick solution".

The problem behind is that the visual identity has changed form a dark to a light mode, so we need to rethink the css. [This is the issue related](https://github.com/Qiskit/qiskit.org/issues/965) to this

PR #1605, opened 2021-03-29
style: avoid using `this` in template

eddybrando:
Fixes https://eslint.vuejs.org/rules/this-in-template.html to prevent https://github.com/Qiskit/qiskit.org/pull/1598 CI pipeline's lint stage from failing


components/ui/PageSection.vue
```diff
@@ -10,13 +10,13 @@
       <div
         class="page-section__main-content"
         :class="{
-          'page-section__main-content_alone': !this.$slots.aside
+          'page-section__main-content_alone': !$slots.aside
```
y4izus, line 13: I think `PageSection` is not used anymore. Could you double check and if that's the case, delete this file?

y4izus (approved):
Thank you very much for doing this!! just one check and we can merge :)

components/ui/PageSection.vue
eddybrando: Good catch!

I found in total 4 components not being used currently:

- components/ui/AppIcon.vue
- components/ui/MdContent.vue
- components/ui/PageSection.vue
- components/ui/SoftwareStack.vue

I deleted those as well 🙂

y4izus: Great!! thank you!!

y4izus (approved):

PR #1606, opened 2021-03-29
feat: create Photon skeleton

### LINKED ISSUE #1599: Photon: Create page skeleton
Create the page main structure in a new feature branch.

eddybrando:
This PR create the skeleton of the Photon LP.

The header isn't added yet since there is currently a PR in review with the new header components.

---

Closes #1599

### DIFF
--- a/pages/photon.vue
+++ b/pages/photon.vue
@@ -1,10 +1,14 @@
 <template>
   <div>
     <header>Header</header>
-    <section class="photon__section">
+    <section class="photon__white-section">
       <h2 class="copy__title">Lorem ipsum</h2>
+      <div class="bx--row">
+        <div class="bx--col-lg-6">left</div>
+        <div class="bx--col-lg-6">right</div>
+      </div>
     </section>
-    <section class="photon__get-it-section">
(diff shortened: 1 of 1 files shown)


pages/photon.vue
```diff
@@ -1,19 +1,222 @@
...
-  <div>Content goes here</div>
+  <div>
+    <header>Header</header>
+    <section class="photon__section photon__section_light">
+      <div class="photon__container">
+        <h2 class="copy__title">
+          Lorem ipsum
+        </h2>
+        <div class="bx--row">
+          <div class="bx--col-lg-6 section-mobile-spacing">
```
y4izus, line 10: Could you tried with `bx--col-md-4` and give me your opinion. I think it could look better for medium sizes and I would like to know if you agree :)

y4izus (commented):
It's a great start point! I like the use of the carbon grid system, thanks for use it here :)

I just have one comment

pages/photon.vue
eddybrando: I like that 👍

PR #1663, opened 2021-04-06
Fix table header spacing

techtolentino:
Small style update to address table header spacing.

Fixes this bug: 👇 
Screen Shot 2021-04-06 at 12 57 27 PM

### DIFF
--- a/components/ui/AppDataTable.vue
+++ b/components/ui/AppDataTable.vue
@@ -53,10 +53,7 @@ export default class AppDataTable extends Vue {
 .bx--data-table th {
   color: $black-100;
   background-color: $cool-gray-20;
-
-  &[aria-sort] {
-    padding: $spacing-05;
-  }
+  padding: $spacing-05;
 }
 
 .bx--data-table tbody tr td, .bx--data-table tbody tr:hover td {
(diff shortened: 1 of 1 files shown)


components/ui/AppDataTable.vue
```diff
@@ -53,6 +53,10 @@ export default class AppDataTable extends Vue {
 .bx--data-table th {
   color: $black-100;
   background-color: $cool-gray-20;
+
+  &[aria-sort] {
```
y4izus, line 57: As you can read [here](https://www.w3.org/TR/wai-aria-practices-1.1/examples/grid/dataGrids.html), `aria-sort` is an attribute that gives sortable capabilities information about the column. That is why I think this is not the best place to add the padding. Instead of this, you can add the `padding` line inside the `.bx--data-table th` definition

y4izus (changes requested):
Thank you Randy! could you take a look at my comment?

components/ui/AppDataTable.vue
techtolentino: @y4izus - I tried adding the padding there first, but it was getting overwritten Carbon's style... 
Screen Shot 2021-04-07 at 7 44 00 AM (2)


and I didn't want to use `!important`.

Thoughts?

techtolentino: @y4izus - issue created here:

https://github.com/carbon-design-system/carbon-components-vue/issues/1161

y4izus: thank you very much!

y4izus (approved):
Thanks!

PR #1665, opened 2021-04-07
Removes min size for accordion content

y4izus:
Right now the accordion component has min-size set to 30.5rem. The consequence is that when we have small content, we can see a bigger space between the link and the bottom of the accordion element:
Captura de pantalla 2021-04-07 a las 10 51 18

This PR removes that min-size, so the accordion size change following the content size:
 
Captura de pantalla 2021-04-07 a las 10 53 21


eddybrando (approved):
LGTM!

PR #1670, opened 2021-04-08
Add summer school page content

techtolentino:
- updated title
- updated mosaic
- Add Tabs and corresponding Agenda DataTables
- Added FAQ section

PR #1678, opened 2021-04-13
Create summer school page (#1670)

korgan00:
For review purpose

To see the page go to `/events/summer-school`


eddybrando:
Thanks a lot for the PR, I'll take a look at it 🙂 

Please remove the changes made to the `package-lock.json` file

new-content/events/summer-school-2021.yaml
```diff
@@ -0,0 +1,259 @@
+metadata:
+  title: 'Qiskit Global Summer School 2021'
+  description: 'The Qiskit Global Summer School 2021 is a two-week intensive summer school designed to empower the next generation of quantum researchers and developers with the skills and know-how to explore quantum applications on their own'
+  image: '/images/events/physics-of-computation/card-image-endicott-house-1981.png'
+
+header:
+  primaryTitle: 'Qiskit Global Summer School 2021:'
```
eddybrando, line 7: IMO calling these "primary" and "secondary" titles is a little confusing, since it implies some kind of either hierarchy or importance, but in this case it's all about presentation.

Probable "titleLine1" and "titleLine2" of something like that might be more straight forward.

pages/events/summer-school.vue
```diff
@@ -0,0 +1,288 @@
...
+    <AppMosaicSection
+      class="summer-school-page__section"
+      :title="eventData.mosaic.title"
+      :elements="eventData.mosaic.elements"
+    />
+
+    <section class="summer-school-page__section">
+      <h2 class="copy__title">
+        {{ eventData.agenda.title }}
+      </h2>
```
eddybrando, line 29: Just as a _nice to have_, we could go for a one-liner here:

```suggestion
      <h2 class="copy__title" v-text="eventData.agenda.title" />
```

pages/events/summer-school.vue
```diff
@@ -0,0 +1,288 @@
...
+      <client-only>
```
eddybrando, line 30: I just tested removing the `client-only` wrapper and it seems to work fine. Can we remove it from here or is there a reason why we need it?

pages/events/summer-school.vue
```diff
@@ -0,0 +1,288 @@
...
+        <cv-tabs>
+          <cv-tab :label="eventData.agenda.week1.tabName">
+            <AppDataTableSection
+              class="summer-school-page__section"
+              :section-title="''"
+              :data-table-columns="agendaColumnsDataTable"
+              :data-table-elements="eventData.agenda.week1.tableData"
+            />
```
eddybrando, line 38: As I understand, we're setting `:section-title="''"` in order to remove the title from the **AppDataTableSection** component.

Since the **AppDataTableSection** component is basically just the **AppDataTable** component plus a title, I think that I'd make more sense to use the **AppDataTable** component here directly, like so:

```vue
            <AppDataTable
              :columns="agendaColumnsDataTable"
              :elements="eventData.agenda.week1.tableData"
            />
```

Some margin may be needed on top.

pages/events/summer-school.vue
```diff
@@ -0,0 +1,288 @@
...
+          </cv-tab>
+          <cv-tab :label="eventData.agenda.week2.tabName">
+            <AppDataTableSection
+              class="summer-school-page__section"
+              :section-title="''"
+              :data-table-columns="agendaColumnsDataTable"
+              :data-table-elements="eventData.agenda.week2.tableData"
+            />
```
eddybrando, line 46: Same as https://github.com/Qiskit/qiskit.org/pull/1678/files#r614989225

pages/events/summer-school.vue
```diff
@@ -0,0 +1,288 @@
...
+          </cv-tab>
+        </cv-tabs>
+      </client-only>
+    </section>
+    <section class="summer-school-page__section">
+      <h2 class="copy__title">
```
eddybrando, line 52: Same nice to have as here https://github.com/Qiskit/qiskit.org/pull/1678/files#r614768522

pages/events/summer-school.vue
```diff
@@ -0,0 +1,288 @@
...
+        Frequently asked questions
+      </h2>
+      <cv-accordion :align="`end`">
```
eddybrando, line 55: Since we're just passing a string literal, there is no need for binding:

```suggestion
      <cv-accordion align="end">
```

pages/events/summer-school.vue
```diff
@@ -0,0 +1,288 @@
...
+            {{ item.question }}
+          </template>
+          <template slot="content">
+            <!-- TODO: HTML content should not be in strings but in components
+            but lacking of a better solution given time constraints. -->
+            <!-- eslint-disable vue/no-v-html -->
+            <p v-html="item.answer" />
+            <!-- eslint-enable -->
+          </template>
+        </cv-accordion-item>
```
eddybrando, line 67: Here we can deconstruct the elements of `eventData.faq`:

```suggestion
        <cv-accordion-item
          v-for="{ answer, index, question } in eventData.faq"
          :key="index"
        >
          <template slot="title">
            {{ question }}
          </template>
          <template slot="content">
            <!-- TODO: HTML content should not be in strings but in components
            but lacking of a better solution given time constraints. -->
            <!-- eslint-disable vue/no-v-html -->
            <p v-html="answer" />
            <!-- eslint-enable -->
          </template>
        </cv-accordion-item>
```

pages/events/summer-school.vue
```diff
@@ -0,0 +1,288 @@
...
+
+    @include mq($until: large) {
+      margin-bottom: $layout-01;
+    }
+  }
+}
+</style>
+
+<style lang="scss">
+// overrides
```
eddybrando, line 221: Please consider using a [deep selector](https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors) to still scope these definitions to this component.

eddybrando (changes requested):
Thanks a lot for the PR! 🙂  I like where this is going!

Please take a look at my comments. Some of them are just recommendations.

Also, since I believe there will be other changes to the way we define and parse the data via nuxt-content, and the way the segments work, I didn't go deep into those code changes. I'll do so in a later stage of the code review, once said changes are made.

pages/events/summer-school.vue
techtolentino: @eddybrando - I've tried using deep selection before and was guided to not using them... 

https://github.com/Qiskit/qiskit.org/pull/951#discussion_r495666584


Unless I'm misunderstanding

pages/events/summer-school.vue
techtolentino: Removed

pages/events/summer-school.vue
eddybrando: This doesn't apply anymore after the newest changes.

pages/events/summer-school.vue
eddybrando: Oh ok, I get what is meant on that comment.

We should in general, when customising Carbon components, try and check if the customisation can be applied globally by updating the Carbon theme variables.

If that's not possible, then we should try and apply the styles globally, like those in the "checkbox", not from this specific component, but putting them on another file inside the **assets** directory.The reason being that

korgan00: I think that deep is not bad. What Salva said is to put that code into the community-theme.scss so you can reuse it in every component without a copy/paste.
We should talk with Russell to unify or rethink the use of the accordion and tabs.

techtolentino: Pls let me know if this is right... 

or if you'd rather have me move these to the community-theme

korgan00: For me is a new refactor issue in collaboration with Russell.

y4izus:
Finally we merged [this PR](https://github.com/Qiskit/qiskit.org/pull/1727), so I'll close this one

PR #1691, opened 2021-04-15
Update summer-school-2021.yaml

justjosie:
Review from core planning team - content updates


techtolentino (approved):
Thanks for the update, @justjosie

PR #1693, opened 2021-04-16
Section order in yaml matters

korgan00:
The order of the sections in the `yaml` file is interpreted to display them accordingly.


eddybrando (approved):
LGTM!

PR #1696, opened 2021-04-16
Update summer school images

techtolentino:
Quick PR to update image assets on the Summer School page


y4izus (changes requested):
Could you replace the image names with more semantic text, please? Thanks!!

y4izus (approved):
Thank you Randy!

PR #1699, opened 2021-04-19
[Example] Content distribution in nuxt/content example

delapuente:
**This PR is an example, it is not meant to be merged but to provide feedback in the form of working code.**

TL;DR despite [templates in `nuxt/content`](https://content.nuxtjs.org/writing#vue-components) enables the use of custom elements in markdown and a more declarative use of embedding content, the UX is still too far from being even reasonable for a human content editor. A different alternative is needed, one **promising CMS is https://www.storyblok.com/**.

This PR use [templates in `nuxt/content`](https://content.nuxtjs.org/writing#vue-components)  to provide a richer alternative to YAML. I only refactored the header, for illustration purposes.

Using templates and slots, a human editor can write standard Markdown, custom Markdown, a mix of Markdown and HTML, and embed Vue components to specify the layout in a more semantic way. One could think of the FAQ section as something like. The only considerations to take into account are that autoclosing tags won't work and the custom elements should be `kebab-case`.

```html
<template #faq-section>
  <faq-topic>
    <template #question>
      Can I use **Markdown** as the question?
    </template>
    <template #answer>
      Some answer in **Markdown**
    </template>
  </faq-topic>

  <faq-topic>
    <template #question>
      Can I use **Markdown** as the question?
    </template>
    <template #answer>
      Some answer in **Markdown**
    </template>
  </faq-topic>
</template>
```

The reality is that, for the content to be considered Markdown, the syntax requires some non-obvious tweaks, including all the content to be at the same level of indentation and the existence of a blank line before and after the content. These restrictions make the code very hard to read.

```html
<template #faq-section>
<faq-topic>
<template #question>

Can I use **Markdown** as the question?

</template>
<template #answer>

Some answer in **Markdown**

</template>
</faq-topic>
<faq-topic>
<template #question>

Can I use **Markdown** as the question?

</template>
<template #answer>

Some answer in **Markdown**

</template>
</faq-topic>
</template>
```


new-content/events/summer-school-2021.md
```diff
@@ -258,3 +245,47 @@ faq:
...
+      :segment="{ action: 'qgss2021-enquiry-form' }"></app-cta>
+  </template>
+  <template #faq-topics>
+  <cv-accordion-item>
+  <template #title>
+    What are the pre-requisites for enrolling in the Summer School?
+  </template>
+  <template #content>
+
+  Minimal prerequisites are required for the Qiskit Global Summer School. If you know <app-link url="https://www.khanacademy.org/math/precalculus/x9e81a4f98389efdf:matrices/x9e81a4f98389efdf:multiplying-matrices-by-matrices/v/matrix-multiplication-intro">how to multiply two matrices</app-link>, and have some programming experience in Python, you are ready for the Summer School.
```
delapuente, line 274: Not sure if using qiskit.org components would have worked with the previous method but this surely work and there is no need of using HTML `a` tags when we can use our `AppLink` components.

delapuente (commented):
I've found our components to be too property-oriented. A more content-oriented approach would leverage the composition capabilities of HTML.

components/ui/AppPageHeaderWithCard.vue
```diff
@@ -1,22 +1,26 @@
 <template>
   <header class="app-page-header">
     <div class="app-page-header__container">
-      <main class="app-page-header__main">
+      <div class="app-page-header__main">
```
delapuente, line 4: According to the HTML specification, `main` must be "[hierarchically correct](https://html.spec.whatwg.org/#hierarchically-correct-main-element)" and so:
> A hierarchically correct main element is one whose ancestor elements are limited to html, body, div, form without an accessible name, and autonomous custom elements. Each main element must be a hierarchically correct main element.

delapuente: @y4izus @eddybrando ⬆️

delapuente:
I think the PR fulfilled its demo purposes and now we can close it.

PR #1723, opened 2021-04-22
Modifies AppPageHeaderWithCard to pass description through slots

y4izus:
With the recent changes on the Summer School page, we need to be more flexible with the header description to allow having links. With this PRs we will be able to do it by passing the description through slots


eddybrando (approved):
Nice! Thanks a lot @y4izus

PR #1770, opened 2021-05-12
Reuse filters and results layout

### LINKED ISSUE #944: 👍👍 Reuse filters and results layout on learn and events pages

y4izus:
Fixes #944 

On this PR, the Learn, Advocates and Events pages share the same component for creating the layout of the aside filters and the results container

### DIFF
--- a/components/learn/TheLearningResourceList.vue
+++ b/components/learn/TheLearningResourceList.vue
@@ -128,16 +128,6 @@ export default class TheLearningResourceList extends Vue {
       margin-bottom: $layout-01;
     }
   }
-
-  &__results {
-    & > * {
-      margin-bottom: $layout-02;
-
-      @include mq($until: large) {
-        margin-bottom: $layout-01;
-      }
-    }
-  }
 }
 </style>
 
--- a/components/learn/TheLearningResourceList.vue
+++ b/components/learn/TheLearningResourceList.vue
@@ -33,19 +33,29 @@
           />
         </cv-tabs>
       </client-only>
-      <div class="bx--row">
-        <fieldset class="bx--col-lg-4 bx--col-md-2 the-learning-resources-list__filter-time">
-          <legend
-            class="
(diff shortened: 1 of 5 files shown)


components/learn/TheLearningResourceList.vue
```diff
@@ -33,19 +33,29 @@
...
-        <fieldset class="bx--col-lg-4 bx--col-md-2 the-learning-resources-list__filter-time">
-          <legend
-            class="
-                the-learning-resources-list__filter-time-label
-                the-learning-resources-list__medium-large-only
-              "
-          >
-            Time to spend learning
-          </legend>
+
```
eddybrando, line 36: We could probably get rid of this line to keep it consistent with other pages:

```suggestion
```

pages/events/index.vue
```diff
@@ -52,31 +40,44 @@
...
+          >
+            <div class="event-page__empty-card-description">
+              {{ emptyCard.description }}
+            </div>
+          </AppCard>
+          <EventCard
+            v-for="event in filteredEvents"
+            v-else
+            :key="`${event.title}-${event.place}-${event.date}`"
+            class="results-item"
```
eddybrando, line 71: I find it quite tricky to figure out where this definition is without prior knowledge (like this PRs code review).

I'm think of another approach that could be easier to understand.

1. Maybe we could define `.results.item` in the component using the definition (**pages/events/index.vue** and **pages/learn.index.vue**).
2. Or maybe we could leave the definition in **components/ui/AppFiltersResultsLayout.vue** but rename it to `.app-filters-results-layout__results-item` to hint to where the definition is.

eddybrando (approved):
Thanks a lot for the great PR! 🙂 

Please just consider my 2 suggestions before merging.

pages/events/index.vue
y4izus: I thought the same when I did it 👍 . I like the second option

Issue #1761, opened 2021-05-10
Display new max size breakpoint on Learn page


y4izus:
Adapt the "learn" page to support the new layout for the "max" breakpoint.

---

Related to #1747

PR #1793, opened 2021-05-14
feat: advocates page "max" breakpoint layout

### LINKED ISSUE #1762: Display new max size breakpoint on Advocates page
Adapt the "advocates" page to support the new layout for the "max" breakpoint.

---

Related to #1747

eddybrando:
This PR adapts the advocates page to support the new layout for the "max" breakpoint.

---

Closes #1762


components/ui/AppPageHeaderFixed.vue
```diff
@@ -1,39 +1,57 @@
 <template>
   <header class="app-page-header-fixed">
-    <div class="app-page-header-fixed__container bx--row">
+    <div
```
y4izus, line 3: Could you try something like this?:

```
    <div class="app-page-header-fixed__container"
     :class="{ 'bx--grid' : !oldContainer,
                     'app-page-header-fixed__old-container: oldContainer}">
      <div class="bx--row">
        <h1 class="bx--col-md-6 bx--col-lg-10"
         :class="{ 'bx--col-sm bx--col-max-8': !oldContainer,
                         'app-page-header-fixed__headline': oldContainer}"
          <slot />
        </h1>
      </div>
    </div>
```
I prefer to have this instead of `v-if v-else` because the structure of the components is the same, the only thing that changes are the classes

Note that if you apply this version, you can delete the `width: 100%;` on L46

y4izus (changes requested):
Thanks!! please, take a look at my comment

components/ui/AppPageHeaderFixed.vue
eddybrando: Nice idea! Will do!

y4izus (approved):
Thanks!! I will merge to make easier [this issue](https://github.com/Qiskit/qiskit.org/issues/1761), but please, take a look with Russell at the height of the `join-section__feature`

PR #1796, opened 2021-05-17
Update airtable and filters for the events page

### LINKED ISSUE #1733: Updates on events pages
Right now, we have two pages pulling event's data from the airtable: the Events page and the Seminar Series page. The airtable location has changed, so, we need to update the code to pull data to the new airtble.

Modifications we need to do:
- [x] Point to the new airtable => change base id and views for both events in general and seminar series (ask me if you have more questions about this)
- [x] Replace the column name _Event Website_ with _Website_
- [x] Replace the column name _Picture?_ with _Image / Icon_
- [x] Replace the column name _SUZIE - for website?_ with _Add to Event Site_
- [x] Replace the column name _PAUL - Seminar Site_ with _Add to Seminar Series Site_
- [x] Replace the column name _Speaker_ with _Speaker (S.S.)_
- [x] Modifications on Type of Event. The new list of options is: Hackathon, Camp, Industry Event, Workshop, Challenge, Networking, Talks. That means we should remove 'Unconference' and 'Online' and add ´Challenge' and 'Networking'
- [x] Modifications on Region. The new list of options is: North America, South America, Asia Pacific, Europe, Africa, Online. That means we should remove 'Americas' and add ´North America', 'South America' and 'Online'
- [x] Notice that now we can have an event with two Region (Europe and Online, for example), so we need to adapt the code for that
- [x] Review that the data is correct, both on the [events page](https://qiskit.org/events/) and on the [seminar series page](https://qiskit.org/events/seminar-series/)
- [x] Update the tests (creating or updating if necessary) to check the functionality
- [x] Update the [wiki](https://github.com/Qiskit/qiskit.org/wiki/qiskit.org-events) with the changes

y4izus:
Fix #1733

### DIFF
--- a/tests/hooks/event-conversion-utils.spec.ts
+++ b/tests/hooks/event-conversion-utils.spec.ts
@@ -145,22 +145,13 @@ describe('getRegions', () => {
 })
 
 describe('getLocation', () => {
-  it('defaults in region TBD if there is no location, no region', () => {
+  it('defaults in region TBD if there is no location', () => {
     const { tbd } = WORLD_REGIONS
 
     const nowhereEvent = new FakeRecord({
       name: 'Nowhere Conference'
     })
-    expect(getLocation(nowhereEvent)).toEqual([tbd])
-  })
-
-  it('defaults in region if there is no location', () => {
-    const { northAmerica } = WORLD_REGIONS
-    const noLocationEvent = new FakeRecord({
-      name: 'Fake Conference',
-      regions: [northAmerica]
-    })
-    expect(getLocation(noLocationEvent)).toEqual([northAmerica])
+    expect(getLocation(nowhereEvent)).toEqual(tbd)
   })
 
   it('gets the location from the record', () => {
--- a/tests/hooks/event-conversion-utils.spec.ts
+++ b/tests/hooks/event-conversion-utils.spec.ts
@@ -15,9 +15,9 @@ import { COMMUNITY_EVENT_TYPES, WORLD_REGIONS } from '~/store/events'
 type RecordFields = {
   name: string,
   picture?: object[],
-  types?: string[]|string,
+  types?: string[],
   location?: string,
-  regions?: string[]|string,
+  regions?: string[],
   startDate?: string,
   endDate?: string,
   website?: string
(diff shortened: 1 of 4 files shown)


eddybrando (approved):
Generally LGTM, but would like to point out that we may need to deploy it carefully and test it soon to see if everything works as expected in production with the new Airtable base.

tests/hooks/event-conversion-utils.spec.ts
```diff
@@ -17,7 +17,7 @@ type RecordFields = {
   picture?: object[],
   types?: string[]|string,
   location?: string,
-  region?: string[]|string,
+  regions?: string[]|string,
```
eddybrando, line 20: Just wondering why not just `string[]`? Does the Airtable API send just a string instead of a string inside an array if only one region is selected?

tests/store/events.spec.ts
```diff
@@ -1,40 +1,41 @@
...
-  const hackathonInAmericas = {
+  const hackathonInSouthAmerica = {
     title: 'Fake event D',
-    types: ['Camp', 'Hackathon'],
-    region: 'Americas'
+    types: [COMMUNITY_EVENT_TYPES.camp, COMMUNITY_EVENT_TYPES.hackathon],
+    regions: [WORLD_REGIONS.southAmerica]
   }
-  const unconferenceOnline = {
+  const talskOnline = {
```
eddybrando, line 31:
```suggestion
  const talksOnline = {
```

tests/store/events.spec.ts
```diff
@@ -1,40 +1,41 @@
...
     title: 'Fake event E',
-    types: ['Unconference'],
-    region: 'Online'
+    types: [COMMUNITY_EVENT_TYPES.talks],
+    regions: [WORLD_REGIONS.online]
   }
 
-  const futureEvents = [unconferenceCampInAsia, hackathonInEurope, campInAfrica]
-  const pastEvents = [hackathonInAmericas, unconferenceOnline]
+  const futureEvents = [workshopCampInAsia, hackathonInEurope, campInAfrica, talskOnline]
```
eddybrando, line 37:
```suggestion
  const futureEvents = [workshopCampInAsia, hackathonInEurope, campInAfrica, talksOnline]
```

eddybrando (approved):
LGTM 🙂  But I cannot test the API locally since I don't have access via an API key.

PR #1797, opened 2021-05-17
feat: learn page "max" breakpoint layout

### LINKED ISSUE #1761: Display new max size breakpoint on Learn page

eddybrando:
This PR adapts the learn page to support the new layout for the "max" breakpoint.

---

Closes #1761

### DIFF
--- a/components/learn/TheLearningResourceList.vue
+++ b/components/learn/TheLearningResourceList.vue
@@ -2,7 +2,7 @@
   <section class="the-learning-resources-list">
     <div class="bx--grid">
       <h2>Learning resources</h2>
-      <div class="bx--row the-learning-resources-list__introduction">
+      <div class="bx--row">
         <div class="bx--col-md-5 bx--col-lg-8 bx--col-xlg-7 bx--col-max-6">
           <p>
             The below are designed and created by the Qiskit team. However, we
--- a/components/learn/TheLearningResourceList.vue
+++ b/components/learn/TheLearningResourceList.vue
@@ -1,9 +1,9 @@
 <template>
   <section class="the-learning-resources-list">
-    <div class="the-learning-resources-list__container">
+    <div class="bx--grid">
       <h2>Learning resources</h2>
       <div class="bx--row the-learning-resources-list__introduction">
-        <div class="bx--col-md-5 bx--col-lg-8 bx--col-xl-7">
+        <div class="bx--col-md-5 bx--col-lg-8 bx--col-xlg-7 bx--col-max-6">
           <p>
             The below are designed and created by the Qiskit team. However, we
             recommend a familiarity with
@@ -111,14 +111,6 @@ export default class TheLearningResourceList extends Vue {
 
   background-repeat: no-repeat;
   background-position: left bottom;
-
-  &__container {
-    @include contained();
-  }
-
-  &__introduction {
-    margin-bottom: $spacing-07;
-  }
 
   &__filter-level {
     margin-bottom: $layout-04;
--- a/components/ui/AppCard.vue
+++ b/components/ui/AppCard.vue
@@ -95,18 +95,10 @@ export default class AppCard extends Vue {
   }
 
   &__content {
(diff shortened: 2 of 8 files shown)


components/learn/TheLearningResourceList.vue
```diff
@@ -1,9 +1,9 @@
 <template>
   <section class="the-learning-resources-list">
-    <div class="the-learning-resources-list__container">
+    <div class="bx--grid">
       <h2>Learning resources</h2>
       <div class="bx--row the-learning-resources-list__introduction">
```
y4izus, line 5: You can delete the `the-learning-resources-list__introduction` class

components/ui/AppCard.vue
```diff
@@ -94,11 +95,18 @@ export default class AppCard extends Vue {
...
 
   &__content {
-    padding: $spacing-05 $spacing-05 $spacing-05 $spacing-07;
+    padding: $spacing-05 $spacing-06;
     flex: 1;
     display: flex;
     flex-direction: column;
-    justify-content: space-between;
+
+    @include mq($from: medium) {
```
y4izus, line 103: These two `mq` are not related with this PR, could you delete them?

components/ui/AppFiltersResultsLayout.vue
```diff
@@ -29,21 +29,38 @@ export default class AppFiltersResultsLayouts extends Vue {
...
 
-  &__results-item {
+  ::v-deep &__results-item {
     margin-bottom: $layout-02;
 
     @include mq($until: large) {
       margin-bottom: $layout-01;
     }
+
+    @include mq($from: medium) {
```
y4izus, line 47: At least at the moment, the card's height should be adapted to the content, so please, delete these `min-height`

y4izus (changes requested):
Thank you! please, take a look at my comments and ask me for another review :)

pages/learn/index.vue
```diff
@@ -24,27 +24,35 @@
...
+          v-for="resource in filteredLearningResources"
+          :key="resource.path"
+          class="bx--col-sm-4 bx--col-xlg-8"
+        >
+          <AppCard
+            :image="resource.image"
+            :title="resource.title"
+            :to="resource.to"
+            :cta-label="resource.ctaLabel"
+            class="app-filters-results-layout__results-item app-filters-results-layout__results-item_2-cols-xlg"
```
y4izus, line 50: Why did you add this `app-filters-results-layout__results-item_2-cols-xlg`? It's not enough with adding

```
@include mq($from: x-large) {
     height: calc(100% - #{$layout-02});
 }
```

to the `&__results-item`?

pages/learn/index.vue
```diff
@@ -153,3 +160,15 @@ export default class LearnPage extends QiskitPage {
   }
 }
 </script>
+
+<style lang="scss" scoped>
+.app-filters-results-layout {
```
y4izus, line 165: Why did you add this code here instead of in the `AppFiltersResultsLayout` component?

y4izus (commented):
Thank you! I have two more comments and we finish :)

pages/learn/index.vue
eddybrando: Because `app-filters-results-layout__results-item` is also used in the **Events Page**, which currently doesn't have the "max" layout, so it would look like this:

Screenshot 2021-05-25 at 12 57 11

pages/learn/index.vue
eddybrando: Good catch!

It makes sense for that definition live in this file, but the name of the class should actually be `learn-page__careful-explanation`.

I'll change it. Thank you! 🙂

pages/learn/index.vue
y4izus: great! thank you!

y4izus (approved):
Great Eddy!! thank you very much!!

PR #1823, opened 2021-05-21
Mega menu search

korgan00:
fix #1808

PR #1832, opened 2021-05-25
Mega menu intersection filter

delapuente:
This is for comparison purposes with #1823. With this implementation, a title only shows when all the words are present in it.


y4izus:
We've already chosen an implementation for the search filter, so I'll close this one

Issue #1857, opened 2021-06-03
Explore instrumentalizing the word searches in the megamenu


vabarbosa:

eddybrando:
### What questions do we want to answer?

1. What search terms are being used the most in the mega menu?
2. What search terms lead to a link click?
3. Which textbook visits are the result of a search in the mega menu?
4. What search terms do not lead to a link click?

### What should we track?

In order to solve these questions, we need to track:

1. The referrer / the element that emits the tracked event (the _mega menu dropdown search input field thingy_)
2. The search term
3. The link target, in the case of a link click

### Tracking with URL parameters

In web analytics, the `ref` URL parameter appears to be a de facto standard for referrers tracking. The same goes for the `q` parameter for search terms tracking.

Amplitude appears to have some "automatic tracking" support for `ref` (https://developers.amplitude.com/docs/javascript#track-referrers), but I wasn't able to confirm what this means since I currently don't have access to the Amplitude dashboard.

There is no mention of the `q` parameter in the Amplitude documentation that I could find. But if it works similarly to the Google Analytics dashboard, we may be able to create a filter to get the search term data for the events containing the _mega menu dropdown_ referrer, and so be able to answer our analytics questions.

The URLs that we would get would be in a syntax similar to this: `<origin + path>/?<referrer>&<search term>`

**Example:** `https://qiskit.org/course/foo-course/?ref=textbook-mega-menu&q=foo`

In this example, "textbook-mega-menu" identifies the _mega menu dropdown, where the user searched for "foo", which resulted in a visit to `https://qiskit.org/course/foo-course/`.

### Tracking of search terms that do not lead to link clicks

Searches that do not result in link clicks won't be tracked with the aforementioned approach. Nevertheless, these might contain valuable information in terms of UX optimisation, that may lead us to rename links, create synonyms, etc.

In order to track these search terms, we may emit tracking events via JS after the user stops writing a search term.

This might be achieved with a timeout function that delays emitting the tracking event for about 300 ms, and cancels every other event within said timeframe.

eddybrando:
### Update regarding tracking with URL parameters

Some of the links in the _Textbook mega menu_ point to URLs that respond with a 302 redirect (Mathigon's implementation).

**Example:** `https://platypus.qiskit.org/course/ch-prerequisites?ref=foo` redirects to `https://platypus.qiskit.org/course/ch-prerequisites/environment-setup-guide-to-work-with-qiskit-textbook`

[...]

Issue #1847, opened 2021-06-02
As a product manger, I want to be able to view textbook analytics, so that I can better improving the learning experience.


lerongil:
Acceptance Criteria: 

- [x] Hotjar correctly instrumented on all learn pages 
- [x] Segment instrumented for search to track words searched and pages clicked #1857 #1859
- [x] CTAs instrumented with Segment #1861
- [x] # of Multiple Choice Questions viewed and # of Multiple Choice Questions completed #1858 #1860
- [x] Multiple Choice Questions instrumented so that we can see % of users answering questions correctly #1858 #1860
- [x] Total number of exercises completed  #1858 #1860

lerongil:
Acceptance criteria point 1 is related to #1851

lerongil:
This user story is finished. Thanks to the team for your efforts. We did find a bug which is described here #1851. We will add this to the backlog.

PR #1890, opened 2021-06-15
Ev issue 1859 track mega menu search terms

eddybrando:
This PR implements the Segment analytics tracking for search terms in the **Textbook Mega Menu**.

The search terms are being tracked via the `trackEvent` function of Segment, similarly to the "Clicked CTA" events.

We're sending some additional event information to Segment with this event:

1. **Location:** The name of the component (for example: `Textbook mega menu`)
2. **Text:** The searched term (for example: `algorithm`)

The tracking event is triggered 750 milliseconds after the user stops typing something in the **Textbook Mega Menu** component.

---

Related to: #1847, #1857

Closes: #1859

### DIFF
--- a/components/textbook-demo/TextbookDemoContentMenuSection.vue
+++ b/components/textbook-demo/TextbookDemoContentMenuSection.vue
@@ -1,7 +1,10 @@
 <template>
   <section class="content-menu-section">
     <div class="bx--grid">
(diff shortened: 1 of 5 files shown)


y4izus (approved):
Really nice job!! Thank you!

PR #1901, opened 2021-06-18
feat(textbook): sticky mega menu

### LINKED ISSUE #1891: Make the mega menu nav sticky

eddybrando:
This PR makes the Textbook Mega Menu component stick to the top when scrolling down in the courses pages.

To test it, go to a course page in the Vercel preview, like `/textbook-demo/course/introduction-course`.

https://user-images.githubusercontent.com/22047320/122562829-c74c8880-d043-11eb-9010-9271984fc1da.mov

---

Closes #1891

### DIFF
--- a/components/textbook-demo/TextbookDemoHeader.vue
+++ b/components/textbook-demo/TextbookDemoHeader.vue
@@ -125,9 +125,8 @@ export default class TextbookDemoHeader extends (Vue as VueConstructor<VueCompon
     margin-top: $layout-03;
 
     &_fixed {
-      position: fixed;
+      position: fixed !important;
       transition: .3s ease-in-out;
-      top: 0;
       width: 100%;
     }
   }
--- a/components/textbook-demo/TextbookDemoHeader.vue
+++ b/components/textbook-demo/TextbookDemoHeader.vue
@@ -20,7 +20,7 @@
     <transition name="scroll-in">
       <TextbookDemoContentMenuSection
         v-if="!appMegaDropdownMenuIsVisible"
-        class="textbook-demo-header__dropdown_fixed"
+        class="textbook-demo-header__dropdown-fixed"
       />
     </transition>
   </header>
@@ -123,12 +123,14 @@ export default class TextbookDemoHeader extends (Vue as VueConstructor<VueCompon
 
   &__dropdown {
     margin-top: $layout-03;
+  }
 
-    &_fixed {
-      position: fixed !important;
-      transition: .3s ease-in-out;
-      width: 100%;
-    }
+  &__dropdown-fixed {
+    position: fixed;
+    top: 0;
+    transition: .3s ease-in-out;
+    width: 100%;
+    z-index: 100;
   }
 
   &__cta {
--- a/components/textbook-demo/TextbookDemoContentMenuSection.vue
+++ b/components/textbook-demo/TextbookDemoContentMenuSection.vue
@@ -25,5 +25,7 @@ export default class TextbookDemoContentMenuSection extends Vue {
   background: $background-color-lighter;
   border-top: 1px solid $border-color;
   border-bottom: 1px solid $border-color;
+  position: sticky;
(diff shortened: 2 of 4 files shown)


eddybrando:
@y4izus, I corrected the issue with discussed offline about the mega menu not properly sticking to the top on the the Textbook home page. This is now fixed with https://github.com/Qiskit/qiskit.org/pull/1901/commits/59c970b8fc716358194fa4659f422397ef9235f6:


https://user-images.githubusercontent.com/22047320/122896701-9f15a000-d349-11eb-9e9c-7aa525801d93.mov

components/textbook-demo/TextbookDemoHeader.vue
```diff
@@ -125,9 +125,8 @@ export default class TextbookDemoHeader extends (Vue as VueConstructor<VueCompon
     margin-top: $layout-03;
 
     &_fixed {
-      position: fixed;
+      position: fixed !important;
```
korgan00, line 128: I was trying to avoid the use of this fixed parameter in order to also avoid the `!important` annotation. I found [here](https://www.geekabyte.io/2014/03/when-css-position-sticky-doesnt-work.html) in a comment: 

> once a sticky element's parent goes out of view, the element will no longer stick.

It is possible to fix but probably not worth it.
To avoid the `!important` you can use a selector with higher specificity: `&_fixed.content-menu-section`

korgan00 (changes requested):
LGTM, just one thing.

components/textbook-demo/TextbookDemoHeader.vue
eddybrando: I'm all for avoiding `!important` when possible, but here I thought it might be a cleaner and safer approach.

By adding specificity via the `.content-menu-section` class, we depend on said class name, but this component isn't aware of that class. The clases in the `TextbookDemoContentMenuSection` component –including (`.content-menu-section`)– aren't exposed in purpose. We would be just kind of using an unintended side-effect here IMO. That class could change down the road making this solution fail silently for instance.

eddybrando: `sticky`, as you point out, makes the element stick to the nearest ancestor and containing block (https://developer.mozilla.org/en-US/docs/Web/CSS/position#values).

I considered some ideas about adjusting the order or the elements to have the "sticky navbar" rendered higher in the DOM tree, but they were much more complex and didn't really provide any significant benefit that out-wages the downside of said complexity.

Besides being generally agains `!important`, it's a feature that we still can use, as long as we are careful and don't overuse it.

korgan00: You are right about the "fail silently" reasoning, but then I think that the `!important` could do something similar in some cases. Also, if we want to make a variant of this class we should overwrite again the `!important` using specificity.

The only permitted use case of `!important` in [Mozilla](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) is to overwrite external CSS such as bootstrap. If we externalize this component, could be a use case. But we are working with `v-deep` in some cases to prevent the use of `!important` with the carbon components, so I think that we should follow the same pattern in all the code.

Maybe this is a good discussion with all the dev team to see edge cases for `!important`. But at this moment, I think is better to follow our current patterns.

eddybrando: There is a very good argument about keeping the component independent in terms of positioning and external geometry as mentioned in BEM's documentation (https://en.bem.info/methodology/faq/#why-is-external-geometry-and-positioning-set-via-the-parent-block), as discussed with @delapuente offline.

I've corrected this with https://github.com/Qiskit/qiskit.org/pull/1901/commits/cfe8be0bf387cb5a01b2ee80742b10c4605615d0.

korgan00 (approved):
Good job

y4izus (approved):
Nice work! thank you!

PR #1936, opened 2021-06-30
feat(segment): new Segment properties structure

### LINKED ISSUE #1511: Create reusable definitions of segments actions' names for tracking

eddybrando:
This PR updates the structure of the data passed to Segment via the _properties_.

---

Closes #1511

### DIFF
--- a/components/advocates/AdvocatesJoinSection.vue
+++ b/components/advocates/AdvocatesJoinSection.vue
@@ -25,6 +25,7 @@
 import Vue from 'vue'
 import { Component } from 'vue-property-decorator'
 import CompactFeature from '~/components/ui/CompactFeature.vue'
+import { GeneralLink } from '~/constants/appLinks'
 
 type Benefit = Pick<CompactFeature, 'icon'|'title'|'description'>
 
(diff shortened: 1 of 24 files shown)


eddybrando:
Moving this to "draft" to investigate an issue with some tracking events being sent multiple times.

eddybrando:
@y4izus as discussed offline, regarding the multiple tracking events, I believe this is caused locally due to Segment retrying the failing requests, but I believe this is not exclusive to the implementation in the Textbook Beta, but just the general behaviour it has in the whole project when ran locally.

techtolentino (approved):
This looks great, @eddybrando 

And thanks for the follow-up to your conversation with @y4izus. 


One question I had: do you have screenshots or examples of what these _multiple tracking events_ looked like?

eddybrando:
> This looks great, @eddybrando
> 
> And thanks for the follow-up to your conversation with @y4izus.
> 
> One question I had: do you have screenshots or examples of what these _multiple tracking events_ looked like?

Sure, @techtolentino!

Here are a couple of screenshots. This happens when you are running the project locally via the `npm run dev-analytics` script, go to the http://localhost:3000/textbook-beta/course/introduction-course page and click on the CTA in red:

image

image

The `t` network events are the tracking events. They keep popping up since the events keep retrying after failing.

y4izus (approved):
Thank you very much Eddy! this refactor will be very helpful!

Issue #2014, opened 2021-07-21
Clean the NuxtJS server warnings


abdonrd:
Right now the `npm run dev` script return this:

```
> NODE_ENV=development nuxt-ts

 WARN  mode option is deprecated. You can safely remove it from nuxt.config

ℹ Parsed 6 files in 0,6 seconds

   ╭───────────────────────────────────────╮
   │                                       │
   │   Nuxt @ v2.14.12                     │
   │                                       │
   │   ▸ Environment: development          │
   │   ▸ Rendering:   server-side          │
   │   ▸ Target:      server               │
   │                                       │
   │   Listening: http://localhost:3000/   │
   │                                       │
   ╰───────────────────────────────────────╯

ℹ Preparing project for development
ℹ Initial build may take a while

 WARN  Skipping content generation. Set GENERATE_CONTENT to enable it.

✔ Builder initialized
✔ Nuxt files generated

 WARN  Browserslist: caniuse-lite is outdated. Please run:
npx browserslist@latest --update-db

Why you should do it regularly:
https://github.com/browserslist/browserslist#browsers-data-updating


✔ Client
  Compiled successfully in 1.71m

✔ Server
  Compiled successfully in 1.75m

ℹ Waiting for file changes
ℹ Memory usage: 861 MB (RSS: 2.11 GB)
ℹ Listening on: http://localhost:3000/
No issues found.
```

I think we can try to fix these ones:

- [ ] mode option is deprecated #2015
- [ ] caniuse-lite is outdated #2016

PR #2015, opened 2021-07-21
Remove deprecated NuxtJS mode option

abdonrd:
Help to fix one of the #2014 warnings


eddybrando (approved):
Nice! Thanks a lot!

Just for future reference: here is a short deprecation notice: https://nuxtjs.org/docs/2.x/configuration-glossary/configuration-mode

PR #2016, opened 2021-07-21
Update the Browserlist Data

abdonrd:
Help to fix one of the #2014 warnings.

Browserlist data updated after run this command:

```
npx browserslist@latest --update-db
```


eddybrando (approved):
Nice! Thank you!

PR #2023, opened 2021-07-26
Drop Node.js v12 from the CI

### LINKED ISSUE #2022: Drop Node.js v12 support
The Node.js LTS (long time support) version is v14.

So I think we can remove it from the CI.

abdonrd:
Fix #2022


eddybrando (approved):
Nice job! Thank you!

PR #2042, opened 2021-08-04
Creates India week women in quantum page

### LINKED ISSUE #2048: India week of women in quantum page

y4izus:
Fix #2048


delapuente:
We will need to update the branch or rebase after merging #2045 and solve the preview problem.

eddybrando (approved):
Nice job! Thanks a lot!

As discussed, there is some general clean up in the CSS that is needed, but it comes from another pages that does the same, so we should handle it on a separate issue.

PR #2088, opened 2021-08-13
Update most of dependencies

abdonrd:
Update most of dependencies to easy Nuxt update.

- Keep the Carbon dependencies with `~` to avoid visual changes on updates.
- Keep the `0.` packages also as `~` to avoid posible breaking changes.
- Also fix lint errors & warnings.
- Also add missed `typescript` dev dependency.
- Also add missed `markdown-it` dependency.


eddybrando:
@abdonrd, the CI's build stage is failing due to `mismatching Vue package versions`: https://github.com/Qiskit/qiskit.org/runs/3325693268?check_suite_focus=true.

I believe that re-generating the `package-lock.json` might solve this.

Alternatively, we could manually define the versions of `vue` and `vue-server-render`, but I thing this is really caused by peer-dependencies, so we probably shouldn't be manually setting those versions numbers.

abdonrd:
Ops! I pushed one last commit that I shouldn't! Erasing...

eddybrando:
I'm afraid I get "an error occurred" in the Heroku preview and I have no access to the logs to try and debug it.

abdonrd:
Working locally, I don't know what happens with Heroku.
But I don't have access to see the logs either.

abdonrd:
One dependency was missing! Fixed!

y4izus (approved):
LGTM!! Thank you very much @abdonrd !

PR #2089, opened 2021-08-13
Update Nuxt to v2.15

abdonrd:
- [x] Use Nuxt v2.15 Typescript built-in support. `nuxt-ts` > `nuxt`
- [x] Remove `@nuxt/typescript-runtime`.
- [x] Remove `ts-node`.
- [x] Update Nuxt Components v2: https://github.com/nuxt/components#migration-guide

---

~~Depends on #2088~~


nuxt.config.ts
```diff
@@ -112,7 +112,20 @@ const config: NuxtConfig = {
...
+    '~/components/global',
+    '~/components/landing',
+    '~/components/layouts',
+    '~/components/learn',
+    '~/components/logic',
+    '~/components/metal',
+    '~/components/overview',
+    '~/components/textbook-beta',
+    '~/components/ui'
+  ],
```
abdonrd, line 128: I would like someone to confirm that this covers all cases.

y4izus: By using this way of register components, nuxt includes prefixes and the nested components are not shown. I've modified the configuration to set the `pathPrefix: false`

abdonrd:
Yay! Seems working now! Thanks @y4izus!

y4izus (approved):
Thank you @abdonrd !

PR #2123, opened 2021-08-18
Re-generate the lock file

abdonrd:
This should make future updates easy for us.


y4izus (approved):
Thank you very much @abdonrd !

Issue #1005, opened 2020-10-12
A new way of rendering HTML snippets inside a component


delapuente:
There are situations in which you want to group a list of content and metadata for parametrizing components:
```js
const features = [
  {
    title: 'More features',
    description: 'Unleash the potential of Quantum Computing...'
  },
  {
    title: 'Faster',
    description: 'Now Qiskit is 10x faster compiling...'
  }
]
```
Potentially, you would want to include HTML markup for formatting, such as `<strong>` or `<li>`:
```js
const features = [
  {
    title: 'More features',
    description: 'Unleash the potential of <strong>Quantum Computing</strong>...'
  },
  {
    title: 'Faster',
    description: 'Now Qiskit is <em>10x</em> faster compiling...'
  }
]
```
When dealing with the first structure, we can use template interpolation but when dealing with the second, we need to use [`v-html` directive](https://vuejs.org/v2/api/#v-html).

The problem with using `v-html` is the the DOM is not processed by Vue.js and so it can cause unexpected behaviour such as not being scoped for CSS and it is also vunerable to XSS attacks sometimes. It would be better to switch to a component approach but the boilerplate is too high for these simple usages.

This issue is about investigating/inventing a solution for this problem.

PR #2157, opened 2021-09-13
Add summer school course card

### LINKED ISSUE #2154: Add new Summer School card to learn.qiskit.org page

techtolentino:
Fixes #2154


components/textbook-beta/StartLearningSection.vue
```diff
@@ -6,7 +6,7 @@
     <article
       v-for="{ title, description, courses } in learningSections"
       :key="title"
-      class="bx--row"
+      class="bx--row start-learning-section__row"
```
y4izus, line 9: Could you double check this? I think you don't need this `start-learning-section__row` because we are already using the carbon styles for rows

y4izus (changes requested):
Thank you very much Randy! please, remove the package.json before merge and please, take a look at my comments and ask me for another review when you finished

components/textbook-beta/StartLearningSection.vue
```diff
@@ -122,6 +122,22 @@ export default class StartLearningSection extends Vue {
         }
       ]
     },
+    {
+      title: 'Summer Schoools',
```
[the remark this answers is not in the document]
techtolentino: Thank you, good catch

components/textbook-beta/StartLearningSection.vue
```diff
@@ -122,6 +122,22 @@ export default class StartLearningSection extends Vue {
...
+      description: 'Are you teaching a course on quantum computing? Qiskit provides freely available materials to enhance your course.',
```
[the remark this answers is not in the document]
techtolentino: Correct, this is all placeholder until we get the final content

components/textbook-beta/StartLearningSection.vue
techtolentino: For this, I added a class that _we can own_, because we need to add space between these "sections"

Screen Shot 2021-09-14 at 5 54 29 AM

I then use that class to add spacing, [here](https://github.com/Qiskit/qiskit.org/pull/2157/files#diff-4e893dec92fbbb19f41f5afab4139ffcec5f88324e74eae9a1369f7bda8511feR174)


Let me know if there's a better way

components/ui/AppCard.vue
```diff
@@ -123,6 +125,8 @@ export default class AppCard extends Vue {
   }
 
   &__title {
+    @include carbon--type-style("expressive-heading-02");
```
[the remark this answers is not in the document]
techtolentino: thanks for this, I found the typography styles and just used an element that was available...

korgan00 (approved):
Good job, LGTM

components/textbook-beta/StartLearningSection.vue
```diff
@@ -122,6 +122,22 @@ export default class StartLearningSection extends Vue {
...
+      courses: [
+        {
+          image: '/images/textbook-beta/summer-school/quantum-computing-and-quantum-learning-2021/header.png',
+          title: 'Quantum Computing & Quantum Machine Learning (2021)',
+          description: 'Designed to empower the next generation of quantum researchers and developers with the skills and know-how to explore quantum applications on their own. Starting with an introductory "crash course" on quantum computing, the materials continue to dive into and explore one key area: quantum machine learning.',
+          cta: {
+            label: 'Go to this resource',
+            url: '/textbook-beta/summer-school/quantum-computing-and-quantum-learning-2021',
```
eddybrando, line 135: This is currently pointing to a 404.

components/textbook-beta/StartLearningSection.vue
```diff
@@ -168,3 +184,19 @@ export default class StartLearningSection extends Vue {
...
+    text-decoration: none;
+    color: $link-color;
+    &:hover {
+      color: $link-hover-color;
+    }
+    &:visited {
+      color: $link-visited-color;
+    }
+  }
+}
```
eddybrando, line 201: Why are we overriding this?

The current color (blue) is the one we defined for the "app links". It can be seen in the homepage for instance.

Personally, I believe that we should either keep them all as currently defined or change them all (for example to purple), but not create a special case here.

I would like to hear what @JRussellHuffman thinks about this. 🙂 

To add some visual help, here is how it would look without this special override (meaning, as we defined the links should look like):

Screenshot 2021-09-15 at 18 11 14

And here is how it looks with this override:

Screenshot 2021-09-15 at 18 12 44

Personally, I like using purple instead of blue, since it feels more on-brand. But we should do so for all links IMO. Also, the shade of purple –again, IMO– is quite hard to be distinguished from regular text. I'd like to see either more contrast or an underline, or some kind of highlighting in general.

eddybrando (changes requested):
Thanks @techtolentino!

As discussed offline, we need to first merge https://github.com/Qiskit/qiskit.org/pull/2159 for the 404 link to point to an existing page.

I also added a comment about the link styling, which I believe is important to clarify as well before adding a special case here.

components/textbook-beta/StartLearningSection.vue
eddybrando: As pointed out offline, since this is going to be merged on a feature branch and not the main branch, it can be safely ignored.

components/textbook-beta/StartLearningSection.vue
techtolentino: Thanks for pointing this out, @eddybrando 

I was confused here because in the design, it shows **no** underline:

Screen Shot 2021-09-15 at 10 16 07 AM

So I went and looked for this style, that has the primary blue but no underline... and found it on the https://qiskit.org site, under "Start locally"

so I tried reusing its classes, to achieve the same style... however, I got the underline 🤔 

this is why I was trying to override.

Nonetheless, I removed the overriding styles... 👍🏽

JRussellHuffman: Apologies for the confusion. I agree with Eddy that we should stick with the default link styling.

Not related to the link, but the text: The copy has been updated very slightly. One sentence is removed. This should be the last edit on the text, but if it changes again, I will let you all know.

eddybrando (approved):
Thank you @techtolentino! 🙂 

There is indeed this issue that you mentioned with the link that gets underlined in the description of the section, but it's because we're inserting it with a regular `a` tag instead of using the `app-link` component, which adds the scoped "no underline" style. And we cannot add the component with `v-html`, so your approach is still the correct one.

This is a related issue, but IMO, out of the scope of this PR.

We should look for a way to include links in these sections also without the need of using `v-html`, which is the only option with the current setup. But this is out of scope. I'll be investigating it separately and propose an alternative.

y4izus (approved):
Thanks Randy!

y4izus:
@eddybrando, we have an issue for the `v-html` thing, just you to know: https://github.com/Qiskit/qiskit.org/issues/1005

Issue #2105, opened 2021-08-13
Setup commitlint


eddybrando:
**[commitlint](https://github.com/conventional-changelog/commitlint)** checks if your commit messages meet the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) format.

It's a suggested add-on in newer Nuxt versions.

korgan00:
This is not decoration. It is not related to aesthetics. I think this could be code-quality

shmolf:
I'm going to do the best I can with this. However, commitlint relies on hooks to enforce the commit message meets the format criteria. Specifically, it relies on the [`commit-msg` hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks).

Typically, this would be enforced either by the developer manually adding the hook to their development environment (no guarantee it'll be followed), or by using a hook library like [Husky](https://typicode.github.io/husky/#/).
- One can't easily use their own Git Hook templates or manually-added hooks, along with Husky.

I recommend Husky, since it's a vetted library that's widely used. However, I can at least provide the configuration files in a PR for now.

shmolf:
I'd like to ask... why use commitlint to enforce the format? What tools are leveraging it?

I ask, because I myself do use [semantic-release](https://semantic-release.gitbook.io/semantic-release/) to auto-tag my repo, and auto-generate my Changelog file.

Semantic Release relies on reading in commit messages written in a standard format, such as `conventional commit`, among others.

shmolf:
> I'm going to do the best I can with this. However, commitlint relies on hooks to enforce the commit message meets the format criteria. Specifically, it relies on the [`commit-msg` hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks).
> 
> Typically, this would be enforced either by the developer manually adding the hook to their development environment (no guarantee it'll be followed), or by using a hook library like [Husky](https://typicode.github.io/husky/#/).
> 
> * One can't easily use their own Git Hook templates or manually-added hooks, along with Husky.
> 
> I recommend Husky, since it's a vetted library that's widely used. However, I can at least provide the configuration files in a PR for now.

### Git Hooks
We can either ask devs to create a `commit-msg`  git hook via
```bash
mkdir -p .git/hooks/ && \
echo '#!/bin/sh
npx --no-install commitlint --edit \"$1\"' > .git\hooks\commit-msg
```

Side note: On windows, I ran into an error when I tried to commit.
```
[...]

PR #2195, opened 2021-10-02
Adds CommitLint for commit-message format validation

shmolf:
https://github.com/Qiskit/qiskit.org/issues/2105

This creates installs the [commitlint](https://github.com/conventional-changelog/commitlint) dev dependency, and configures it to reference [conventional commit format](https://www.conventionalcommits.org/en/v1.0.0/).

I've self-tested by providing a poorly-formed commit message, which yielded:
```
⧗   input: poor form
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]

✖   found 2 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
```

As discussed in the ticket ([ref 1](https://github.com/Qiskit/qiskit.org/issues/2105#issuecomment-932801389) & [ref 2](https://github.com/Qiskit/qiskit.org/issues/2105#issuecomment-932822208)), we should still address the necessity of some sort of hook system, whether it's Git or Husky.


techtolentino:
After further discussion, the team has decided to not go with commit linting at the moment.

Thanks for this PR but we'll close for now until something changes.

PR #2339, opened 2021-11-30
Update homepage sections/layout

### LINKED ISSUE #2333: Homepage: Create layout for "What Can Qiskit Do" section
the new homepage designs can be found [here](https://www.figma.com/file/ndskBjmtWzjhdPCRrwMGBi/2009-home-page-key-features?node-id=119%3A11778)

- [x] remove mosaic section
- [x] keep What Can Qiskit Do section
- [x] rearrange section order

techtolentino:
Fixes #2333 

Screengrab showing the rearranged sections

https://user-images.githubusercontent.com/6276074/144128497-cc48d351-b904-4505-9189-e9a4ee5ef09f.mov


components/landing/TheQuickStart/StartOnTheCloud.vue
```diff
@@ -30,8 +30,6 @@ export default class StartOnTheCloud extends Vue {
 <style lang="scss" scoped>
 .start-on-the-cloud {
   &__wrapper {
-    @include responsive-grid-bg('/images/grid/grid-white.svg', 36rem);
-
```
eddybrando, line 34: Nice catch, @techtolentino! I didn't see the change in the design at first.

We need to confirm with @JRussellHuffman if this is intended. But if it is, please also:

1. delete the image in **static/images/grid/grid-white.svg**, as it's no longer used, and
2. remove the `background-position` (also from the `@include` scope in line 41) and `background-repeat` attributes, as they have no longer an effect.

components/landing/TheQiskitCapabilitiesSection/index.vue
```diff
@@ -0,0 +1,22 @@
...
+  @Prop(String) title!: string
+  @Prop(String) description!: string
+}
+</script>
+
+<style lang="scss" scoped>
+.qiskit-capabilities-section {
+  // todo: add styles
+}
+</style>
```
eddybrando, line 22: I believe we shouldn't include this just now as it has no effect and we're also not 100% we are going to need scoped styles in this component, yet.

So it'd be better to just ship what the components needs to work at the moment and include the styles later as needed in the following PR.

```suggestion
```

components/landing/TheQiskitCapabilitiesSection/index.vue
```diff
@@ -0,0 +1,22 @@
+<template>
+  <article class="bx--grid page-section qiskit-capabilities-section">
```
eddybrando, line 2: Related to the comment in lines 17 to 22.

```suggestion
  <article class="bx--grid page-section">
```

components/landing/TheQiskitCapabilitiesSection/index.vue
```diff
@@ -0,0 +1,22 @@
...
+    <h2 v-text="title" />
```
eddybrando, line 3: I believe we don't need to pass this information as a prop since it won't change depending on its context (as it's only found in 1 context: the homepage), so it doesn't need to be dynamic.

```suggestion
    <h2>What can Qiskit Do</h2>
```

components/landing/TheQiskitCapabilitiesSection/index.vue
```diff
@@ -0,0 +1,22 @@
...
+  </article>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+import { Component, Prop } from 'vue-property-decorator'
+
+@Component
+export default class TheQiskitCapabilitiesSection extends Vue {
+  @Prop(String) title!: string
```
eddybrando, line 13: I believe we don't need to pass this information as a prop since it won't change depending on its context (as it's only found in 1 context: the homepage), so it doesn't need to be dynamic.

```suggestion
```

components/landing/TheQiskitCapabilitiesSection/index.vue
```diff
@@ -0,0 +1,22 @@
...
+  @Prop(String) description!: string
```
eddybrando, line 14: This prop isn't used:

```suggestion
```

components/landing/TheQiskitCapabilitiesSection/index.vue
```diff
@@ -0,0 +1,22 @@
+<template>
+  <article class="bx--grid page-section qiskit-capabilities-section">
+    <h2 v-text="title" />
+  </article>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+import { Component, Prop } from 'vue-property-decorator'
```
eddybrando, line 9: If we remove the props (comments in lines 13 and 14):

```suggestion
import { Component } from 'vue-property-decorator'
```

pages/index.vue
```diff
@@ -41,62 +28,8 @@ import { GeneralLink } from '~/constants/appLinks'
 export default class LandingPage extends QiskitPage {
   routeName = 'qiskit-landing-page'
 
-  mosaicSectionTitle = 'What can Qiskit do'
-  mosaicSectionDescription = `Qiskit accelerates the development of quantum applications by
+  qiskitCapabilitiesSectionTitle = 'What Can Qiskit Do'
+  qiskitCapabilitiesSectionDescription = `Qiskit accelerates the development of quantum applications by
```
eddybrando, line 32: We might not need these variables if we decide to drop the props from the **TheQiskitCapabilitiesSection** component:

```suggestion
```

pages/index.vue
```diff
@@ -41,62 +28,8 @@ import { GeneralLink } from '~/constants/appLinks'
...
-    label: 'Full overview',
-    segment: { cta: 'overview', location: 'mosaic' }
-  }
 }
 </script>
-
-<style lang="scss">
-.landing-page {
-  &__grid {
-    @include responsive-grid-bg-strip('/images/grid/grid-mosaic.svg', 112rem, auto, 16rem);
```
eddybrando, line 85: We should also delete the file in **static/images/grid/grid-mosaic.svg** as it's no longer used.

eddybrando (changes requested):
Thank you very much, @techtolentino!

Please take a look at the comments in the review. 🙂

pages/index.vue
techtolentino: Thanks for the review, @eddybrando. I'll wait for confirmation from Russell before I finalize the deletions...

components/landing/TheQuickStart/StartOnTheCloud.vue
techtolentino: following our conversation in Slack, this will stay for now until

pages/index.vue
techtolentino: Apologies - this is correct, in that it's an unused file. Deleted.

eddybrando (approved):
Perfect! Thank you very much, @techtolentino!

PR #2351, opened 2021-12-03
Code cell for new homepage

korgan00:
Fix: https://github.com/Qiskit/qiskit.org/issues/2350

### DIFF
--- a/components/landing/TheQiskitCapabilitiesSection/CodeCell.vue
+++ b/components/landing/TheQiskitCapabilitiesSection/CodeCell.vue
@@ -1,25 +1,25 @@
 <template>
   <code class="code-cell">
-    <div class="code-cell__block" :class="{'code-cell__block_active' : activeCells[0]}">
-      <pre class="code-cell__line"><span class="code-cell__comment"># Qiskit quantum circuits libraries</span></pre>
-      <pre class="code-cell__line">quantum_circuit = qiskit.circuit.library.QuantumVolume(5) <span class="code-cell__comment">#1</span> <span class="code-cell__comment">#3</span></pre>
-      <pre class="code-cell__line">quantum_circuit.draw()</pre>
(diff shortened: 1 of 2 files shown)


components/landing/TheQiskitCapabilitiesSection/CodeCell.vue
```diff
@@ -0,0 +1,65 @@
+<template>
+  <code class="code-cell">
+    <div class="code-cell__block" :class="{'code-cell__block_active' : activeCells[0]}">
+      <pre class="code-cell__line"><span class="code-cell__comment"># Qiskit quantum circuits libraries</span></pre>
```
techtolentino, line 4: i like the `cell__line` naming

components/landing/TheQiskitCapabilitiesSection/CodeCell.vue
```diff
@@ -0,0 +1,65 @@
...
+    margin: -$spacing-03 0;
+    transition: all 0.2s ease-in;
+
+    &_active {
+      opacity: 1;
+      border-left: 0.125rem solid $border-active-color;
+    }
+  }
+
+  &__comment {
```
techtolentino, line 57: can we comma separate these two selectors since they share the same style property?

techtolentino (changes requested):
Looks awesome, @korgan00 

just a small comment, otherwise this is good IMO.

techtolentino (approved):
This looks great, @korgan00 

thank you

components/landing/TheQiskitCapabilitiesSection/CodeCell.vue
eddybrando: As far as I can see, we are not really using the class `code-cell__line`, so we should remove it.

eddybrando (changes requested):
Thank you very much, @korgan00!

I added just a couple of small change requests.

eddybrando (approved):
Thanks a lot for the changes, @korgan00!

PR #2425, opened 2022-01-11
update qml course url

vabarbosa:
update course URL


techtolentino (approved):
great catch, @vabarbosa 👏🏽

PR #2420, opened 2022-01-11
Remove scroll-based code cell highlighting

techtolentino:
Resolves #2419 

This PR **temporarily** removes the scroll-based highlighting.
We'll reintroduce this styling once #2411 is complete.

Branch preview here - https://qiskit-org-pr-2420.dcq4xc5i083.us-south.codeengine.appdomain.cloud/

Issue #2411, opened 2022-01-11
the `scrollBetweenSections` mixin is no longer working


techtolentino:
### Steps to reproduce the problem

1. Go to qiskit.org/overview or qiskit.org/metal
2. Scroll down the page
4. See that the scroll effect isn't working

### What is the current behavior?

The `scrollBetweenSections` mixin used to produce an effect, where as the user scrolls up/down the page, corresponding sections would become active. 

Currently, there is no corresponding interaction.

### What is the expected behavior?

For the `/overview` page, when the user scrolls, we should see the _'page list of contents'_ activate other sections, starting with `Applications`, then `Algorithms`, and so forth.

For the `/metal` page, underneath the section called "Our vision", the images on the right should change, as the user scrolls up/down.

### Screenshots (optional but very helpful)

https://user-images.githubusercontent.com/6276074/148957131-bf89960e-48da-44d5-8743-1ffca9043c37.mov

https://user-images.githubusercontent.com/6276074/148957170-487f99ca-84a9-4f1d-8faf-7837e3ae5583.mov

vabarbosa:
@abdonrd this looks like it is related to running `nuxt generate` (& target static). do you happen to know why the `scrollBetweenSections` mixin would work when you run `nuxt run dev` but does not work with `nuxt generate`?

vabarbosa:
Resolved by https://github.com/Qiskit/qiskit.org/pull/2427

PR #2427, opened 2022-01-12
workaround issue with the `scrollBetweenSections`

vabarbosa:
Related: https://github.com/Qiskit/qiskit.org/issues/2411

this PR is a workaround for the issue of the `scrollBetweenSections` not working when using `nuxt generate`.
it also reverts the temporary change from https://github.com/Qiskit/qiskit.org/pull/2420

preview url: https://qiskit-org-pr-2427.dcq4xc5i083.us-south.codeengine.appdomain.cloud/

---
the main issue appears to be with the `.scrollable` component getting reinitialized/rerendered after the `InspectionObserver` has been initially set. as a result, the callback (from scrolling) is not happening since the original node no longer exists


vabarbosa:
@korgan00 any thoughts on the issue here and a perhaps a better way to resolve instead of just [adding a timeout/delaying the InspectionObserver](https://github.com/Qiskit/qiskit.org/pull/2427/files#diff-eed58d8ecfd1765c3f93845d380926fff3cf3e278e9c95b9331083c083318290R20-R27)?

korgan00:
> @korgan00 any thoughts on the issue here or perhaps a better way to resolve (instead of just [adding a timeout/delaying the InspectionObserver](https://github.com/Qiskit/qiskit.org/pull/2427/files#diff-eed58d8ecfd1765c3f93845d380926fff3cf3e278e9c95b9331083c083318290R20-R27))?

I am thinking about it. I want to be able o test it on my computer. I prefer the [`nextTick`](https://vuejs.org/v2/api/#Vue-nextTick), but I don't know if it will work here. Also, you can call the `updateObserved` method to simplify the code.

vabarbosa:
> > @korgan00 any thoughts on the issue here or perhaps a better way to resolve (instead of just [adding a timeout/delaying the InspectionObserver](https://github.com/Qiskit/qiskit.org/pull/2427/files#diff-eed58d8ecfd1765c3f93845d380926fff3cf3e278e9c95b9331083c083318290R20-R27))?
> 
> I am thinking about it. I want to be able o test it on my computer. I prefer the [`nextTick`](https://vuejs.org/v2/api/#Vue-nextTick), but I don't know if it will work here. Also, you can call the `updateObserved` method to simplify the code.

yeah, i tried `nextTick` but it did not work.

mixins/scrollBetweenSections.ts
```diff
@@ -16,12 +16,12 @@ export default class ScrollBetweenSections extends Vue {
...
-      .querySelectorAll('.scrollable')
-      .forEach((section) => {
-        (this._observer as IntersectionObserver).observe(section)
-      })
+    )
+    this.updateObserved()
+  }
+
+  updated () {
+    this.$nextTick(() => this.updateObserved())
```
vabarbosa, line 24: @techtolentino @korgan00 i think i got it! i was able to resolve the issue using `nextTick` within the `updated` hook. feel free to review, test, check out the [preview link](https://qiskit-org-pr-2427.dcq4xc5i083.us-south.codeengine.appdomain.cloud/) and let me know if you see any problems or have any concerns. thanks.

techtolentino: this looks good to me, but I'll wait for @korgan00 

👏🏽 

thanks @vabarbosa

korgan00: Yeah! It looks so good to me. I always forget the `updated` callback.

Amazing! Thank you!

PR #2433, opened 2022-01-17
Prevents GitHub Actions to be executed by forks

abdonrd:
Also:
- Extract the node version to a env variable.
- Clean the naming to better read:
  Screenshot 2022-01-17 at 10 53 58

### DIFF
--- a/.github/workflows/preview-cleanup.yml
+++ b/.github/workflows/preview-cleanup.yml
@@ -6,9 +6,9 @@ on:
 
 jobs:
   code-engine:
-    # Run the job if it's not triggered by a PR or if the PR is from the same repository
-    # This is because this job requires a secret, and secrets cannot be shared with forks
-    if: github.event.pull_request == null || github.event.pull_request.head.repo.full_name == github.repository
+    # Run only from the original repository
+    # Because this job requires secrets, which cannot be shared with forks
+    if: github.event.pull_request.head.repo.full_name == github.repository
 
     uses: Qiskit/gh-actions/.github/workflows/code-engine-cleanup.yml@main
     with:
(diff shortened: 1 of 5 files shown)


Tansito (dismissed):
LGTM!

.github/workflows/preview-cleanup.yml
```diff
@@ -1,12 +1,15 @@
...
-# run when a pull request is closed
 on:
   pull_request:
     types: [closed]
 
 jobs:
   code-engine:
+    # Run the job if it's not triggered by a PR or if the PR is from the same repository
+    # This is because this job requires a secret, and secrets cannot be shared with forks
+    if: github.event.pull_request == null || github.event.pull_request.head.repo.full_name == github.repository
```
vabarbosa, line 11: for my own curiosity, if this is set to only run`on: pull_request` when would `github.event.pull_request == null`?

.github/workflows/preview-deploy.yml
```diff
@@ -1,12 +1,14 @@
...
+on:
+  pull_request:
 
 jobs:
   code-engine:
-    # Prevents action to be executed by forks
-    if: github.event.pull_request.head.repo.full_name == github.repository
+    # Run the job if it's not triggered by a PR or if the PR is from the same repository
+    # This is because this job requires a secret, and secrets cannot be shared with forks
+    if: github.event.pull_request == null || github.event.pull_request.head.repo.full_name == github.repository
```
vabarbosa, line 10: same question as above, if this is set to only run`on: pull_request` when would `github.event.pull_request == null`?

vabarbosa (approved):
thanks. looks good... just had one question

.github/workflows/preview-cleanup.yml
abdonrd: You're right! No needed in this case!

eddybrando (approved):
LGTM! Thanks, @abdonrd!

PR #2474, opened 2022-01-31
Add contributors statement to qml course

frankharkins:
## Changes

Just added a line explaining who worked on the qml course.


eddybrando (approved):
Thank you @frankharkins! I think this is a nice addition.

PR #2532, opened 2022-03-30
fix(events): events date range filter

### LINKED ISSUE #2518: Events in qiskit.org/events showing up as past events when the start date has passed
From events team:

We noticed that an event happening this week, QIP, is listed as “past” on the qiskit website. It looks like its happening because the start date has passed. I quickly pinged Yaiza as I wasn’t aware she had shifted roles, and she said it should just be a minor change on the [filter](https://github.com/Qiskit/qiskit.org/blob/4967303450137da3b616a05d9f861d7181f4ab51/hooks/event-conversion-utils.ts#L48) so that it goes by the end date instead to keep the event on the “Upcoming” tab.
Is this something the team could help with?

eddybrando:
## Changes

This pull request fixes an issue with our way of filtering the Airtable events by date.

We were only taking into consideration the start date of the events. With this change, we will be taking also the end dates into account whenever it makes sense.

Closes #2518

## Implementation details

Currently, we use an Airtable filtering function to fetch the events that are within a specified date range. But this filtering function is only taking into account the start date of the event.

We couldn't figure out a way to add conditional logic to filter sometimes for end dates and sometimes for start dates based on the data in the database and the requirements of the external function calling the fetching function. At least not on a simple way. There might be a way to do this with the filtering parameter of Airtable, but it might result in a very complex string that is impossible to maintain without having to refer to the Airtable documentation, and also might change at any time.

To keep it simple, readable and testable, and reduce our dependency on an external functions, we're now creating our own filtering function.

## How to read this PR

1. You can scan the updates of the fetched data (**advocates.json**, **past-community-events.json**, **past-seminar-series-events.json**, **upcoming-community-events.json**, **upcoming-seminar-series-events.json**). This only updates the data for local development.
2. Review the `isEventInDateRange` function and its usage.
3. Review the tests.

### DIFF
--- a/hooks/event-conversion-utils.ts
+++ b/hooks/event-conversion-utils.ts
@@ -270,5 +270,6 @@ export {
   getDates,
   formatDates,
   filterWithWhitelist,
+  isEventInDateRange,
   SeminarSeriesEvent
 }
--- a/hooks/event-conversion-utils.ts
+++ b/hooks/event-conversion-utils.ts
@@ -17,6 +17,8 @@ import {
 
 type SeminarSeriesEvent = {
   date: string,
+  startDate: string,
+  endDate: string,
   image: string,
   institution: string,
   location: string,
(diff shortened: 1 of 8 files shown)


hooks/event-conversion-utils.ts
```diff
@@ -59,14 +54,63 @@ function getEventsQuery (apiKey: string, days: number, view: string, filters: st
...
+ * or after today.
+ * If the "days" parameter is positive, the event must happen in the future,
+ * between today and the given number of days after today.
+ * If the "days" parameter is negative, the event must happen in the past,
+ * between today and the given number of days before today.
+ * @param event The event to check.
+ * @param days The number of days before and after today.
+ * @returns Whether the event happens within the specified range.
+ */
+function isEventInDateRange (
```
techtolentino, line 68: I really like this @eddybrando — this _does_ feel better than struggling w/ Airtable filters

hooks/event-conversion-utils.ts
```diff
@@ -157,6 +207,14 @@ function getRegions (record: any): WorldRegion[] {
   return recordRegion || [WORLD_REGIONS.tbd]
 }
 
+function getStartDate (record: any): string {
```
techtolentino, line 210: nice

techtolentino, line 210: nice

techtolentino (approved):
Elegant solution @eddybrando 👏🏽

techtolentino:
@eddybrando - just wanted to add a note here, not sure if it's an official bug:

On the current [Events page](https://qiskit.org/events/), I see an event for `April 8, 2022`.

I think this bug might be back 😓 
Screen Shot 2022-04-12 at 12 07 58 PM

Issue #2503, opened 2022-02-21
As a developer, I want to understand the entire qiskit ecosystem so that I can take advantage of everything that is offered


JRussellHuffman:
Related to https://github.com/Qiskit/qiskit.org/pull/2258

acceptance criteria:
- [ ] finalize designs based off draft created by Abby Mitchell in the above PR
- [ ] determine page location and navigation within Qiskit.org

JRussellHuffman:
Designs can be found here: https://www.figma.com/file/N34caXM86bY1778Whk1c31/2503-ecosystem-page?node-id=64%3A5897

eddybrando:
@JRussellHuffman is this a user story? If so, we should probably change the status to "User story" and create individual issues for each of the tasks described by the acceptance criteria.

JRussellHuffman:
Yes, we probably should, but most of the development effort is being taken care of by Abby Mitchell (not sure her github handle to tag her) and @y4izus. 

It's still worth breaking out into tasks, because this page should link off of the community tab in our header.

1ucian0:
Is this still pending?

JRussellHuffman:
Nope. This was the story to track the design and the design is done. Closing.

PR #2546, opened 2022-04-27
Refactor AppDataTable

y4izus:
## Changes

Refactor of the AppDataTable component as part of https://github.com/Qiskit/qiskit.org/issues/2503

This code also fixes and issue on the [summer school page](https://qiskit.org/events/summer-school/), where the table doesn't scroll and makes part of the page not fit the width of the page
Captura de Pantalla 2022-04-27 a las 12 21 28

## Implementation details
On the mentioned issue, we need to have Carbon components on the AppDataTable and we need to be able to change their color. With the previous version of this component, we weren't allow to change the color because the `fill` property was overwritten when building the dynamic component for the cell inside AppDataTable.

After talking with @korgan00 , we thought the best option was to extract the definition of the cells to have more flexibility

## How to read this PR

I've refactored the tables on the following pages:
- [Summer school page](https://qiskit.org/events/summer-school/)
- [Seminar series page](https://qiskit.org/events/seminar-series/)
- [Physics of computation page](https://qiskit.org/events/physics-of-computation/)
- [India week of women in quantum page](https://qiskit.org/events/india-week-of-women-in-quantum/)

### DIFF
--- a/components/ui/AppDataTable.vue
+++ b/components/ui/AppDataTable.vue
@@ -27,7 +27,7 @@ import { Component, Prop } from 'vue-property-decorator'
 import { GeneralLink } from '~/constants/appLinks'
 
 export interface TableRowElement {
-  component: string,
+  component?: string,
   styles: string,
   data: string | GeneralLink,
 }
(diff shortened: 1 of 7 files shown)


eddybrando (approved):
Thank you @y4izus!

This PR is very helpful! It handles a couple of issues that we had since the use of our tables has changed with the time and we need more flexibility.

Now the AppDataTable component isn't as needed as it was before, so we might need to reconsider if it is needed at all. We can replace it (and most likely a lot of other code) with a combination of a Carbon Qiskit theme and Qiskit web components. But this is beyond the scope of this PR.

Issue #2595, opened 2022-05-12
Broken links in MegaMenu


techtolentino:
### Steps to reproduce the problem

1. Go to '/textbook-beta'
2. Click on dropdown menu, Labs 4-8 

### What is the current behavior?

- Clicking on Labs 4 - 8 take you to a `404` page in the Textbook
- Lab 9 is missing in the menu
- 2 links in Games and Demos are also leading to a `404` page in the Textbook

### What is the expected behavior?

The user should arrive at the correct page for all links

PR #2596, opened 2022-05-12
Fix broken links in MegaMenu

techtolentino:
## Changes

Fixes https://github.com/Qiskit/qiskit.org/issues/2595

## Implementation details

Updated links to pages that exist

## Note
I'm not sure how much of these links will need to be updated, once this PR merges https://github.com/Qiskit/platypus/pull/967. I just wanted to get the fix in, in case users come to the site and use the menu, between major merges. In other words, I didn't wanna leave broken links on the site for the sake of UX.

### DIFF
--- a/constants/megaMenuLinks.ts
+++ b/constants/megaMenuLinks.ts
@@ -171,20 +171,6 @@ const GAMES_AND_DEMOS: MegaDropdownMenuGroup = {
       }
     },
     {
-      label: 'Bonus Level: Make your own puzzles',
-      url: `${baseUrl}${pathGamesAndDemos}/bonus-level-sandbox#bonus-148-0`,
-      segment: {
-        cta: 'bonus-level-make-your-own-puzzles', location: sectionGamesAndDemos
-      }
-    },
-    {
-      label: 'Level 5: Proving the Uniqueness of Quantum Variables',
-      url: `${baseUrl}${pathGamesAndDemos}/bonus-level-sandbox#bonus-153-0`,
-      segment: {
-        cta: 'level-5-proving-the-uniqueness-of-quantum-variables', location: sectionGamesAndDemos
-      }
-    },
-    {
       label: 'Estimating Pi Using Quantum Phase Estimation Algorithm',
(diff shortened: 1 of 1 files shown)


constants/megaMenuLinks.ts
```diff
@@ -172,14 +172,14 @@ const GAMES_AND_DEMOS: MegaDropdownMenuGroup = {
...
-      url: `${baseUrl}${pathGamesAndDemos}/bonus-level-make-your-own-puzzles`,
+      url: `${baseUrl}${pathGamesAndDemos}/bonus-level-sandbox#bonus-148-0`,
       segment: {
         cta: 'bonus-level-make-your-own-puzzles', location: sectionGamesAndDemos
       }
     },
     {
       label: 'Level 5: Proving the Uniqueness of Quantum Variables',
-      url: `${baseUrl}${pathGamesAndDemos}/level-5-proving-the-uniqueness-of-quantum-variables`,
+      url: `${baseUrl}${pathGamesAndDemos}/bonus-level-sandbox#bonus-153-0`,
```
vabarbosa, line 182: do we need these two sections? they are part of the same page as the item above (Bonus Level: Sandbox)?
in addition removing them will make it consistent with platypus:

**qiskit.org**
image

**platypus**
image

i think these two should be consistent (which i know should come once is https://github.com/Qiskit/web-components/issues/74 & https://github.com/Qiskit/platypus/issues/955 are addressed)

vabarbosa (changes requested):
nice @techtolentino just one comment. thanks

constants/megaMenuLinks.ts
techtolentino: Good call... will fix now

techtolentino: @vabarbosa - should be fixed and consistent w/ Textbook now...
Screen Shot 2022-05-12 at 4 13 29 PM

vabarbosa (approved):
@techtolentino looks good. thanks for the quick turnaround

Issue #2612, opened 2022-05-19
Expired slack link


mrvee-qC:
### Steps to reproduce the problem

1. Go to [qiskit.org](qiskit.org)
2. Click on click on any of the slack links **as a fresh user** (use incognito or a new browser)
3. See error

### What is the current behavior?

The current link https://ibm.co/joinqiskitslack redirects to an inactive join link for new users. However existing users will be redirected to the slack workspace. This affects new people from joining in the qiskit slack.

### What is the expected behavior?

The link should redirect new users to qiskit.slack.com to login and join.

### Screenshots (optional but very helpful)

<img width="1512" alt="Screenshot 2022-05-19 at 9 46 57 AM" src="https://user-images.githubusercontent.com/61380731/169186639-66359625-6d6e-4ccd-9968-dc89d66009db.png">

### Environment (optional but very helpful)
 - Device: [e.g. Laptop, iPhone 11, Smart Fridge]
 - OS: [e.g. Windows, iOS]
 - Browser: [e.g. Chrome, Safari]

### Additional context (optional)

Changing link to ibm.biz/qiskitslack should do the trick

Add any other context that could help debug the problem.

techtolentino:
Hi @mrvee-qC - thank you for this issue 👏 

I'm personally having a difficult time reproducing this bug, maybe because all of my devices have Slack?

You'll see in the screenshot below that I use a fresh Safari, private browser (I almost _never_ use Safari)... and I'm still not able to reach the inactive screen you shared above.


https://user-images.githubusercontent.com/6276074/169528154-692210c1-393e-4020-8f5e-e457e42b319e.mov


Do you think it's because I already have Slack on this machine? 🤔 

@eddybrando - any thoughts here?

mrvee-qC:
Hi @techtolentino - thank you so much for the response!

It seems the redirect has now changed to a link that's working. The issue is no longer reproducible and is fixed! I think we can close this issue and ill also close my PR which changed the link to ibm.biz/qiskitslack since the original link redirect works just fine. Glad this was fixed! :) 

Thank you so much!

mrvee-qC:
Hi @techtolentino!

[...]

PR #2613, opened 2022-05-19
Change inactive slack invite

mrvee-qC:
## Changes

Minor url change from https://ibm.co/joinqiskitslack to https://ibm.biz/qiskitslack. Issue explained in #2612 

## Implementation details
Changed in 5 locations which I could find. If there are more, please feel free to change the same.


mrvee-qC:
Redirect for the original link has been fixed! This fix is no longer needed. Will close this!

PR #2656, opened 2022-06-08
Add accessible labels to test icons on ecosystem page

### LINKED ISSUE #2653: Add alt text to test status icons on Ecosystem page
### Feature description

The [ecosystem page](https://qiskit.org/ecosystem/) has cards that have tables with information about the test status of the packages. This information is given in the form of coloured icons:

Screenshot 2022-06-07 at 17 41 53
 
We need to add an alt text to make the icons accessible by providing a description of the icons.

These are the alt texts that should be added:

- Passing tests (for the green icon)
- Failing tests (for the red icon)
- No test information (for the gray icon)

### Alternatives

_No response_

### Additional context

https://www.w3.org/WAI/tutorials/images/informative/

rutterjt:
## Changes

Added accessible labels to test status icons on ecosystem page.

Closes  #2653

## Implementation details

Since the icons are inline `<svg>` elements, they do not support adding an `alt` property. So, since there is no visible text to describe the icons, I opted to implement the accessible labels with `aria-label` as [recommended by MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role#associated_wai-aria_roles_states_and_properties).

### DIFF
--- a/components/ecosystem/TestTable.vue
+++ b/components/ecosystem/TestTable.vue
@@ -1,16 +1,30 @@
 <template>
   <AppDataTable :columns="['Status', 'Test Type', 'Terra Version']">
-    <cv-data-table-row v-for="(row, rowIndex) in tableData" :key="`${rowIndex}`">
-      <cv-data-table-cell v-for="({component, styles, data}, elementIndex) in row" :key="`${elementIndex}`">
+    <cv-data-table-row
+      v-for="(row, rowIndex) in tableData"
+      :key="`${rowIndex}`"
+    >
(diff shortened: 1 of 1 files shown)


techtolentino (approved):
@rutterjt thanks for this

LGTM 👏 

Aside from reading the MDN docs, I also see the `aria-label` attribute recommended in the [Carbon icons-vue readme](https://github.com/carbon-design-system/carbon/tree/main/packages/icons-vue)

Issue #2638, opened 2022-05-31
Bug: Ecosystem page label inconsistency


techtolentino:
### Where does the issue happen?

https://qiskit.org/ecosystem

### What is the content issue?

On the Ecosystem page, the filter labels and the actual card tags have inconsistent casing. This issue would be to fix them so that there's a consistency.

Issues:
- Filters use capital casing
- some tag labels are in ALL CAPS
- at least one has both `prototypes` and `PROTOTYPES`

### Screenshots

<img width="1569" alt="Screen Shot 2022-05-31 at 10 41 56 AM" src="https://user-images.githubusercontent.com/6276074/171251049-e4d9ab59-0ab7-4f5d-9539-fcacc30f531d.png">

### Additional information

question:
some of the cards have tag labels that are not filterable... should all tag labels be in the filters list? 🤔 

e.g.
- hardware
- chemistry
- design

eddybrando:
Correct me if I'm wrong, but I believe the "tier" options are not related to the labels in the cards. @javabster, is that the case?

If so, this wouldn't be a bug (not as described here at least), but it might be a UX issue indeed for which we might want to open a new issue.

javabster:
heya yep I can clarify:

the 'tier' filter options relate to which ecosystem tier a repo belongs to (i.e. what value does it have for `member.tier` in the json we pull in from the ecosystem repo. The labels as they appear on each card is a concatenated array of `members.tier` and `members.labels` (`label ` data comes from literal github labels attached to that given repo). We wanted to have both the github labels and the tiers represented on the app cards and I was trying to work within the bounds of the structure we already have in place with the `AppCard` which sets all the card labels the same purple colour. So I decided to leave the tier labels in all caps (which is how they come straight from the json object) just as a way to differentiate from the other labels

javabster:
so basically @techtolentino this was meant to be a feature not a bug 😉

javabster:
I guess options would be:

a) leave as is
b) switch the tier filters to all upper case
c) switch the tier card labels to title case
d) adjust the AppCard component to allow different styling of labels and convert the tier labels to a different colour
e) other

I don't have a strong preference, open to other ideas too :)

techtolentino:
@eddybrando @javabster -- good to know! I'm okay with leaving as is; and since I'm not the user, let's not file it as a UX bug yet.

(We can revisit if this becomes an issue for anyone else, going fwd.)

PR #2660, opened 2022-06-10
Ecosystem: Refactor AppCard for different tag colours

### LINKED ISSUE #2681: Explanation about the tiers in ecosystem
### Feature description

It would be great to have some explanation of the tiers in the ecosystem page. A sentence or two that can pop up when the user hovers on the tiers or some sort of tooltip 

Screenshot 2022-06-30 at 16 37 37
 

### Alternatives

_No response_

### Additional context

Related to https://github.com/Qiskit/qiskit.org/issues/2135

javabster:
## Changes

This PR refactors the AppCard component to accept objects as well as arrays for the tags. This then allows us to edit the colours of the tags as needed for the ecosystem page

based on discussions from #2638 

UPDATE: closes #2681 

cc @1ucian0 @IceKhan13

## Screenshots
screencapture-localhost-3000-ecosystem-2022-06-10-16_29_50


javabster:
@JRussellHuffman I used the default 'purple' option that comes with the carbon tag (which i believe is accessible by default). Happy to change the styles if you have a preferred colour for branding/accessibility purposes :)

techtolentino:
@JRussellHuffman 

The colors for the text + label, that @javabster introduced pass the color contrast tests, using the [WCAG Extension [1]](https://chrome.google.com/webstore/detail/wcag-color-contrast-check/plnahcmalebffmaghcpcmpaciebdhgdf?hl=en) in Chrome, as well as the [Color Contrast Checker tool [2]](https://marijohannessen.github.io/color-contrast-checker/), recommended by Carbon.

1) Screen Shot 2022-06-14 at 10 36 58 AM (2)

2) Screen Shot 2022-06-14 at 10 40 06 AM (2)

components/ui/AppCard.vue
```diff
@@ -17,12 +17,20 @@
...
-            kind="purple"
+            kind="dark-purple"
+          />
+        </div>
+        <div v-else-if="tags" class="app-card__tags">
+          <cv-tag
+            v-for="tag in Object.values(tags).flat()"
+            :key="tag"
+            :label="tag"
+            :kind="tag === tags.tier ? 'purple' : 'dark-purple'"
```
techtolentino, line 33: @javabster - please check the [Carbon Vue tag component](https://vue.carbondesignsystem.com/?path=/info/components-cvtag--standard), to see if this is allowed... I think `purple` is a valid value, however, I suspect that `dark-purple` is causing issues in the console:

Screen Shot 2022-06-14 at 11 09 37 AM (2)


we might have to leverage one of the pre-existing colors?

techtolentino (dismissed):

javabster:
thanks for the comments @techtolentino! I've addressed your comments and now theres no console warnings when I run locally 😄

techtolentino:
Hey @JRussellHuffman 

Earlier in our discussion, we talked about leaving the app card tags as all dark purple. However, after speaking with @javabster, it turns out that there _is_ a need to distinguish between different tag tiers... and that users need to be able to differentiate.

I'm just commenting here to start the discussion, but @javabster will have more info.

thanks!

javabster:
Just to add a bit more clarity on the motivations for this PR: 

After discussion with @1ucian0 and @IceKhan13 we thought it was confusing to have the 'tier' and 'repo label' tags on the AppCard to be in the same dark purple colour (similar to what was discussed in #2638). We thought we should either a) make the 'tier' tags a different colour or b) figure out a different way to display the 'tier' on the AppCard other than as a tag. I opened this PR using approach (a) as it was technically simpler to implement, not realising it had design implications due to the colour naming restrictions from the carbon component.

JRussellHuffman:
There are a few problems with just changing the colors.
1. accessibility. Rule of thumb, color cannot be the primary carrier of information because colorblind users wouldn't be able to tell the difference between the tags. In other words, color can't be the only difference between the cards if they mean different things, but adding an icon (for example) would solve for this requirement
2. UX. This goes hand in hand with the accessibility requirement, but if I'm a user on the page and the only difference between the tags is color, it's unclear to me how they are different.

It sounds like the "b" option of figuring out a different way to display the 'tier' might be better here. But I can help more if you can share some details about what the tiers are.

eddybrando:
**Note:** We're currently waiting for the resolution of an offline discussion to get a final UI suggestion. Until then, we shouldn't merge this PR since we haven't decided _how_ we want to handle these cases.

FYI: @JRussellHuffman @javabster @techtolentino

techtolentino:
I'm ok w/ closing this PR @eddybrando

eddybrando:
Hi @javabster, just to confirm: We've decided to implement the following UI, right?

Screenshot 2022-07-01 at 13 58 27

Source: https://www.figma.com/file/N34caXM86bY1778Whk1c31/2503-ecosystem-page?node-id=307%3A8021

javabster:
@eddybrando yes that's the plan currently!

But now that #2681 has come up, i'm actually wondering if it would be better to go with this option instead and kill 2 birds with one stone? 

image

@1ucian0 do you have a preference?

1ucian0:
Yeap. Looks good to me😊 thank you!

javabster:
After some internal discussion we have decided to implement the following design, this will also close #2681 

image

y4izus (changes requested):
Hey Abby!! Thank for working on this!! I have some comments, but let's start first with the visual and behavioral part. After fixing that, I would like to do another review with more focus on the code itself :)

- On medium and small screens, there is a misalignment between the title of the card and the tags in the advocates page (where we use the same component)

Now:
Captura de Pantalla 2022-07-11 a las 13 33 52

With your change:
Captura de Pantalla 2022-07-11 a las 13 33 32

- On the filters section, I only see "Main" and "MAIN". I thought we should have all our tiers
Captura de Pantalla 2022-07-11 a las 13 37 37

- I think the text on the tooltip is missing
Captura de Pantalla 2022-07-11 a las 13 38 30

- There are two different tags for "community", should we have just one?
Captura de Pantalla 2022-07-11 a las 13 40 30
Captura de Pantalla 2022-07-11 a las 13 40 40

- On the [figma file](https://www.figma.com/file/N34caXM86bY1778Whk1c31/2503-ecosystem-page?node-id=307%3A8021) content on the tags start with uppercase and the rest is lower case. on this PR we have both everything on lower case or everything on uppercase. Example:
Captura de Pantalla 2022-07-11 a las 13 42 35

- Filter is case sensitive. When I select "Main" I don't get any card
Captura de Pantalla 2022-07-11 a las 13 46 21
I don't get any message neither... @JRussellHuffman , @techtolentino , if it's not done, could you define what happened when the filter doesn't return any result? Thank you!!

javabster:
Hey Yaiza thanks for your comments! I'll get working on the changes, just a few responses:

- Yep the filler text is still there bc I'm waiting on some changes in the backend before I can update, I probably should have changed this to a draft PR as it's not technically finished yet sorry 😅 
- Can you rerun `npm run generate` and see if you still see the capitalisation issue? I think you may have an outdated version of the `memebrs.json` as I don't see this issue when I run locally
- I think we need to keep the 2 different 'community' tags as one refers to a Tier (which has specific criteria attached to it) and the other is just a more general tag that the project owners have the option to set for their repo. When i get the tooltip text hopefully this distinction will be clearer 😄

y4izus:
You're right Abby, I forgot to run `npm run generate`, sorry!!

javabster:
Hey @y4izus I made a few changes based on your feedback:

- removed semicolons ✅ 

- fixed margin on purple tags:
Screenshot 2022-07-15 at 12 37 09 PM

- I'm just waiting on [this](https://github.com/qiskit-community/ecosystem/pull/178) ecosystem PR to get merged then i can update the tooltip filler text

lmk if there's anything else 😄

javabster:
Screenshot 2022-07-21 at 3 38 53 PM

Tooltips now work! This is ready for a final review I think @y4izus

eddybrando:
The one UX issue that I have with this implementation is that I believe that the tiers' descriptions would be more useful to the user before starting to filter, so that they could know what they are filtering for.

With this approach, the description is only seen in the results but not in the filter control elements. To know what the filters refer to in detail, one would need to first find every different tier tag in the cards and hover over the tooltips.

Screenshot 2022-08-01 at 14 23 37

eddybrando (changes requested):
Thank you very much @javabster 🙂 

I have two minor UI-related comments:

---

1. The tags should be aligned to the right and not have this space:

image

---

2. These spacing seems off to me. It feels like the tags need some bottom margin, but I couldn't find a Figma design with this layout, where the tags are aligned to the left on certain breakpoints:

image

The designs that I found [in Figma](https://www.figma.com/file/N34caXM86bY1778Whk1c31/2503-ecosystem-page?node-id=64%3A3953) only show a version with inline tags aligned to the right:

Screenshot 2022-08-01 at 14 36 34

@JRussellHuffman could you please help us clarify how it should look like?

eddybrando:
> @eddybrando @JRussellHuffman
> 
> 1. Sure I'll adjust the spacing 👍
> 2. For the tag spacing on the line below this is how it's always been due to the way the AppCard component was built. you can see the same effect for the Advocates page currently in prod. I can change it if you're happy with changing the underlying behaviour of the component, or if it requires more thought as it will affect multiple pages perhaps it could be put into a followup issue?
> 
> As for the tooltip I was following the design from Russell that we agreed upon [a number of weeks ago](https://github.com/Qiskit/qiskit.org/pull/2660#issuecomment-1177869093). We need some way to indicate on the app card itself the difference between the tier tags and the general tags. Perhaps we could add the tooltip to the filters section in addition to keeping them here?

Thank you! 🙂

Thanks for pointing out that the current design of the AppCard already has the spacing "issue" that I mentioned. There is no need to make any further change to that in this PR as it is out of scope.

And yes, I'm also happy with moving forward with the implementation of what was agreed upon. It's definitely a big improvement and valuable to the user. I just wanted to share some UX concerns I have, but we could implement them later if we think they bring any additional value.

javabster:
@eddybrando I adjusted the spacing! lmk if theres anything else needed on this 😄 


Screenshot 2022-08-04 at 10 50 52 AM

y4izus (changes requested):
Thank you Abby!! I have some comments about the code. Also, could you double check the design of the card? I see the license not aligned
Captura de Pantalla 2022-08-08 a las 13 06 44

Lastly, could you check the console too?? I'm getting a lot of warnings
Captura de Pantalla 2022-08-08 a las 13 07 55

javabster:
Thanks @y4izus I've addressed your comments 😄 I also added a bit more margin under the tags as i was editing the app card component anyway, so this also addresses @eddybrando comment from earlier

Latest version looks like the following when i run locally, I don't see the spacing issue or any errors in the console:

Screenshot 2022-08-08 at 2 56 23 PM
Screenshot 2022-08-08 at 2 56 39 PM

components/ui/AppCard.vue
```diff
@@ -201,3 +207,56 @@ export default class AppCard extends Vue {
...
+
+  &__custom-pill {
+    @include type-style('caption-01');
+
+    background-color: $tag-background-color;
+    color: $tag-text-color;
+    display: inline-flex;
+    min-width: 0;
+    max-width: 100%;
+    min-height: 1.5rem;
```
y4izus, line 235: Please, use for margin and padding the carbon spacing variables like you did [here](https://github.com/Qiskit/qiskit.org/pull/2660/files#diff-39a1c5dd7d7f4b0c976425ad09d8aae1069e5758cbe7a1f616550c1dc1e4ef1eR242)

components/ui/AppCard.vue
```diff
@@ -201,3 +207,56 @@ export default class AppCard extends Vue {
...
+    border-radius: 6.9375rem;
+    cursor: default;
+    vertical-align: middle;
+    word-break: break-word;
+
+    .bx--tooltip__trigger svg {
+      fill: white;
+      margin-left: $spacing-02;
+
+    :hover svg {
```
y4izus, line 252: Could you add the spaces/tabs needed to this block, is it inside `.bx--tooltip__trigger svg`, right?

components/ui/AppCard.vue
```diff
@@ -70,6 +86,10 @@ export default class AppCard extends Vue {
   get hasTags () {
     return Array.isArray(this.tags) && this.tags.length > 0
   }
+
+  get hasTooltipTags () {
```
y4izus, line 90: `hasTags` and `hasTooltipTags` do the same with different variables, you can create an unique method and reuse it passing it the variable

y4izus (changes requested):
Thank you very much Abby!! Just a few more small modifications and we are done!! 👏

pages/ecosystem/index.vue
```diff
@@ -168,6 +170,16 @@ import { GeneralLink } from '~/constants/appLinks'
...
     },
 
+    getTierNames (tiers: any) {
+      const tierNames = tiers.map((tier: any) => tier.name)
+      return tierNames
+    },
+
+    getTierDescription (tier: string) {
+      const tip = (this as any).tiers.find((tip: any) => tip.name === tier)
+      return tip.description
```
y4izus, line 180: Sorry, I've forgotten this one. Please, add an option just in case there are no description: `return tip.description || ""` , for example

y4izus (approved):
LGTM!!!! Thank you Abby!!! 👏

eddybrando (approved):

PR #2671, opened 2022-06-16
Bug: Fixed filters bug and put "Main" at top of filters list

javabster:
## Changes

Due to an update to the ecosystem repo yesterday the filters functionality on the qiskit.org page broke. This PR fixes the issue (caused by changing between upper and lowercase).

This PR also arranges the order of the tier filters so that "Main" is always at the top of the list (and the rest is in the order they appear in `members.json`

@1ucian0 @IceKhan13

## Screenshots
Screenshot 2022-06-16 at 11 52 50 AM

### DIFF
--- a/store/ecosystem.ts
+++ b/store/ecosystem.ts
@@ -11,7 +11,6 @@ const getters = <GetterTree<State, any>> {
    * List of members filtered by selected tiers.
    */
   filteredMembers ({ members, tierFilters }): any[] {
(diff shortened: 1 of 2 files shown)


techtolentino (dismissed):
Thanks @javabster 

I was able to pull this down locally and test prod (via `npm run generate && npm start`)

LGTM, just waiting for checks to pass before we merge.


🚢

techtolentino (approved):
Nicely handled, @javabster 👏

PR #2683, opened 2022-07-08
Update code owners

y4izus:
## Changes

I've added myself as code owner and removed Josie because she is not working with us anymore.

We have an error with Kallie on this file and it looks like it is because she doesn't have write access to the repository. @vabarbosa , @eddybrando , could you check this, please ? Thank you!


eddybrando (approved):
Thank you very much @y4izus!

I invited @KallieFerguson to the repo. When she accepts, she'll the proper access to be a code owner as well.

PR #2684, opened 2022-07-08
Ecosystem: fix test table bug

javabster:
## Changes
This PR fixes a bug where the test type field in the test table was being formatted in the frontend which caused issues when things in the backend repo were updated. Now the backend handles the string formatting so we can remove the formatting on this side and the test table can just update dynamically.

The bug fix is only removing 2 lines (133 and 134 in `ecosystem/index.js` but there are also some linting fixes.

## Screenshots
before (with bug)
Screenshot 2022-07-08 at 11 57 33 AM

after (without bug)
Screenshot 2022-07-08 at 11 59 18 AM

### DIFF
--- a/pages/ecosystem/index.vue
+++ b/pages/ecosystem/index.vue
@@ -1,27 +1,24 @@
 <template>
   <main>
     <AppPageHeaderFixed class="ecosystem-header__hero">
-      <br> Explore
+      <br>
+      Explore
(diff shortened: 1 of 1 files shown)


techtolentino (approved):
Thanks @javabster 

I was able to review this locally and all is well!

🚢

PR #2685, opened 2022-07-08
Ecosystem: Test table updates

javabster:
## Changes

There are 3 main enhancements to the test table in this PR:

- Make the test version table header dynamic, as not every project will be tested against terra (e.g. some might be tested agains nature)
- add tooltips to each test type to explain what it means (@1ucian0 @IceKhan13 I would appreciate your thoughts on the wording for these)
- add a new column with links where users can see the test logs for a specific test

## Screenshots
Screenshot 2022-07-08 at 4 50 13 PM
Screenshot 2022-07-08 at 4 50 22 PM

### DIFF
--- a/components/ecosystem/TestTable.vue
+++ b/components/ecosystem/TestTable.vue
@@ -1,5 +1,12 @@
 <template>
-  <AppDataTable :columns="['Status', 'Test Type', 'Terra Version']">
+  <AppDataTable
+    :columns="[
+      'Status',
+      'Test Type',
+      `${tableData[0][2].packageName} Version`,
+      'Test Logs',
+    ]"
+  >
     <cv-data-table-row
       v-for="(row, rowIndex) in tableData"
       :key="`${rowIndex}`"
@@ -10,6 +17,14 @@
       >
         <!-- eslint-disable vue/no-v-html -->
         <span v-if="component == 'span'" :style="styles" v-html="data" />
+
+        <AppCta
+          v-else-if="component == 'link'"
+          class="app-card__cta"
+          v-bind="{ url: data, label: 'see test logs' }"
+          is-wider
+          kind="ghost"
+        />
         <CheckmarkFilled16
           v-else-if="component === true"
           :fill="styles"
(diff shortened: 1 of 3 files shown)


pages/ecosystem/index.vue
```diff
@@ -130,14 +130,29 @@ import { GeneralLink } from '~/constants/appLinks'
...
+          const timestamp = new Date(res.timestamp * 1000).toLocaleString(
+            'en-UK',
+            {
               timeZone: 'UTC'
-            })
-            return { terraVersion, testType, passed, timestamp }
+            }
+          )
+          // Convert package name to title case
+          const packageName = res.package
```
y4izus, line 142: Could you add a check here? if `res.package` is `undefined` the page crash

pages/ecosystem/index.vue
```diff
@@ -130,14 +130,29 @@ import { GeneralLink } from '~/constants/appLinks'
...
+            .replace(/-/g, ' ')
```
y4izus, line 143: You can use [replaceAll](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) instead of using that Regex

y4izus (dismissed):
Thank you!! In addition to my comments below, could you check if the modification of the packageName to title is working?? I'm afraid is not, because I'm getting 'pytket-extensions' instead of 'Pytket Extensions' (for example)

components/ecosystem/TestTable.vue
```diff
@@ -38,26 +57,41 @@ import { TableRowElement } from '~/components/ui/AppDataTable.vue'
...
-      {
-        component: 'span',
-        data: terraVersion
-      }
-    ])
+    return filteredData.map(
+      ({ passed, testType, packageVersion, logsLink, packageName }: any) => [
+        {
+          component: passed,
+          styles: passed ? '#42be65' : '#da1e28',
```
techtolentino, line 76: non-blocking comment: it would've been nice to use classes and Carbon color tokens...

techtolentino (approved):
Approving since I've seen that you addressed all of @y4izus's comments... 

LGTM

PR #2760, opened 2022-08-23
Ecosystem: Fix tooltip bug

### LINKED ISSUE #2759: Bug: tooltip box gets out out of the window

1. Go to qiskit.org/ecosystem
2. Hover on tier tooltip, close to the edge of the window
3. Part of the box is not possible to read.

It seems like tooltip is not reactive. I assume this is a generic problem of the component.

one option is to move the `^` to the right, so the box can extend to the left and fit. The opposite needs to happen on the left margin of the window. 

### Screenshots

Screenshot 2022-08-23 at 10 21 59

javabster:
## Changes
fixes #2759 

## Implementation details
This bug is caused by a css item that was previously moved into the scoped section but actually should have stayed in the global section. This caused that css to be ignored for some reason so this PR just moves it to the correct place so it is actually implemented.

## Screenshots
now the text is not cut off by the side of the screen anymore:
Screenshot 2022-08-23 at 11 17 08 AM
Screenshot 2022-08-23 at 11 17 35 AM

cc: @1ucian0

### DIFF
--- a/components/ui/AppCard.vue
+++ b/components/ui/AppCard.vue
@@ -160,11 +160,6 @@ export default class AppCard extends Vue {
   }
 }
 
-.bx--tooltip__trigger.bx--tooltip--bottom.bx--tooltip--align-center
-.bx--assistive-text {
-  width: 10rem;
-}
-
 .app-card_vertical {
   flex-direction: column;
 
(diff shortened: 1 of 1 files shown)


techtolentino (approved):
thank you for fixing this @javabster 

I'm good w/ these changes

y4izus (approved):
LGTM too!! thank you Abby!!

PR #2778, opened 2022-09-07
[Suggestions] Change contact email

frankharkins:
## Changes

`hello@qiskit.camp` --> `hello@qiskit.org`

as I'm currently unsure the `.camp` email is monitored.

### DIFF
--- a/constants/menuLinks.ts
+++ b/constants/menuLinks.ts
@@ -233,7 +233,7 @@ const SECONDARY_FOOTER_LINKS: Array<NavLink> = [
   },
   {
     label: 'Contact',
-    url: 'mailto:hello@qiskit.camp'
+    url: 'mailto:hello@qiskit.org'
   },
   {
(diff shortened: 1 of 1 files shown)


techtolentino (approved):
Great catch and thanks for the fix, @frankharkins 

I'm approving so we can get the fix into the footer.

Non-blocking question for you or @lerongil — what do we do with the other pages such as [Qiskit Camp Africa](https://qiskit.org/events/africa/) or [Qiskit Camp Europe](https://qiskit.org/events/europe/)?

Should these pages be removed altogether? They _are_ 3 years old.

I ask because they also have references to the `hello@qiskit.camp` email address... 

If they should be removed, I can start a new issue to remove all unused or deprecated pages

techtolentino:
Going to merge since this PR is good to go. We'll handle the updating or removal of any other pages at a separate time.


Thanks @frankharkins

PR #2783, opened 2022-09-12
Add Qiskit Fall Fest page

techtolentino:
## Changes

Closes https://github.com/Qiskit/qiskit.org/issues/2786

The code in this PR was authored by @leenatantawy; she is no longer interning w/ IBM Quantum at the moment, and has returned to school. I simply took her code and cleaned a few pieces so that it'd be prepared in a way that's closer to our workflow.

## Implementation details

- Added Qiskit Fall Fest page (placeholder)
- Added assets for Fall Fest 2022
- Added University Directory section
- Added placeholder content (to be updated in future PR, once content/info is available)

## How to read this PR
- Review Fall Fest page
- Review components, data/content
- Review UI 

## Screenshots
https://user-images.githubusercontent.com/6276074/189978116-b8173092-be90-4557-8094-acb6957fe2f3.mov


constants/fallFest2022Content.ts
```diff
@@ -0,0 +1,273 @@
...
+    day: 'Friday, July 29',
+    university: 'Qiskit Global Summer School 2022 Commencement & Wrap Celebration',
+    type: '-',
+    url: 'Commencement'
+  }
+]
+
+const scheduleToTableData = (slot: eventSchedule) => ([
+  {
+    styles: 'min-width: 10rem; display: inline-block; font-weight: bold;',
```
yoshiemuranaka, line 178: this is interesting...is this the best way to apply styles? not familiar with this code base, but just expressing concern, as I'm sure we've all experienced trying to track down seemingly random styles that are abstracted from markup or follow a non-standard approach 
(☝️ comment not blocking)

pages/events/fall-fest.vue
```diff
@@ -0,0 +1,230 @@
...
+    // So, disabling. Review the reule 'no-descending-specificity' here:
+    // https://stylelint.io/user-guide/ignore-code
+    // stylelint-disable no-descending-specificity
+    & .bx--tabs__nav-item {
+      &:hover:not(.bx--tabs__nav-item--selected):not(.bx--tabs__nav-item--disabled) .bx--tabs__nav-link,
+      &:not(.bx--tabs__nav-item--selected):not(.bx--tabs__nav-item--disabled):not(.bx--tabs__nav-item--selected) .bx--tabs__nav-link:focus,
+      &:not(.bx--tabs__nav-item--selected):not(.bx--tabs__nav-item--disabled):not(.bx--tabs__nav-item--selected) a.bx--tabs__nav-link:active,
+      &:not(.bx--tabs__nav-item--disabled) .bx--tabs__nav-link, .bx--tabs__nav-item--selected:not(.bx--tabs__nav-item--disabled) .bx--tabs__nav-link:focus {
+        color: $text-color-light;
+      }
```
yoshiemuranaka, line 196: not sure if this is needed? I tried simplifying the declaration, but didn't notice any difference when this whole block is removed...?

yoshiemuranaka (commented):
✨ looks good!
left a few small comments

pages/events/fall-fest.vue
techtolentino: yeah this isn't attractive target selecting, but historically, it was used to override the default styling...


https://user-images.githubusercontent.com/6276074/190221733-9ee4ace6-596c-4940-aa5d-2cdae3a90896.mov

constants/fallFest2022Content.ts
techtolentino: This is copied from an older approach that this code base had implemented, even going a couple of summers back, and it exists in a couple of other places in the code 🤔 

We can open code quality issue? but forgive me for thinking that we let it be for now.

pages/events/fall-fest.vue
yoshiemuranaka, line 196: Is there a difference when you just remove the :hover, :not, etc selectors (line 191-196)? it looks like it is just targeting the text color, which doesn't seem to change even in the video... 🤔 

```suggestion
 
```

constants/fallFest2022Content.ts
yoshiemuranaka: Yeah, seems reasonable to let it be...if this is a pattern we're using in other places in the code, doesn't seem worth trying to solution now. Thanks for the explanation @techtolentino !

yoshiemuranaka (approved):

pages/events/fall-fest.vue
yoshiemuranaka: meant to reply in this thread 🙇 https://github.com/Qiskit/qiskit.org/pull/2783#discussion_r971110566
^also not blocking, just potentially less code 😅

techtolentino: you're right! 

nice catch ⚾ and thanks for the specifics

techtolentino: your suggestion worked, thank you 🙇🏼‍♂️

Issue #2746, opened 2022-08-11
Bug: Quantum Explorers event not showing on Events page


techtolentino:
### Steps to reproduce the problem

1. Go to https://qiskit.org/events
2. See that the Quantum Explorers event is not listed 

### What is the current behavior?

See that the Quantum Explorers event is not listed 

### What is the expected behavior?

The Quantum Explorers event should be listed.

The event runs from Aug 1, 2022 - Feb 28, 2023.

eddybrando:
I can confirm that it's currently in the Airtable database, so we need to check the problem on our end.

anamariarojas123:
@eddybrando @techtolentino any update on this?

techtolentino:
@anamariarojas123 - we discussed this in our dev backlog sync this morning... I believe @eddybrando is going to give it a look.

Thank you for following up

Issue #2675, opened 2022-06-27
[Events] A single-day event card moves to "past events" the day it is happening


1ucian0:
### Steps to reproduce the problem

go to https://qiskit.org/events/ the same day of a single-day event.

### What is the current behavior?

An event happening, say July 1st does not show up in "upcoming events" on July 1st, but in "past events". 

### What is the expected behavior?

An event on a date X should be visible during that date,  AoE. 

### Screenshots

<img width="773" alt="Screenshot 2022-07-01 at 15 14 11" src="https://user-images.githubusercontent.com/766693/176907962-adb95f50-3814-42aa-8acf-a2389ef9b255.png">

PR #2791, opened 2022-09-16
Added new conditions so that events happening today will appear on th…

memolina2323:
### Problem: A single-day event card moves to "past events" the day it is happening
To solve issue [#2675 ](https://github.com/Qiskit/qiskit.org/issues/2675) I added some more conditions in file hooks/event-conversion-utils.ts that work similar to the ones we added in [#2746](https://github.com/Qiskit/qiskit.org/issues/2746) .
Also included the case of events happening "today" in the tests, by adding this special cases to the test in the file tests/hooks/event-conversion-utils.spec.ts.

### DIFF
--- a/hooks/event-conversion-utils.ts
+++ b/hooks/event-conversion-utils.ts
@@ -73,14 +73,19 @@ function isEventInDateRange (
   const today: Date = new Date()
   const eventStartDate: Date = new Date(startDate)
   const eventEndDate: Date = new Date(endDate)
-  const isFutureRange: boolean = days > 0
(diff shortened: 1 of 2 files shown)


techtolentino (approved):
Great work here @memolina2323 

I was able to test this locally by doing:
```
npm install
npm run generate
npm start
```

And the results are as you've described... 

:clap: nicely done



https://user-images.githubusercontent.com/6276074/191559183-708e6fed-a19a-403b-901c-361ee2395440.mov

vabarbosa (approved):
@memolina2323 this is great! thank you!

PR #2842, opened 2022-10-14
Add card subtitle feature

frankharkins:
## Changes

For the new 'basics of quantum information' course, we need a way to separate the title of the specific course from it's wider series. This PR adds the ability to add a subtitle to a card for this specific purpose.

## Implementation details

I've added an extra HTML tag & some css to `AppCard.vue`, and you can now send an optional `subtitle` value to the card. From my tests, these changes don't seem to affect the appearance of the other cards. Hope I've done this right!

No cards currently use this except the new basics of quantum information course which isn't merged yet, so you'll need to add a `subtitle` field to an existing card to try this out (see [3f5efc0](https://github.com/Qiskit/qiskit.org/pull/2831/commits/3f5efc04b2339a315410e45a0401c18a985239c0) for an example).

## Screenshots

Screenshot 2022-10-14 at 09 52 38

### DIFF
--- a/components/ui/AppCard.vue
+++ b/components/ui/AppCard.vue
@@ -18,9 +18,9 @@
           <h4 class="app-card__title">
             {{ title }}
           </h4>
-          <h4 v-if="subtitle" class="app-card__subtitle">
+          <h5 v-if="subtitle" class="app-card__subtitle">
             {{ subtitle }}
-          </h4>
+          </h5>
         </div>
         <div class="bx--row">
           <div v-if="hasTags(tags)" class="app-card__tags">
@@ -168,6 +168,7 @@ export default class AppCard extends Vue {
   }
 
   &__subtitle {
+    font-size: 0.875rem;
     font-style: italic;
     font-weight: normal;
   }
--- a/components/ui/AppCard.vue
+++ b/components/ui/AppCard.vue
@@ -14,9 +14,14 @@
     />
     <div class="app-card__content">
       <header class="app-card__header">
-        <h4 class="app-card__title">
-          {{ title }}
-        </h4>
+        <div>
+          <h4 class="app-card__title">
+            {{ title }}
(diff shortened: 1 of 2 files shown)


JRussellHuffman (dismissed):

vabarbosa (changes requested):
@frankharkins this is great. i tried it locally and was working as expected with one additional change needed:

add `subtitle?: string` to the [`Course` type](https://github.com/Qiskit/qiskit.org/pull/2842/files#diff-dd9762b1bfe2e9a850cb7647190e15bbacdcea5af9078fd8f10865493320f86eR49) definition:

image

otherwise, we get errors in the console, if the Course type is not updated:

image

components/ui/AppCard.vue
```diff
@@ -14,9 +14,14 @@
...
     <div class="app-card__content">
       <header class="app-card__header">
-        <h4 class="app-card__title">
-          {{ title }}
-        </h4>
+        <div>
+          <h4 class="app-card__title">
+            {{ title }}
+          </h4>
+          <h4 v-if="subtitle" class="app-card__subtitle">
```
y4izus, line 21: The `h` HTML tags [corresponds to the levels of nested sections](https://html.spec.whatwg.org/multipage/sections.html#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements:~:text=corresponds%20to%20the%20levels%20of%20nested%20sections), so I think that if we have a title = `h4`, the subtitle should be `h5`

y4izus (changes requested):
Thank you Frank! one more comment :D

components/ui/AppCard.vue
frankharkins: Thanks! Done

vabarbosa (approved):
thank you @frankharkins

PR #2853, opened 2022-10-20
Create style for h5

frankharkins:
## Changes

Moved styling for card subtitle to `assets/scss/layout/_typography.scss` as a rule for `h5`. Should be no visual changes. See #2842 for context.


techtolentino (commented):
@frankharkins thanks for the PR

for a move like this, I think we'll have to get @JRussellHuffman's input, to see if we're introducing a new heading style at the typographic system's level.

this decision would impact our design system across multiple projects as well, if it resides here in the typography file.

in my view, having these styled to the `app-card__subtitle` was the right choice, and I'd only recommend placing it here if the design team decides:
- yes, we will be using `h5` headings now
- yes, they will all be normal font-weight, and italicized 

Open to discussions on this...

JRussellHuffman:
For now, I'm good with it. This is the first time we've had a subheading like this, and hopefully the only time, as most projects shouldn't have another level of hierarchy. 

Does it change things in our design system in that it pushes things, or does it just add another thing? If it's the latter, I think it's unproblematic, but let's discuss it further if it's the former.

techtolentino (approved):
Thanks @frankharkins

JRussellHuffman:
Good to know. It might actually be a good thing in the future to have a header element with italics. thanks!

PR #2911, opened 2023-01-11
Fix accessibility for internal links

### LINKED ISSUE #2847: Bug: the 'get learning' link on the homepage should use better semantic html

I'm opening this issue on behalf of a user who reported this to me and would prefer to remain anonymous

1. Go to https://qiskit.org/
2. Scroll down to 'learn with Qiskit' section
3. try cmd + click on 'Get Learning'

Because the text is a <span> it doesn't respond to shortcut commands as expected like a link or button would. I suspect it probably presents some other accessibility issues as well.

Visually the text looks like a link or a button so would be great if it behaved as users expect a link or a button to behave

techtolentino:
## Changes

Fixes #2847 

## Implementation details

- updated "Get learning" link to use relative path
- added check for `isInternal` links, which are `url`'s that begin with a `/`
- added `isInternal` links to the `isAnchor` check which will render a valid `href` 
- also fixes the "Go to this course" links for the `AppCard`s on the `/learn` page

## Branch preview
https://qiskit-org-pr-2911.dcq4xc5i083.us-south.codeengine.appdomain.cloud/

## Screenshots
- user can now tab to the "Get learning" link
- user can `cmd + click` to open link in a new tab

https://user-images.githubusercontent.com/6276074/211906227-3cc75713-c2ab-4837-87d3-29c0468c479a.mov

Screenshot 2023-01-11 at 11 59 13 AM (2)

### DIFF
--- a/components/landing/TheLearnSection/LearnCard.vue
+++ b/components/landing/TheLearnSection/LearnCard.vue
@@ -30,7 +30,7 @@ import { GeneralLink } from '~/constants/appLinks'
 @Component
 export default class LearnCard extends Vue {
   learnLink: GeneralLink = {
-    url: 'learn',
+    url: '/learn',
     label: 'Get learning',
     segment: { cta: 'get-learning', location: 'learn-card' }
   }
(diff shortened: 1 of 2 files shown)


y4izus (approved):
Awesome Randy!! thank you!!

Issue #2736, opened 2022-08-05
[Advocates Filter] Ordering of the locations on advocates page is arbitrary


HuangJunye:
### Where does the issue happen?

https://qiskit.org/advocates/

### What is the content issue?

After filing https://github.com/Qiskit/qiskit.org/issues/2733, I thought about the ordering of the locations. I am not sure if there were any reasons behind the ordering of the locations in the way it is now:

- North America
- South America
- Oceania
- Africa
- Europe
- Asia

I propose the change to alphabetical order (for N&S America, which uses America to rank):

- Africa
- North America
- South America
- Asia
- Europe
- Oceania

### Screenshots

![Screen Shot 2022-08-05 at 10 25 40](https://user-images.githubusercontent.com/7631333/183035994-75802664-5d38-4c38-b2c6-cbdf969a72ff.png)

HuangJunye:
Sorry for filling this as a separate issue from #2733 as I only thought about this afterwards. Thanks for the quick fix for that one! 

I would also like to bring this to the advocate squad for discussion first. Please put this on hold for the time being. Thanks.

y4izus:
I have removed the bug label since it's not a bug a new request

HuangJunye:
@GemmaDawson agrees with your suggestions.

- Africa
- America (North)
- America (South)
- Asia
- Europe
- Oceania

We can remove the `ON HOLD` label.

PR #2918, opened 2023-01-16
feat(advocates): order and rename locations

y4izus:
## Changes

Fix https://github.com/Qiskit/qiskit.org/issues/2736

## Implementation details

On the advocates page:

- `North America` is now `America (North)
- `South America` is now `America (South)
- Locations now are on alphabetical order

## How to read this PR

- Go to the [advocates page](https://qiskit.org/advocates/) 
- Check that the locations are in alphabetical order
- Check that the label on the advocate card also replace North/South America with America (North/South)
- Check that everything still works as expected

## Screenshots

Captura de pantalla 2023-01-16 a las 16 27 42

### DIFF
--- a/store/advocates.ts
+++ b/store/advocates.ts
@@ -1,8 +1,8 @@
 import type { ActionTree, GetterTree, MutationTree } from 'vuex'
 
 const ADVOCATES_WORLD_REGIONS = Object.freeze({
-  northAmerica: 'North America',
-  southAmerica: 'South America',
+  northAmerica: 'America (North)',
+  southAmerica: 'America (South)',
   oceania: 'Oceania',
   africa: 'Africa',
   europe: 'Europe',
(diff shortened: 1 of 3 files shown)


eddybrando (changes requested):
Great work on the PR @y4izus!

I noticed that the filter functionality is not working as expected. One possibility could be that there is a mismatch between the UI labels and Airtable filters.

As a suggestion, can we map them to ensure proper functionality? For example, currently, there is no advocate with the "Oceania" region in Airtable, but only "Australia", this could be causing the problem.

### Screen recording

https://user-images.githubusercontent.com/22047320/212869855-ee9002b6-808e-4abd-b2e4-f9171befc9cd.mov

y4izus:
> Great work on the PR @y4izus!
> 
> I noticed that the filter functionality is not working as expected. One possibility could be that there is a mismatch between the UI labels and Airtable filters.
> 
> As a suggestion, can we map them to ensure proper functionality? For example, currently, there is no advocate with the "Oceania" region in Airtable, but only "Australia", this could be causing the problem.
> 
> ### Screen recording
>  Screen.Recording.2023-01-17.at.11.07.22.mov

I think the problem is that the advocates hasn't been re-generated. As you can see in my picture, the label on the Advocate card is `America (North)`, while on your video is still `North America`. Data on the airtable was updated yesterday to be `America (North)`

I think the `Australia` problem is another one not related with this issue, I'm gonna try to see what happen with that

eddybrando:
> I think the problem is that the advocates hasn't been re-generated. As you can see in my picture, the label on the Advocate card is `America (North)`, while on your video is still `North America`. Data on the airtable was updated yesterday to be `America (North)`
> 
> I think the `Australia` problem is another one not related with this issue, I'm gonna try to see what happen with that

Ok, I understand. It's quite tricky to test because there is a parallel issue that causes the preview to use outdated data.

As discussed offline, it's OK for me to merge this when you see fit. 🙂

eddybrando (approved):

PR #2926, opened 2023-01-20
feat: download and store Airtable images

### LINKED ISSUE #2894: Download and store advocate's photos from Airtable during build
### Feature description

Airtable recently implemented a change to make the URLs of attachments (like the advocate's photos) expire after a couple of hours:

> On November 8, 2022, Airtable will improve the security of our attachment URLs by incorporating an expiring links functionality across our product surface areas. That means when you share attachments stored in Airtable via URL with external colleagues the link you provide them will expire after a couple of hours. Read on to learn more about how to prepare for the upcoming changes.
>
> **Source:** https://support.airtable.com/docs/changes-to-airtable-attachments

This caused the following bug in our advocates page (https://qiskit.org/advocates/):

- https://github.com/Qiskit/qiskit.org/issues/2893

In order to quickly solve that issue, we've temporarily opted out of the expiring URLs. But this option will be removed on February 8th, 2023.

As recommended by Airtable, we need to download the files during the API fetch and store them on our end so that they remain accessible.

### Additional context

- Airtable information about the change: https://support.airtable.com/docs/changes-to-airtable-attachments
- Advocates' Airtable base: https://airtable.com/app8koO4BZifGFhCV/tblAATp37ybjFZtO3/viwdqD6xFc5JvpQrB?blocks=bipWydYrrs5oss3UF
- API documentation for the Advocates' Airtable base: https://airtable.com/app8koO4BZifGFhCV/api/docs

eddybrando:
## Changes

This pull request addresses an issue caused by Airtable's recent change to make the URLs of attachments expire after a couple of hours. The problem was that the profile pictures for advocates were not being displayed.

As recommended by Airtable, we need to download the files during the API fetch and store them on our end so that they remain accessible.

Closes #2894

## Implementation details

- A new method "storeImage" was added to the "AirtableRecords" class to download the images and store them in the project.
- The "getImage" methods for the advocates and events were adjusted to use the "storeImage" method.
- Two new files called "static/images/advocates/downloaded/.gitkeep" and "static/images/events/downloaded/.gitkeep" were added to the project to keep the directories for the downloaded files in git.
- Any necessary adjustments were made related to the async functions and Promises to ensure that they are working properly.
- All tests were adjusted to ensure that the new functionality is working as expected and new tests were added for the "storeImage" method.

## How to read this PR

- You can start by reviewing the code of the "storeImages" method and its tests.
- Then you can take a look at the changes in the "getImages" methods and the adjustments to the tests for the advocates and events records.


hooks/advocate-conversion-utils.ts
```diff
@@ -42,38 +42,46 @@ class AdvocatesAirtableRecords extends AirtableRecords {
...
     const imageAttachment = attachments && findImageAttachment(attachments)
     const imageUrl = imageAttachment && getImageUrl(imageAttachment)
-    return imageUrl || fallbackImage
+
+    if (!imageUrl) {
+      return fallbackImage
+    }
+
+    const imageName = await this.storeImage(imageUrl, 'images/advocates/downloaded')
+    return imageName
```
y4izus, line 84:
```suggestion
    return await this.storeImage(imageUrl, 'images/advocates/downloaded')
```

hooks/advocate-conversion-utils.ts
```diff
@@ -42,38 +42,46 @@ class AdvocatesAirtableRecords extends AirtableRecords {
...
     }
+    return advocate
   }
 
   public getName (record: any): string {
     return record.get(this.recordFields!.name)
   }
 
-  public getImage (record: any): string {
+  public async getImage (record: any): Promise<string> {
```
y4izus, line 73: We have two `getImage` functions, one for advocates and one for events. They only differ on the URLs, so I think we can create a reusable function passing to it the two URLs we need

Thinking in loud now: in addition, does it make sense to have functions like `getName`, where we only access an object property? if we want a function for that, we could create something like:

```ts
public getElementFromRecord (element:string, record: any): any {
    return record.get(this.recordFields![element])
  }
```

y4izus (changes requested):
In general looks great for me Eddy! Thank you very much! I have only 4 comments/requests :D

hooks/advocate-conversion-utils.ts
eddybrando: As discussed offline, I like these improvement ideas. We'll follow up on https://github.com/Qiskit/qiskit.org/issues/2934 to avoid extending the scope of this pull request.

hooks/airtable-conversion-utils.ts
```diff
@@ -111,6 +115,23 @@ class AirtableRecords {
...
+   * @return {Promise<string>} The path to the stored image
+   */
+  public async storeImage (imageUrl: string, targetDir: string): Promise<string> {
+    const imageFileName = `${this.id}.jpg`
+    const imageFilePath = `static/${targetDir}/${imageFileName}`
+    const imagePublicPath = `/${targetDir}/${imageFileName}`
+    const response = await axios.get(imageUrl, { responseType: 'arraybuffer' })
+    const imageBuffer = Buffer.from(response.data, 'binary')
+    fs.writeFileSync(imageFilePath, imageBuffer)
+    return imagePublicPath
```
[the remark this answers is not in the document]
eddybrando: Adjusted with https://github.com/Qiskit/qiskit.org/pull/2926/commits/7f9be3ec9e45575dc399172c5ae133b95e56d516.

I also updated the tests so that they are more meaningful and properly test that the expected URLs are being used to fetch the images that will be stored.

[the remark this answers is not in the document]
y4izus: thank you!

y4izus (approved):

PR #2931, opened 2023-01-23
Issue 2866 Show Account Button in Learn Page

### LINKED ISSUE #2866: Show the account button in the learn page
### Feature description

The discoverability of the user account is quite difficult because there is no account button in the Learn page. https://qiskit.org/learn/

We should show the account button.

### Additional context

_No response_

mariacloehb:
Fix #2866 

**Changes:**
Modified vue component default-max to showcase account button in the learn page and in courses' overview page.

**Files modified:**
- default-max.vue

**Implementation details:**
Added an if - else statement in the qiskit-ui-shell divisor to show the account button in the learn page and in each course's overview page. Proceeded to add a boolean to work with the if - else statement.  

Coauthored by:  @y4izus and @techtolentino


techtolentino (changes requested):
This looks great @mariacloehb 

My only feedback is that I also think that the course _overview_ pages should also include the User Account icon, because those have the URL structure of:

`https://learn.qiskit.org/learn/course/<some-course-title-here>`

these pages ☝🏼 fall _within_ the experience of any `/learn` page... 

please let me know if that makes sense.


Here's an example:

https://user-images.githubusercontent.com/6276074/214403498-fe3c57ca-34e9-444a-ab97-99e5dbd6ab5c.mov

y4izus:
> This looks great @mariacloehb
> 
> My only feedback is that I also think that the course _overview_ pages should also include the User Account icon, because those have the URL structure of:
> 
> `https://learn.qiskit.org/learn/course/<some-course-title-here>`
> 
> these pages ☝🏼 fall _within_ the experience of any `/learn` page...
> 
> please let me know if that makes sense.
> 
> Here's an example:
> 
>  Screen.Recording.2023-01-24.at.12.31.16.PM.mov

Good catch Randy!! Thank you!

layouts/default-max.vue
```diff
@@ -18,6 +19,8 @@ export default class MaxLayout extends Vue {
   onClick (e: CustomEvent) {
     this.$trackClickEvent(`${e.detail?.label?.toLowerCase()}`, 'menu')
   }
+
+  isLearnPage: boolean = this.$route.path === '/learn/course'
```
techtolentino, line 23: @mariacloehb for this, I think we need to consider both scenarios:

if the path _is equal to_ `/learn` OR if the path _includes_ `/learn/course`

y4izus: it should be 
```suggestion
  isLearnPage: boolean = this.$route.path.includes('/learn')
```

why?:
- `this.$route.path === '/learn/course'` will only recognized the URL `qiskit.org/learn/course`, so it won't work with `qiskit.org/learn` or `qiskit.org/learn/course/<course-name`
- `this.$route.path.includes('/learn/course')` will recognize `qiskit.org/learn/course/<course-name`, but not the URL `qiskit.org/learn`

layouts/default-max.vue
```diff
@@ -18,6 +19,8 @@ export default class MaxLayout extends Vue {
   onClick (e: CustomEvent) {
     this.$trackClickEvent(`${e.detail?.label?.toLowerCase()}`, 'menu')
   }
+
+  isALearningPage: boolean = this.$route.path.includes('/learn')
```
techtolentino, line 23: very nice!

I'm good with this... thank you!

🚢

techtolentino (approved):

layouts/default-max.vue
techtolentino: great point... I like this approach better

PR #2955, opened 2023-01-31
feat(events): add calendar sync section

### LINKED ISSUE #2954: [Events] Add Calendar Subscription section
## Description
We'll need to implement the Events Subscription section by adding the following to the `/events` page.

## Screenshot
Screenshot 2023-01-31 at 8 23 33 AM (2)

### Additional context

[Figma design here](https://www.figma.com/file/YTYDopstugJbJIzJrVPFwx/qiskit.org-event-subscribe?node-id=4742%3A11137&t=VpumwAvcIK2Cm6fU-4)

Related discovery/design issue [here](https://github.com/Qiskit/qiskit.org/issues/2922)

techtolentino:
## Changes

Closes #2954 

## Preview link
https://qiskit-org-pr-2955.dcq4xc5i083.us-south.codeengine.appdomain.cloud/events

## Implementation details
- Added a section to the `/events` page
- Used the `CvCodeSnippet` component for users to copy the subscribe URL
- Added child components, each for the different calendar instructions
  - (I thought this might be easier to maintain, but open to feedback)

## Screenshots

Screenshot 2023-01-31 at 10 04 36 AM (2)


eddybrando:
In the **Apple** tab, the second point says:

> In the menu, select File > New Calendar Subscription

In every other case we use quotes around the labels to be clicked or otherwise selected. Should we also do that here and have it be something like this?

> In the menu, select "File" > "New Calendar Subscription"

pages/events/index.vue
```diff
@@ -83,9 +83,38 @@
...
         <template slot="extra-info">
-          <div class="event-page__start-an-event">
+          <div class="event-page__section">
+            <h3>Follow our event calendar</h3>
+            <p class="event-page__section__description">
+              <!-- eslint-disable-next-line vue/singleline-html-element-content-newline -->
+              Stay up to date with all of our scheduled events by following our calendar. You can view the calendar by visiting <AppLink v-bind="qiskitCalendarLink">{{ qiskitCalendarLink.label }}</AppLink>, or subscribe to it by adding to the calendar app of your choice.
+            </p>
+            <div class="event-page__tabs">
+              <cv-tabs aria-label="navigation tab label">
```
eddybrando, line 93: The `aria-label` attribute should label the tabs. It could be something like "Calendar apps".

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role#example

pages/events/index.vue
```diff
@@ -156,57 +193,81 @@ export default class EventsPage extends QiskitPage {
...
+    }
+  ]
 
-  getCheckedFilters (filter: string) {
-    return (this as any)[filter]
-  }
+    qiskitCalendarSyncLink = 'https://qisk.it/calendar-sync'
+    qiskitCalendarLink: GeneralLink = {
+      url: 'https://qisk.it/calendar',
+      label: 'https://qisk.it/calendar',
```
eddybrando, line 214: ~This label should be more helpful to the user. It could be something like "Calendar of Qiskit events".~

Actually, `label` is not part of the anchor elements' attributes and it doesn't have any effect to set it: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a?retiredLocale=de#attributes

We should be using the `title` attribute instead.

But this is pointing to a bigger issue that might be affecting multiple links in our page and that is out of the scope of this pull request. We might want to create an issue to address this.

pages/events/index.vue
```diff
@@ -83,9 +83,38 @@
...
+                  <p class="event-page__sync">
+                    Start by copying the calendar subscription link
+                    <cv-code-snippet
+                      kind="inline"
+                      feedback-aria-label="Copy calendar sync url"
+                    >
+                      <span>{{ qiskitCalendarSyncLink }}</span>
+                    </cv-code-snippet>
+                  </p>
+                  <component :is="calendar.component" class="event-page__instructions" />
```
eddybrando, line 110: For this particular case, I think that having the tabs in this component might be easier to maintain instead of creating 3 new components with very little content and no logic or style.

If we want to do this, we could achieve it like this:

```suggestion
                  <div class="event-page__instructions">
                    <template v-if="calendar.name === 'Google'">
                      <cv-list ordered>
                        <cv-list-item>
                          Open your Google Calendar by visiting <AppLink url="https://calendar.google.com/calendar/u/0/r/settings/addbyurl?pli=1">
                            this link
                          </AppLink>
                        </cv-list-item>
                        <cv-list-item>Paste the events subscription link you copied above</cv-list-item>
                        <cv-list-item>Click "Add Calendar"</cv-list-item>
                      </cv-list>
                    </template>
                    <template v-else-if="calendar.name === 'Outlook'">
                      <cv-list ordered>
                        <cv-list-item>Open your Outlook Calendar and click on “Add calendar”</cv-list-item>
                        <cv-list-item>Select the menu item “Subscribe from web”</cv-list-item>
                        <cv-list-item>Paste the events subscription link you copied above</cv-list-item>
                        <cv-list-item>Add a name for your calendar</cv-list-item>
                        <cv-list-item>Click “Import”</cv-list-item>
                      </cv-list>
                    </template>
                    <template v-else>
                      <cv-list ordered>
                        <cv-list-item>Open your Apple Calendar</cv-list-item>
                        <cv-list-item>In the menu, select File > New Calendar Subscription</cv-list-item>
                        <cv-list-item>Paste the events subscription link you copied above</cv-list-item>
                        <cv-list-item>Click "Subscribe"</cv-list-item>
                        <cv-list-item>Click "Ok"</cv-list-item>
                      </cv-list>
                    </template>
                  </div>
```

eddybrando (changes requested):
Thanks a lot Randy! 🙂 

I just added a couple of comments.

pages/events/index.vue
techtolentino: I'm only using `label` because that's what it's called in the `GeneralLink` [interface](https://github.com/Qiskit/qiskit.org/blob/main/constants/appLinks.ts#L8-L10)

Just pointing out that we're _not_ using it as an anchor tag attribute, but only to render the text of the anchor... 

Maybe the interface could be refactored? You're right though, it'll involve a lot of changes across the whole application.

pages/events/index.vue
techtolentino: Updated in https://github.com/Qiskit/qiskit.org/pull/2955/commits/95fe85daeb5f39f4ece6f4de2aef27381c5d1cd5

pages/events/index.vue
techtolentino: FYI, had a discussion w/ @y4izus on this, and decided to keep the calendar instructions separate from this page component. I agree that there's little content w/ no logic or style... but we'll still have these benefits

- easy to find/maintain, since it's cleanly separated into their own files
- doesn't pollute the parent page (which will continue to grow as we potentially add the actual calendar as well)

The main ask was to rename the component and restructure how they were organized, done here https://github.com/Qiskit/qiskit.org/pull/2955/commits/2308580cf6a79698c60a1279bca8f12fb85995de

pages/events/index.vue
```diff
@@ -83,9 +83,38 @@
...
+            <p class="event-page__section__description">
+              <!-- eslint-disable-next-line vue/singleline-html-element-content-newline -->
+              Stay up to date with all of our scheduled events by following our calendar. You can view the calendar by visiting <AppLink v-bind="qiskitCalendarLink">{{ qiskitCalendarLink.label }}</AppLink>, or subscribe to it by adding to the calendar app of your choice.
+            </p>
+            <div class="event-page__tabs">
+              <cv-tabs aria-label="Calendar applications">
+                <cv-tab
+                  v-for="(calendar, calendarIndex) in calendars"
+                  :id="`tab-${calendar.name}`"
+                  :key="calendarIndex"
```
y4izus, line 97: To use index as key is a bad practice, you can read the reason [here](https://stackoverflow.com/questions/44531510/why-not-always-use-the-index-as-the-key-in-a-vue-js-for-loop#:~:text=Because%20arrays%20are,using%20an%20index.). Although in this case we will not have that issue because we are not mutating the array, I think we should use good practice and use the `name` instead of the `index` as `key`

y4izus: And if you use the `name` as `key`, you can do this:

```ts
v-for="{ name } in calendars"
```

pages/events/index.vue
```diff
@@ -156,57 +193,81 @@ export default class EventsPage extends QiskitPage {
     }
   ]
 
-  get noEvents (): boolean {
-    return (this as any).filteredEvents.length === 0
-  }
+  calendars = [
```
y4izus, line 196: I think `calendarsInstructions` or something similar could be more semantic

y4izus (changes requested):
Thank you Randy! just some minor changes :D

pages/events/index.vue
techtolentino: I did a few things to update this:

- renamed `calendars` to `calendarsInstructions`
- inside that array, instead of calling it a `component` property, I'm now calling it `instructions`
  - reason: so that on [this line](https://github.com/Qiskit/qiskit.org/pull/2955/files#diff-e89e9da7098c8b828fffda21f9b905c5668d8726f92332ddd3612c55d7bc0d9eR110), I can avoid writing `<component :is="component">`

- updated to `v-for="{ name, instructions } in calendars"`


please let me know if this makes sense

y4izus: makes totally sense, thank you!

y4izus (approved):

eddybrando (approved):

PR #2961, opened 2023-02-06
feat: removes the store

### LINKED ISSUE #2946: Vue 3 update: Replace Vuex with Pinia

y4izus:
## Changes

Fix #2946 

The store was no needed, so it has been deleted instead of replacing vuex with pinia

## Implementation details

- Logic from the stores has been moved to where it's used (components or pages)
- Types from the stores have been moved to a separate files inside the `./types/` folder
- Store tests have been removed


pages/events/index.vue
```diff
@@ -157,39 +134,95 @@ const extraFilters = [
...
+const { data: pastEvents } = useLazyAsyncData(
+  'fetch-past-events',
+  async () => await import('~/content/events/past-community-events.json') as CommunityEvent[]
+)
 
-const updateWholeFilter = (filter: string, filterValues: string[]) => {
-  const { commit } = this.$store
-  const payload = { filter, filterValues }
+const activeSet = ref('upcoming')
+const regionFilters = ref([] as string[])
```
eddybrando, line 148: This type assertion can be replaced with a regular type.

```suggestion
const regionFilters = ref<string[]>([])
```

pages/events/index.vue
```diff
@@ -157,39 +134,95 @@ const extraFilters = [
...
+const typeFilters = ref([] as string[])
```
eddybrando, line 149: This type assertion can be replaced with a regular type.

```suggestion
const typeFilters = ref<string[]>([])
```

pages/ecosystem/index.vue
```diff
@@ -129,7 +128,31 @@ useHead({
...
+  'fetch-members',
+  async () => await import('~/content/ecosystem/members.json')
+)
+
+const { data: tiers } = useLazyAsyncData(
+  'fetch-tiers',
+  async () => await import('~/content/ecosystem/tiers.json')
+)
+
+const tierFilters = ref([] as string[])
```
eddybrando, line 141: This type assertion can be replaced with a regular type.

```suggestion
const tierFilters = ref<string[]>([])
```

eddybrando (changes requested):
Thanks a lot @y4izus! It looks very good!

I added a couple of comments.

We'll also need another functional test once the migration to Nuxt 3 is a little more stable.

And as discussed offline, we'll need to update the README and ARCHITECTURE markdown files later. But since we're doing a lot of changes as part of this migration, we prefer to update them at the very end.

pages/events/index.vue
y4izus: I don't mind using one or the other, but right know we are mixing both types 😅 . I would be nice to choose one of them

eddybrando (approved):

Issue #2902, opened 2023-01-05
[Events] Add event description to event tile


javabster:
### Feature description

It would be great if the event tiles on qiskit.org/events had a 1 sentence description for what the events are. 

### Additional context

The tiles on qiskit.org/ecosystem page uses a version of the same tile component as the events page but it does have space for a description, so we could probably use the same component with a similar implementation

y4izus:
Note for devs:
- We are already using the component `EventCard` that is prepared for having a description (see [here](https://qiskit.org/events/summer-school/#:~:text=The%20Qiskit%20Global%20Summer%20School%20returns%20as%20a%20two%2Dweek%20intensive%20course%20focused%20on%20Quantum%20Simulations%20and%20more!))
- We need to pull the event description from the Airtable and show it on the card 
- The column name is "Event Description"
- The Airtable already have content, so running `npm run generate` on local will pull info to check on your machine how it will look

PR #2965, opened 2023-02-06
feat(events): add description to EventCard

techtolentino:
## Changes

Resolves https://github.com/Qiskit/qiskit.org/issues/2902

## Implementation details

- pulls in event abstract field

## How to read this PR

- review how `event.abstract` string is passed into the `EventCard` (as slot)

## How to review locally
```shell
npm run generate
npm run start
```

## Screenshots

https://user-images.githubusercontent.com/6276074/217284332-2ac0f39d-593a-4250-99d2-21a88dcdc5f0.mov

### DIFF
--- a/components/events/EventCard.vue
+++ b/components/events/EventCard.vue
@@ -9,8 +9,8 @@
     :segment="segment"
     :vertical-layout="verticalLayout"
   >
-    <div v-if="description" class="event-card__description">
-      {{ description }}
+    <div class="event-card__description">
+      <slot v-if="$slots.default" />
     </div>
     <div>
       <div v-if="location" class="event-card__detail">
@@ -44,7 +44,6 @@ export default class EventCard extends Vue {
   @Prop(String) title!: string
   @Prop(String) image!: string
   @Prop({ type: String, default: '' }) institution!: string
-  @Prop({ type: String, default: '' }) description!: string
   @Prop(String) location!: string
   @Prop(String) date!: string
   @Prop(String) time!: string
--- a/components/events/EventCard.vue
+++ b/components/events/EventCard.vue
@@ -9,8 +9,8 @@
     :segment="segment"
     :vertical-layout="verticalLayout"
   >
-    <div class="event-card__description">
-      <slot v-if="$slots.default" />
+    <div v-if="description" class="event-card__description">
+      {{ description }}
     </div>
     <div>
       <div v-if="location" class="event-card__detail">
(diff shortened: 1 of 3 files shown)


components/events/EventCard.vue
```diff
@@ -9,8 +9,8 @@
     :segment="segment"
     :vertical-layout="verticalLayout"
   >
-    <div class="event-card__description">
-      <slot v-if="$slots.default" />
+    <div v-if="description" class="event-card__description">
```
y4izus, line 12: You don't need to change this component, it's already prepare to have the `description`. We did it with slots instead of passing the description just in case that description needed to including something else (links, ...). If we have just a string, we lose that capability

y4izus (changes requested):
Just revert one thing and update how you pass the description on the events page and it's done!! thank you!

components/events/EventCard.vue
techtolentino: Updated in this commit https://github.com/Qiskit/qiskit.org/pull/2965/commits/848504a19790f0d2634aa9adacb76e141951bdde

techtolentino:
@JRussellHuffman — I really believe that in order to have consistent cards rendered, we'd need consistent _inputs_ /content.

The components try their best to handle what's given to them, but the same component can't handle a wide variety of content passed to it... so I agree that we need to enforce some sorta guidelines around what's acceptable in terms of 

- image (for example: provide dimensions, suggest centered)
- title (aim for character or word length?)
- description (aim for word length or line count?)

doing things like this ☝🏼 would get us the _most_ uniformity out of the cards... 


__
to answer your UI request, would a solution like this work? 👇🏼 

https://user-images.githubusercontent.com/6276074/218546942-3d42d352-b1db-4851-b55d-5d2fa911709d.mov


otherwise, if we always wanted the event info (type, location, time + CTA) to always be at the bottom... we'd need to give the description some sort of minimum height.


https://user-images.githubusercontent.com/6276074/218547167-cca050d1-7970-45dd-92b6-284818f3939b.mov

JRussellHuffman:
Would it be possible to enforce consistent input/content? In other words, can we put character count limits and required fields in airtable (where I assume the input is coming from?

Regarding the UI, I have two thoughts on the screen cap below:
1. Can there be more of a margin around the pill/tag to the right?
2. Can the blank space be between the description text and the other information (eg. location, date, etc)?
Screen Shot 2023-02-13 at 3 44 20 PM

y4izus:
Pero offline conversations, we decided to merge this changes and rethink the new card structure with more calm

y4izus (approved):

PR #2967, opened 2023-02-10
Adjust nuxt configuration

### LINKED ISSUE #2947: Vue 3 update: Adjust Nuxt configuration

y4izus:
## Changes

Fix #2947 

First steps to adjust nuxt config. We will need to configure more things in the future, but I think this is a good start point

## Implementation details

- Updates nuxt.config file
- Updates tsconfig file
- Comment some code and add some TODO to have in mind in the future


eddybrando (approved):
Thanks a lot @y4izus! This will helps us a lot moving forward.

As discussed offline, we can leave the Node upgrade for later to avoid extra complexity and to continue focusing on the bare minimum changes to get the app running on Nuxt 3.

PR #2971, opened 2023-02-13
feat: nuxt 3 minimal working app

eddybrando:
## Changes

This PR makes the necessary changes to have a working Nuxt 3 app.

A lot of the code is commented out so that the app is able to load, since we still need to make some adjustments to migrate our application from Nuxt 2.

## Implementation details

Commented out a lot of code. Also moved the "plugins" directory to "plugins.old" until we adapt it to work on Nuxt 3.

The config files have been moved to "old-config-files" for later review and integration (if needed).

The original package.json has been moved to package.json.old for reference so that we can start adding the necessary dependencies.


techtolentino (approved):
Thank you @eddybrando 🚀

PR #2978, opened 2023-02-15
style: configure lint for Nuxt 3

### LINKED ISSUE #2950: Vue 3 update: Adjust lint

eddybrando:
## Changes

This PR configures ESLint, Prettier and Stylelint. It also solves easily fixable issues. Other more complex issues will have to be solved in future PRs involving the migration of those files.

Closes #2950

## Implementation details

Configured a small ESLint, Prettier and Stylelint setup. The files ".eslintrc.js" and ".stylelintrc.js" contain these configurations.

## How to read this PR

1. Review the new config files
2. Review the package JSON scripts
3. See the changes related to the automatic lint fixes


.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  parser: "vue-eslint-parser",
+  parserOptions: {
+    parser: "@typescript-eslint/parser",
+  },
+  extends: ["@nuxtjs/eslint-config-typescript", "plugin:prettier/recommended"],
```
y4izus, line 6: I think we should add `"extends": "eslint:recommended"` to the `extends` section. Read [this](https://eslint.org/docs/latest/use/getting-started#configuration:~:text=%7B-,%22extends%22%3A%20%22eslint%3Arecommended%22,-%7D)

.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  parser: "vue-eslint-parser",
```
y4izus, line 2: It would be nice to add this config too:
```
env: {
    browser: true,
    node: true,
  },
```

what do you think?

y4izus (approved):
I have two questions (not blockers, you can merge :D )

.eslintrc.js
eddybrando: Added in https://github.com/Qiskit/qiskit.org/pull/2978/commits/919baa0ede9a4b58bf5215337b7ef0698558d879

PR #2981, opened 2023-02-15
[prototype] Providers content

techtolentino:
⚠️ please do not merge ⚠️ 

### What is this PR?
- testing the content structure that's being uploaded by @HuangJunye 
- updating our component template, types, and interfaces where necessary
- prototyping the providers content just to become aware of any issues we might have

### Branch preview
https://qiskit-org-pr-2981.dcq4xc5i083.us-south.codeengine.appdomain.cloud/providers/

### Current screenshot [WIP]

https://user-images.githubusercontent.com/6276074/219172111-12f0a30d-05b3-48c6-b898-996e86ea3777.mov

PR #2985, opened 2023-02-20
[prototype] Providers w/ Content

techtolentino:
⚠️ pls do not merge yet ⚠️ 

## Changes
- Collaborating w/ @HuangJunye to model the latest Providers page content
- Work is still in progress

## Preview branch
https://qiskit-org-pr-2985.dcq4xc5i083.us-south.codeengine.appdomain.cloud/providers


constants/providersContent.ts
```diff
@@ -50,84 +52,171 @@ const REAL_HARDWARE: ProvidersSection = {
...
-      title: 'IBM Quantum Qiskit Runtime',
-      content: {
-        image: '/images/providers/applications/optimization.png',
-        description: 'The Qiskit Optimization package covers the whole range from high-level modeling of optimization problems, with automatic conversion of problems to different required representations, to a suite of easy-to-use quantum optimization algorithms that are ready to run on classical simulators, as well as on real quantum systems.',
-        cta: {
-          label: 'Solve the Max Cut Problem',
-          url: 'https://qiskit.org/documentation/optimization/tutorials/01_quadratic_program.html'
-        }
-      }
+      title: 'IBM Qiskit Runtime',
```
HuangJunye, line 57: I am OK with putting content directly to `.ts` But I also want to let you know that the long term plan is to be able to populate these fields using the Ecosystem infrastructure. People should be able to submit their own providers as an Qiskit Ecosystem project with extra requirement specific to providers (e.g. hello world examples). With that in mind, would it be better to write the content in `.ts` or in json similar to the ecosystem?

HuangJunye (commented):
Thanks a lot for making the prototype. This looks great!

constants/providersContent.ts
techtolentino: Oh, this is good information to know... 

@y4izus what do you think? 

Seems like, at some point (whether no or in another PR), we should refactor the code so we can accept JSON... especially if it will be similar to how the Ecosystems pages are populated.

Thanks @HuangJunye

HuangJunye:
I noticed that the installation code cell doesn't render correctly. I remember that it was working in https://github.com/Qiskit/qiskit.org/pull/2981. Possibly something merged in `main` broke this. 

Screenshot 2023-02-20 at 19 25 01

Also, the text in this cell should take the content from "installation" field for example `pip install qiskit-ibm-runtime`.

HuangJunye:
Can we disable scrolling of the hello world example code cell? The code cell appears to scale with the number of code lines so there is nothing to scroll for. 


https://user-images.githubusercontent.com/7631333/220178581-5d1f2b98-1f02-420d-81d4-5a1cb978b7e7.mov

The horizontal position of scroll bar is right after the longest code line which varies. If we do use scroll bar, I think it would be better for it to show up on the rightmost position in all instances.

Screenshot 2023-02-20 at 19 28 41
Screenshot 2023-02-20 at 19 28 49

HuangJunye:
Something for @1ucian0 and @JRussellHuffman to discuss. Looking at the preview, I think it is better to change the flow of the content inside each card.

current flow
- pip install
- image
- description
- links
- hello world example


Screenshot 2023-02-20 at 19 45 05

proposed flow
- image
- description
- pip install
- hello world example
- links
image

1ucian0:
I'm okay with it

JRussellHuffman:
Also agreed on moving the links to the bottom. What do you all think about moving the top? I don't entirely know the use case, but at a glance, I think it makes sense to have the description upfront before they get to the code.

techtolentino:
Had a sync w/ @y4izus and @HuangJunye this morning, about the prototype and this PR:

Summary:
- for the content structure, we'll go w/ JSON for now
- the `helloworldExample` code will be an array of strings
- in the UI, the provider content will follow this order:
  - description
  - image
  - pip install
  - hello world example
  - links
  
  Any design update or layout refactoring can happen after we get this ☝🏼  working and merged. 
  
  Expect a new PR soon, to reflect these decisions.

constants/providersContent.ts
y4izus: Per offline conversations we decided to do the JSON file now

PR #3006, opened 2023-03-01
Update 'Entanglement in action' preview image

frankharkins:
## Changes

Updates the preview image as the page is no longer under construction.


techtolentino (approved):
Very nice, thanks you @frankharkins

PR #3011, opened 2023-03-03
exclude ecosystem/ subdirectories from rclone sync

1ucian0:
## Changes

exclude ecosystem/ subdirectories from rclone sync, so does not get affected in the next sync.

## Implementation details

In order to extend the qiskit ecosystem with some of the component documentation (like `q.o/ecosystem/qiskit-nature`), the subdirectories in `ecosystem/` should be excluded from `rclone sync`

Thanks @mtreinish for the help!


deploy/static-sites.txt
```diff
@@ -1,2 +1,3 @@
 documentation/**
-textbook/**
\ No newline at end of file
+textbook/**
+ecosystem/**
```
mtreinish, line 3: So the one thing to check here is if this will conflict with updating https://qiskit.org/ecosystem/ from this repo (AFAIK that page is still managed here). This file gets used with `--exclude-from` in rclone which means this will not sync any files which match any expression here. This is the correct syntax to expand external (relative to this repo) of the qiskit.org/ecosystem path, because we use `rclone sync` to ensure deleted paths from the build get removed on the host, this is necessary to enable usage of the subdir outside of this repo as otherwise rclone would delete externally managed files as their not in the source side of the sync. But we should check if this doesn't block updates to the existing page, and if it does if we can adjust this filter to enable updates from here but also not blocking external usage of the subdir.

The docs on filtering in rclone are here:  https://rclone.org/filtering/

1ucian0, line 3: I think this will exclude subdirectories, but not the files. Am I right @mtreinish ?
```suggestion
ecosystem/*/
```

1ucian0: other option is to move to `--filter-from`, but not sure.

eddybrando (approved):
I believe this should achieve what's expected:

1. The `qiskit.org/ecosystem` page will still be created from `pages/ecosystem.vue` in this repository
2. Other files like `pages/ecosystem/foo.vue` will generate a static file `dist/ecosystem/foo/index.html` but the files inside subdirectories of `dist/ecosystem/` won't be copied to production, hence no `qiskit.org/ecosystems/foo` page will be available if coming from this repository
3. We will be able to add other pages like `qiskit.org/ecosystems/foo` from other repositories and those pages won't be deleted or otherwise modified by deployments from this repository

PR #3022, opened 2023-03-10
refactor(events): migrate to Nuxt 3

### LINKED ISSUE #3002: Vue 3 update: Migrate events pages

eddybrando:
## Changes

This PR makes the necessary changes to the "Events" page to make it work with Nuxt 3.

Closes #3002


techtolentino (changes requested):
This is a great upgrade @eddybrando 

Only question I had was if you planned on including the Calendar tab, with the iFrame 

👏🏼

eddybrando:
> This is a great upgrade @eddybrando
> 
> Only question I had was if you planned on including the Calendar tab, with the iFrame
> 
> 👏🏼

I think that we can do another PR at the end where we add these new changes, if that's ok.

Issue #3047, opened 2023-03-22
[Providers] Update Quick Start widget installation content


techtolentino:
### Feature description

After reviewing with @1ucian0, we need to make a couple of updates to the homepage Quick Start feature. 

Currently, the Installation that is shown in the Providers Quick Start, is a single line, but should be updated to multiline

#### How it currently renders
![Screenshot 2023-03-22 at 9 19 20 AM](https://user-images.githubusercontent.com/6276074/226970319-5098405f-556c-40d2-b42f-6f24d6d2693e.png)

#### How it should render (two lines)
```
pip install 
qiskit-ibm-provider
```

### Acceptance criteria
- [x] This should be updated to render as two lines
- [x] For this task, we can update our local json files. 
    - [Example of installation code location ](https://github.com/Qiskit/qiskit.org/blob/feat/providers/content/providers/cloud-simulators.json#L10-L12) for the Provider page data
    - [Example of installation code location ](https://github.com/Qiskit/qiskit.org/blob/feat/providers/content/providers/quick-start.json#L6-L8) for the Provider homepage Quick Start widget
    - Please note that there are multiple instances of these, across several files within `/content/providers/*.json`

### Additional context

Going forward, this should be updated on @HuangJunye's scripts that generates these installation code lines

1ucian0:
The box should render:
```
pip install qiskit 
pip install qiskit-ibm-provider
```

The first `pip install qiskit` should always be there. While it is not technically necessary, I think it helps to drive the point that this box is about installing qiskit *with* something.

PR #3053, opened 2023-03-22
Update provider installation code to multiline

techtolentino:
## Changes

Resolves https://github.com/Qiskit/qiskit.org/issues/3047
Co-author: @eddybrando 

## Implementation details

- Updates the JSON so that the component can recognize how the installation code should be multiline 
- cc: @HuangJunye (we'll need to update the json script to output this way)

## Branch preview
https://qiskit-org-pr-3053.dcq4xc5i083.us-south.codeengine.appdomain.cloud/

### DIFF
--- a/content/providers/quick-start.json
+++ b/content/providers/quick-start.json
@@ -4,6 +4,7 @@
     "image": "/images/providers/qiskit-ibm-runtime.png",
     "description": "Qiskit Runtime is a quantum computing service and programming model that allows users to optimize workloads and efficiently execute them on quantum systems at scale. The programming model extends the existing interface in Qiskit with a set of new primitive programs.",
     "installation": [
+      "pip install qiskit",
       "pip install qiskit-ibm-runtime"
     ],
     "websiteCta": {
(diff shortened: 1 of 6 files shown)


HuangJunye (changes requested):
Install of adding the same `pip install qiskit` to every provider entry of the json files, can we add that in the source code? The installation code block should always include `pip install qiskit`.

content/providers/quick-start.json
```diff
@@ -766,11 +776,11 @@
...
     "description": "Time dependent open quantum systems solver.",
     "installation": [
+      "pip install qiskit",
       "git clone https://github.com/0tt3r/QuaC-qiskit", 
-      "cd QuaC-qiskit",
-      "pip install"
+      "cd QuaC-qiskit"
     ],
-    "websiteCta": {
+     "websiteCta": {
```
eddybrando, line 783: It seems like this change was unintended:

```suggestion
    "websiteCta": {
```

HuangJunye (approved):
Looks great. Thanks for fixing this!

PR #3108, opened 2023-04-06
fix: last fixes for Nuxt 3 migration

### LINKED ISSUE #2952: Vue 3 update: Fix other things that break because of the change 😬

eddybrando:
## Changes

This PR makes the last fixes created during the Nuxt 3 migration.

Closes #2952


assets/scss/main.scss
```diff
@@ -5,10 +5,17 @@
 @include carbon.grid-css-grid;
 @include carbon.grid-flex-grid;
 
+@font-face {
```
techtolentino, line 8: I'm wondering if we need to include the `.ttf` file in our code base, or if we can use a similar approach to what they do in saiba.

Screenshot 2023-04-10 at 1 17 04 PM

techtolentino (changes requested):
Just had one comment about Plex

Thanks @eddybrando

assets/scss/main.scss
eddybrando: Sure, we can do that. Changed in https://github.com/Qiskit/qiskit.org/pull/3108/commits/d5c8c4739ccfd44328918d8cfc267c0461e54466.

techtolentino (approved):
Thank you @eddybrando 

This all LGTM 👏🏼

PR #3116, opened 2023-04-12
fix: content generation on production

eddybrando:
## Changes

This change makes the production build fetch the content as expected.


y4izus (approved):
thank you!

techtolentino (approved):

PR #3123, opened 2023-04-13
replace "&nbsp;" for "" in json code

1ucian0:
When coping from the code boxes, a non-printable char is introduced. 

Maybe comes from the "&nbsp;"?


y4izus:
Conversations happening offline, please, don't do anything here yet :)

PR #3126, opened 2023-04-14
converting content/providers/*.json to YAMLs

1ucian0:
using the included script `python convert.py file.json` that creates `file.yaml` in same place.


y4izus (approved):
Thank you!

PR #3138, opened 2023-04-18
Feat/providers content files tests

eddybrando:
## Changes

This PR adds robust testing to the YAML content files to ensure that the required content for the providers feature exists and is in the expected format.


eddybrando:
Replaced by #3139

PR #3139, opened 2023-04-18
test(providers): content files

eddybrando:
## Changes

This PR adds robust testing to the YAML content files to ensure that the required content for the providers feature exists and is in the expected format.

### DIFF
--- a/content/ecosystem/members.json
+++ b/content/ecosystem/members.json
@@ -1,5 +1,1327 @@
 [
   {
+    "name": "quantuminspire",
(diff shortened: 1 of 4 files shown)


y4izus (approved):
For now this PR is ok for me, but I think we should have a discussion to talk about the guidelines we want to follow with vitest. For example, I think the `.spec` on the name of the file is a remain from when we were using Jest. What do you think?

Please, resolve the conflicts before merge

PR #3149, opened 2023-04-20
fix(segment): SSR compatibility

eddybrando:
## Changes

This PR fixes the issues with Segment during SSR, since `window` and `document` (both accessed by Segment) are not available in the server.

It also introduces the **trackClick** directive, similar to Saiba.


y4izus (approved):
Awesome! Thank you!

PR #3151, opened 2023-04-20
fix: fixed design on emptyCards

mariacloehb:
## Changes

Changed the design on the EmptyCards, previously taking up more space than needed, and now they've been adjusted to the right width. 
Changes have been applied to the index.vue file in the events folder. 

@techtolentino contributed to this issue.

## Implementation details
Changes have been made around the Ui Card regarding empty events, inside the if-else condition of event cards in index.vue.

## Screenshots

As seen in the screenshot, the width is no longer occupying all the screen, it now is the same design as regular events. 

Screenshot 2023-04-20 at 15 04 48

### DIFF
--- a/pages/events/index.vue
+++ b/pages/events/index.vue
@@ -64,7 +64,7 @@
         <template #results>
           <div v-if="noEvents" class="cds--row">
             <div class="cds--col-sm-4 cds--col-xlg-8">
-              <UiCard
+              <UiCard :image="emptyCard.img" :title="emptyCard.title">
                 :image="emptyCard.img"
                 :title="emptyCard.title"
               >
--- a/pages/events/index.vue
+++ b/pages/events/index.vue
@@ -62,15 +62,18 @@
           </div>
         </template>
         <template #results>
-          <UiCard
-            v-if="noEvents"
-            :image="emptyCard.img"
-            :title="emptyCard.title"
-          >
-            <div class="event-page__empty-card-description">
-              {{ emptyCard.description }}
+          <div v-if="noEvents" class="bx--row">
+            <div class="cds--col-sm-4 cds--col-xlg-8">
+              <UiCard
+                :image="emptyCard.img"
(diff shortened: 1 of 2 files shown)


pages/events/index.vue
```diff
@@ -62,15 +62,18 @@
...
         </template>
         <template #results>
-          <UiCard
-            v-if="noEvents"
-            :image="emptyCard.img"
-            :title="emptyCard.title"
-          >
-            <div class="event-page__empty-card-description">
-              {{ emptyCard.description }}
+          <div v-if="noEvents" class="bx--row">
```
techtolentino, line 65: Since our migration to Nuxt 3, we updated our Carbon packages. This will now be `class="cds--row"`, instead of `class="bx--row"`

techtolentino (changes requested):
@mariacloehb thanks for the PR

just a couple of notes to fix.

pages/events/index.vue
```diff
@@ -62,15 +62,18 @@
...
+            <div class="cds--col-sm-4 cds--col-xlg-8">
+              <UiCard
```
techtolentino, line 67:
```suggestion
              <UiCard :image="emptyCard.img" :title="emptyCard.title">
```

To get rid of the lint error (in this particular case), we should place the component props on one line

pages/events/index.vue
techtolentino, line 65:
```suggestion
          <div v-if="noEvents" class="cds--row">
```

Since migrating to Nuxt 3, we now use a newer Carbon library; therefore,  any grid class that used to begin with `bx--`  will now use `cds--`

techtolentino (changes requested):
@mariacloehb just two changes... feel free to commit them once you understand the reasons.

Everything else looks good - thank you

techtolentino (approved):
Thanks @mariacloehb 👏🏼

PR #3164, opened 2023-04-28
fix(events): prevent timezones issues using UTC

eddybrando:
## Changes

In some very specific scenarios, timezones differences might result in some [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) comparisons to fail in our functions that process the events data fetched form Airtable. This PR helps to avoid that by ensuring that we're working in UTC.

## Implementation details

* Set "today" to start at 00:00:00 UTC.
* Explicitly set UTC in some cases where it was missing.
* Changed the test file to align with the realistic date format that we get from Airtable, which includes the timezone in the date string.


y4izus (approved):
OMG!! Good catch Eddy!! Thank you!

Issue #2163, opened 2021-09-16
Qiskit org 404


techtolentino:
Following the implementation of Qiskit/platypus#559, we can use this implementation to add one for qiskit org.

techtolentino:
Moving this out of our current priorities because it doesn't make sense to give this focus/time, while we're migrating to Nuxt.

techtolentino:
Nuxt 3 has now migrated and this is ready to be worked on.

PR #3200, opened 2023-05-09
feat/error-page

techtolentino:
## Changes

Closes https://github.com/Qiskit/qiskit.org/issues/2163

## Implementation details

- add error page
- add assets
- remove `404.html`
- use `error` Object in props, leveraging the `error.statusCode` and `error.message`, instead of hardcoding the `statusCode`

## How to test
- to test `404`: try to navigate to a URL that doesn't exist
- to test a `500`: go to `members.json` and change an object `key`
  - (in my example, I altered members.json to have a key property of `"testssResults"`, instead of `"testsResults"`)
  - then visit `localhost:3000/ecosystem`

## How to read this PR

- Please review the `error.vue` page of implementation and logic
- Refresh the page to see the random messages and images rendering
- Check CTAs to ensure they work

## Screenshots

Showing dynamic `error.statusCode`s and `error.message`s

https://github.com/Qiskit/qiskit.org/assets/6276074/f90aa123-854a-41c9-9c35-03b3af21e9ec


y4izus (changes requested):
Following the [documentation for rendering an error page](https://nuxt.com/docs/getting-started/error-handling#rendering-an-error-page), we can read that this `error.vue` page, [receives a prop with the code of the error](https://nuxt.com/docs/getting-started/error-handling#rendering-an-error-page:~:text=This%20page%20has%20a%20single%20prop%20%2D%20error%20which%20contains%20an%20error%20for%20you%20to%20handle). That means, that this page is used for every error and on this PR is only used for the 404. 

I think this is not a good implementation because we are showing a 404 page for every error we can have. I've talked with @abdonrd about this and he agrees.

Since this is not a priority this sprint and requires more research time, I suggest to do the changes once you finish your priorities

error.vue
```diff
@@ -0,0 +1,127 @@
...
+          <div class="cds--col-lg-4">
+            <h1 class="error-page__title">{{ props.error?.statusCode }}</h1>
+            <p>
+              {{ props.error?.message }}
+            </p>
+            <p class="error-page__message">
+              Or as we like to say,
+              <!-- eslint-disable vue/no-v-html -->
+              <span v-html="randomMessage" />
+              <!-- eslint-enable -->
```
eddybrando, line 16: We can use "eslint-disable-next-line" instead of disabling and enabling again:

```suggestion
              <!-- eslint-disable-next-line vue/no-v-html -->
              Or as we like to say, <span v-html="randomMessage" />
```

error.vue
```diff
@@ -0,0 +1,127 @@
...
+            </p>
+
+            <UiCta
+              :label="BackToHomeCTA.label"
+              :segment="BackToHomeCTA.segment"
+              :url="BackToHomeCTA.url"
```
eddybrando, line 22: We usually camel-case the variables:

```suggestion
              :label="backToHomeCta.label"
              :segment="backToHomeCta.segment"
              :url="backToHomeCta.url"
```

error.vue
```diff
@@ -0,0 +1,127 @@
...
+            />
+            <UiCta
+              class="error-page__ghost-btn"
+              :label="BackToDocsHome.label"
+              :segment="BackToDocsHome.segment"
+              :url="BackToDocsHome.url"
```
eddybrando, line 28: We usually camel-case the variables:

```suggestion
              :label="backToDocsCta.label"
              :segment="backToDocsCta.segment"
              :url="backToDocsCta.url"
```

error.vue
```diff
@@ -0,0 +1,127 @@
...
+            <img
+              class="error-page__img"
+              :src="errorImgSrc"
+              alt="Playful illustration of a cat in a cardboard box"
+            />
+          </div>
+        </div>
+      </div>
+    </div>
+    <FooterComponent old-container theme="light" />
```
eddybrando, line 42: The "old-container" variant doesn't match the width of the top navbar, and the "light" theme is the default value, so we don't need to pass it:

```suggestion
    <FooterComponent />
```

We could also remove the `theme="light"` attribute from these files:

- layouts/default-max.vue
- layouts/default.vue

error.vue
```diff
@@ -0,0 +1,127 @@
...
+  </div>
+</template>
+
+<script setup lang="ts">
+import { GeneralLink } from "./constants/appLinks";
```
eddybrando, line 47: "GeneralLink" has changed since the PR was first opened, so I changed it for the new "Link".
I also added here the import of the "NuxtError" type.

```suggestion
import { NuxtError } from "#app";
import { Link } from "~/types/links";
```

error.vue
```diff
@@ -0,0 +1,127 @@
...
+
+const props = defineProps({
+  // eslint-disable-next-line vue/require-default-prop
+  error: Object,
+});
```
eddybrando, line 52: We can be more precise with the type of the error object:

```suggestion
interface Props {
  error: NuxtError;
}

const props = defineProps<Props>();
```

error.vue
```diff
@@ -0,0 +1,127 @@
...
+
```
eddybrando, line 53: We can define the error message as the page title as well:

```suggestion

const pageTitle = computed<string>(() => {
  return props.error.statusMessage ?? "Error";
});

useHead({
  title: pageTitle,
});

```

error.vue
```diff
@@ -0,0 +1,127 @@
...
+const BackToHomeCTA: GeneralLink = {
```
eddybrando, line 54: We usually camel-case the variables:

```suggestion
const backToHomeCta: Link = {
```

error.vue
```diff
@@ -0,0 +1,127 @@
...
+  url: "/",
+  label: "Back to Qiskit.org home",
+  segment: {
+    cta: "back-to-home",
+    location: "error-page",
+  },
+};
+
+const BackToDocsHome: GeneralLink = {
```
eddybrando, line 63: We usually camel-case the variables:

```suggestion
const backToDocsCta: Link = {
```

error.vue
```diff
@@ -0,0 +1,127 @@
...
+
+const messages = [
+  "this isn’t exactly a <em><strong>superposition</strong></em> to be in.",
+  "looks like something here is an <em><strong>entangled</strong></em> mess.",
+  "we apologize for the <em><strong>interference</strong></em> you’re seeing on this page.",
+  "you’ve reached the edge of the multi-verse.",
+  "this page loads fine another dimension, but not this dimension.",
+];
+
+const randomMessage = computed(() => {
```
eddybrando, line 88: Just to improve readability of the expected type:

```suggestion
const randomMessage = computed<string>(() => {
```

error.vue
```diff
@@ -0,0 +1,127 @@
...
+            <UiCta
+              class="error-page__ghost-btn"
+              :label="BackToDocsHome.label"
+              :segment="BackToDocsHome.segment"
+              :url="BackToDocsHome.url"
+              kind="ghost"
+            />
+          </div>
+          <div class="cds--col-lg-8">
+            <img
```
eddybrando, line 33: I think that we could use NuxtImg here to optimise the cats.

eddybrando (changes requested):
Thanks a lot @techtolentino! The page is looking really good!

Here are a couple of comments 🙂

error.vue
techtolentino: cats have been optimized 

😸

Issue #1997, opened 2021-07-19
Google search thumbnail text


JRussellHuffman:
### Steps to reproduce the problem
Google search "Qiskit text beta"

### What is the current behavior?
The text reads "No information is available for this page."
<img width="351" alt="Screen Shot 2021-07-19 at 11 53 16 AM" src="https://user-images.githubusercontent.com/7633881/126197530-c14c009c-4012-47bb-b289-3d5250687952.png">

### What is the expected behavior?
We should put some text here. Let's discuss what that text should be below.

eddybrando:
I'm currently getting this, which isn't much better:

<img width="403" alt="Screenshot 2021-10-09 at 15 41 00" src="https://user-images.githubusercontent.com/22047320/136660223-d8dcaa29-71f1-445c-920c-b076e81cd971.png">

y4izus:
Since that page doesn't exist anymore and we cannot add the description to it, what do you suggest to do @eddybrando ?

eddybrando:
This appears to be an issue caused by the way we're currently redirecting from /textbook-beta to /learn with a [meta redirect](https://www.w3.org/TR/WCAG20-TECHS/H76.html) instead of a [server 301 redirect](https://www.semrush.com/blog/301-redirects/).

I'll check and see if it is possible and if it makes sense to use server redirects for this and similar pages, and check what other signals we can give the search engines to stop that page from ranking instead of /learn.

eddybrando:
I opened [a PR](https://github.com/Qiskit/qiskit.org/pull/3204) that aims to reduce the time that it takes for the redirections to happen. This might result in the redirection pages to be replaced by their targets in the search result pages.

There is no way to know if this works immediately, but we should try it out and see if there are any changes in the next days, and if not, continue looking for alternatives.

eddybrando:
I've additionally opened another PR to handle the redirects server-side. If that one gets merged, I'll create a following PR to remove the client-side redirections.

CC @1ucian0

PR #3204, opened 2023-05-11
fix(seo): remove layouts for faster redirects

eddybrando:
## Changes

This PR removes the use of the default layout for the pages that are used only for redirection.

Related to #1997

## Implementation details

We have some pages that we keep to redirect users to other pages.

The redirections are set to happen as soon as possible (0ms) but currently the client is trying to load the layout first, which includes the standard header and footer, which take some time to load. Only after the layout is loaded, the redirection happens.

This delay might be the reason why search engines like Google are still ranking these pages even though they are 301s and only their target should be indexed.

By not using any Nuxt layout, we can expect the redirections to happen faster and maybe the 301 pages will be removed from the SERPs.

### DIFF
--- a/pages/learn/intro-qc-qh.vue
+++ b/pages/learn/intro-qc-qh.vue
@@ -6,3 +6,12 @@
     />
   </head>
 </template>
+
+<script setup lang="ts">
+definePageMeta({
+  layout: false,
+  pageTitle:
+    "2020 Qiskit global summer school on quantum computing and quantum hardware",
+  routeName: "introduction-to-quantum-computing-and-quantum-hardware-2020",
+});
+</script>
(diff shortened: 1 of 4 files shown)


techtolentino (approved):
@eddybrando very nice, and thanks for the explanation. 

I think at some point, we should just totally remove the `/textbook-beta/summer-school/quantum-computing-and-quantum-learning-2021.vue`

I say that because anything that resided at `/textbook-beta/` was kinda short lived, and it was only being used as a temporary solution. Plus, now that we have a better `/learn` page, _and_ Platypus is up and running, do we need to still support this one link?

(Unless this one link was widely shared 🤷🏼‍♂️)

Issue #2852, opened 2022-10-19
Bug: App Card labels overlap title


techtolentino:
### Steps to reproduce the problem

1. Go to the [Ecosystem page](https://qiskit.org/ecosystem)
2. Scroll down to the bottom and find `QuantumCircuits.jl`
4. On desktop, see how the labels almost collide with the title 

### What is the current behavior?

On desktop, see how the labels almost collide with the title 

### What is the expected behavior?

It'd be an improvement if the title showed fully, with spacing, while the labels wrapped to the next line

### Screenshots

![Screen Shot 2022-10-19 at 8 01 30 AM](https://user-images.githubusercontent.com/6276074/196728674-4776dd7e-648f-4c01-99e2-de30d051f124.png)

### Browser

desktop

### Additional information

**Initially seen on LinkedIn**
<img width="1082" alt="Screen Shot 2022-10-19 at 7 57 08 AM" src="https://user-images.githubusercontent.com/6276074/196728491-1b726c40-3f93-4451-a48e-2fb3fdec3657.png">

eddybrando:
![Screenshot 2023-04-14 at 17 26 27](https://user-images.githubusercontent.com/22047320/232087318-a57523cc-028b-427e-897a-ffa12822731e.png)

That specific issue has been fixed. The labels now are shown bellow the title.

Nevertheless, spacing of the labels doesn't look ideal. I'll work on that.

techtolentino:
how about on desktop or LG breakpoints and up? mine is showing this:

![Screenshot 2023-04-14 at 8 49 59 AM (2)](https://user-images.githubusercontent.com/6276074/232092992-635dff98-1dbd-421e-b3cb-d36190aaa161.png)

mariacloehb:
@techtolentino And myself were discussing on a few things regarding this issue, and we'd like to suggest considering having the card tags in their own row, under the card title, and having them left aligned. The original layout didn't take into account the possibility of having multiple tags on the tip right corner, making the tags overline with the title. 

Here are some proposals:
· all tags would be in its own row below the card title
· we could update the purple color for the tags with the light purple shown in the [Carbon Tag Component](https://web-components.carbondesignsystem.com/?path=/story/components-tag--default).

these proposals would also work nicely with the other cards (Events, Advocates)
@y4izus @JRussellHuffman

y4izus:
[...]

PR #3205, opened 2023-05-11
feat(advocates): update layout with tab sections

### LINKED ISSUE #3194: [Advocates] Get QAMP info from Airtable and show it on the page
Once https://github.com/Qiskit/qiskit.org/issues/3192 and https://github.com/Qiskit/qiskit.org/issues/3193 are done, we can connect the UI with the data coming from Airtable

**Acceptance criteria**
- [x] Get the data from airtable
- [x] Show the data on the page
- [x] Test should still working
- [x] Add more tests if you think it's necessary 

---
Part of https://github.com/Qiskit/qiskit.org/issues/3055

techtolentino:
## Changes

Closes #3194
Co-authored-by: @eddybrando

## Implementation details

- removed "Meet the Advocates" blurb
- added tab sections to introduce QAMP project tab
- displayed the fetched data from Airtable

## Preview

https://qiskit-org-pr-3205.dcq4xc5i083.us-south.codeengine.appdomain.cloud/advocates


pages/advocates.vue
```diff
@@ -31,3 +58,20 @@ useHead({
   ],
 });
 </script>
+
+<style lang="scss" scoped>
+@use "~/assets/scss/carbon.scss";
+@use "~/assets/scss/helpers/index.scss" as qiskit;
```
eddybrando, line 64: I believe there is no need for this import:

```suggestion
```

types/advocates.ts
```diff
@@ -25,4 +25,19 @@ interface Advocate {
   slackUsername?: string;
 }
 
-export { AdvocatesWorldRegion, ADVOCATES_WORLD_REGION_OPTIONS, Advocate };
+interface QampProject {
+  name: string;
+  url: string;
+  website?: string;
+  tags: string[];
```
eddybrando, line 32: "tags" is a little bit too generic for me 🤔... In this case, I think that we could call it "cohort" since we already know we're only display cohorts as tags in the cards.

Also, we're only display one single cohort, so we don't need an array.

types/advocates.ts
```diff
@@ -25,4 +25,19 @@ interface Advocate {
   slackUsername?: string;
 }
 
-export { AdvocatesWorldRegion, ADVOCATES_WORLD_REGION_OPTIONS, Advocate };
+interface QampProject {
+  name: string;
+  url: string;
+  website?: string;
```
eddybrando, line 31: I believe we're only display one URL which is the GitHub issue, right?

If that's the case, I think that `githubIssueUrl: string` might be more precise.

types/advocates.ts
```diff
@@ -25,4 +25,19 @@ interface Advocate {
...
+  tags: string[];
+  mentors?: string[];
+  mentees?: string[];
+  contributors?: string[];
```
eddybrando, line 35: The should all be required, right?

An empty string would also be valid if, for instance, there are no extra contributors in a given project.

Also, I think "contributors" are called "collaborators" in our database. It might be better to reflect that here too.

types/advocates.ts
techtolentino: we can update this, though look at this comment in Figma, from @HuangJunye 

> _The card should contain project title, project categories tags, mentors names, mentees names, collaborators (non-advocates team members) names, project gitihub link. Expand to see project descriptions?_

it mentions the possibility of having project category tags... maybe they should be separated?

Cohort in the top right corner
Project cagetories in its own row?

We're also exploring how we can update/improve the Ecosystem card, in [this conversation](https://github.com/Qiskit/qiskit.org/issues/2852#issuecomment-1545778983)

types/advocates.ts
techtolentino: These are great questions! 

Are you able to see what the data looks like already? Maybe we can sync on what the mock data shape could look like.. 

This is why I was trying to ask if we can get a more ideal shape of the project object. 

(Remember, I was mentioning to you and @y4izus that I'm going off of assumptions in this initial PR, and it's a little difficult because I could potentially be implementing things that will just have to be undone 😬 )

types/advocates.ts
y4izus: I would suggest following the design decided for the ecosystem page, having:
- `cohort: string` on the top right corner. We should ask Gemma if a project belongs to only one cohort and, if it can be in more than one, we should have `cohort: string[]`
- `categories: string[]` on the row after the title for showing the categories

y4izus: We need to know if a project can belong to more than 1 cohort, if that is the case and we restrict the number of main tags to 1, we should probably rethink the card

y4izus: Per offline conversations, a project will only have one tag to show one cohort

y4izus (approved):
LGTM!

PR #3209, opened 2023-05-12
feat: remove redirect pages

### LINKED ISSUE #1997: Google search thumbnail text

eddybrando:
We now have server-side redirection in place, so we no longer need client-side redirection for these pages.

## Changes

This PR removes the pages that were used for client-side redirection since we now have server-side redirection in place for those routes.

This change should also result in search engines understanding that these routes are 301 redirects and that they shouldn't be indexed as individual search results.

Closes #1997

## Implementation details

As mentioned in https://github.com/Qiskit/qiskit.org/issues/1997#issuecomment-1544076829, I created a PR to add server-side redirects to the pages that were being redirected client-side. Since that PR got merged, now we have proper server redirection in place and we can remove these pages that were only used for redirection.

You can check that the server-side redirection is in place by visiting a site like https://www.redirect-checker.org/ to test each route.

### DIFF
--- a/pages/learn/intro-qc-qh.vue
+++ b/pages/learn/intro-qc-qh.vue
@@ -1,8 +1,0 @@
-<template>
-  <head>
-    <meta
-      http-equiv="refresh"
-      content="0; URL=/learn/summer-school/introduction-to-quantum-computing-and-quantum-hardware-2020"
-    />
-  </head>
-</template>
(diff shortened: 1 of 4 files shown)


y4izus (approved):
LGTM! Thank you!

PR #3210, opened 2023-05-12
feat: use semantic HTML images

### LINKED ISSUE #3190: SEO: Use semantic HTML images instead of background images
### Feature description

We should try and use image tags (``) instead of CSS images (`<div style="background-image:url(example-image.jpg)">`) to make them crawlable and processable by search engines.

### Additional context

- Information from Google about this topic: https://developers.google.com/search/docs/appearance/google-images#help-us-discover-all-your-images

eddybrando:
## Changes

This PR replaces the uses of CSS background images with semantic HTML image tags. Only images of the background grid are left ass CSS background images.

In order to optimise the images for UX and SEO, this PR also introduces the use of [Nuxt Image](https://v1.image.nuxtjs.org/).

Closes #3190

## Implementation details

- Installed Nuxt Image and configured the Carbon breakpoints for image size optimisation
- Replaced CSS background images with semantic HTML `` tags via the NuxtImage component
- Replaced other images with NuxtImage components
- Whenever possible (due to minimal quality loss), use webp as the preffered format due to its optimised compression for web
- In the Metal Capabilities Section, we were no longer using both videos and images (we're now only using videos). Removed the checks for images


components/Landing/HeroMoment/LandingHeroMomentComponent.vue
```diff
@@ -48,6 +54,7 @@ const getStartedLink: GeneralLink = {
   segment: { cta: "get-started", location: "hero-moment" },
 };
 
+// TODO: Fix hydration text content mismatch
```
eddybrando, line 57: This refers to a warning displayed in the console when in the development environment.

The issues is out of the scope of this PR but it's good to keep track of it.

techtolentino (commented):
thanks for this PR @eddybrando - this is a good update... 

just one question
Should there be standard values for the `sizes`? It looks like we use various px values depending on the use case.

I'm wondering if these serve the same purpose as Carbon breakpoints 🤔

eddybrando:
> Should there be standard values for the `sizes`? It looks like we use various px values depending on the use case.

It would be nice, but if we want to optimise as much as possible we really need to handle each case separately. Sometimes it's not only the size defined by the width but also the perceived quality loss due to the compression at certain sizes, which certain shapes and certain colours, so we need to manually test each one if we want great results.

We'll never be perfect (😢) but we can spend a couple of minutes when introducing new images to trying and figuring out good sizes for optimising at the breakpoints.

components/Ui/UiCard.vue
```diff
@@ -6,14 +6,16 @@
...
-        'background-image': `url(${image})`,
-      }"
-    />
+    <div v-if="image" class="card__image-container">
+      <nuxt-img
+        class="card__image"
+        :class="imageContain ? 'card__image_contain' : null"
+        format="webp"
+        loading="lazy"
+        sizes="sm:300px md:650px lg:250px"
```
techtolentino, line 15: https://github.com/Qiskit/qiskit.org/assets/6276074/c6caa245-be3c-498a-a196-8819fa13a45a

Question: is this happening to you? (For the UiCard that's used on the event pages, some of them appear a little blurry to me.)

techtolentino (approved):
I've reviewed this a few times; all my questions have been addressed


Thanks @eddybrando - this is looking great.

PR #3212, opened 2023-05-12
Fix missing spaces in External Recommended readings

### LINKED ISSUE #3211: Bug: missing spaces in External Recommended readings
### Where does the issue happen?

https://qiskit.org/learn/course/basics-quantum-information/

### What is the content issue?

In some of the External Recommended readings, we're missing spaces after the `,`

### Screenshots

Screenshot 2023-05-12 at 7 36 50 AM

### Additional information

_No response_

tungbq:
## Changes
- Add missing space to pages/learn/course/basics-quantum-information.vue
- Closes #3211

### DIFF
--- a/components/Learn/LearnExternalRecommendedReadings.vue
+++ b/components/Learn/LearnExternalRecommendedReadings.vue
@@ -77,6 +77,7 @@ withDefaults(defineProps<Props>(), {
 
   &__title {
     font-style: italic;
+    margin-left: carbon.$spacing-01;
   }
 
   &__reference {
--- a/pages/learn/course/basics-quantum-information.vue
+++ b/pages/learn/course/basics-quantum-information.vue
@@ -65,7 +65,7 @@ many that cover this material`;
 const links: RecommendedReading[] = [
   {
     url: "https://www.khanacademy.org/math/linear-algebra",
-    author: "Khan Academy,",
+    author: "Khan Academy, ",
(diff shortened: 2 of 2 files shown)


techtolentino (changes requested):
Thank you @tungbq - this is one version of the fix. After looking around other courses that have External Recommended Readings, I'm noticing that it's [happening elsewhere](https://qiskit.org/learn/course/quantum-states-and-qubits/) as well.

Are you comfortable w/ making a CSS change? I'm thinking what if we just add some `padding` or `margin` so that the same treatment can handle all cases?

What are your thoughts?

tungbq:
> Thank you @tungbq - this is one version of the fix. After looking around other courses that have External Recommended Readings, I'm noticing that it's [happening elsewhere](https://qiskit.org/learn/course/quantum-states-and-qubits/) as well.
> 
> Are you comfortable w/ making a CSS change? I'm thinking what if we just add some `padding` or `margin` so that the same treatment can handle all cases?
> 
> What are your thoughts?

@techtolentino Thanks for reviewing my changes and the suggestion! Yes, I agree with you that we need to improve the fix to cover all case by using CSS. I've update the latest patch to cover this, please have a look, thanks.

techtolentino (approved):
Checked this out locally

nicely done 👏🏼 

thank you @tungbq

PR #3214, opened 2023-05-14
Bug: bakends compatibility should be backend compatibility

tarunsamanta2k20:
## Changes
Not sure when this happened but I see that the title of the provider component on the homepage was changed to say "backends compatibility". I don't think it's grammatically correct to use a plural noun before the word "compatibility". Instead it should be "backend compatibility" or if you want to emphasise the plurality of backends then I would recommend "compatible with multiple backends"

Screenshot 2023-05-02 at 4 55 44 PM


techtolentino (commented):
This sounds correct to me

@javabster or @1ucian0 just checking w/ you in case you had any input

Thanks

javabster:
LGTM 🚀

techtolentino (approved):
LGTM and @javabster 

🚀

PR #3244, opened 2023-05-22
feat(qamp): fetch data

### LINKED ISSUE #3193: [Advocates] Research "Get info from a Relational DB on Airtable"
The structure of the Airtable we have to use for the QAMP page is a relational one, meaning that the info on some columns points to another table. We will need to access all the info, so we need to check how to access it

**Acceptance criteria**
- [x] Find out how to get info from a relational airtable. You can share how to do it here or on our slack channel

Note: If you need to see the airtable, please contact me

----

Part of https://github.com/Qiskit/qiskit.org/issues/3055

eddybrando:
## Changes

This PR implements the fetching and conversion of the QAMP data from Airtable.

Part of #3194

## Implementation details

Since the data from the "Projects" table references data from the "Cohorts" and "Participants" tables, we needed to create classes with methods to fetch that data as well. They are all very similar in implementation to what we already had in place for "Advocates" and "Events".

We've also added some tests for the new classes.

## How to read this PR

- Start with the **types/qamp.ts** file to get an idea of the shape of the QAMP objects.
- Check **content/qamp/qamp-projects.json** to see an example of the fetched and converted QAMP data.
- See the **nuxt.config.ts** that now also starts the fetching of the QAMP projects data.
- Check the **hooks/update-qamp-projects.ts** file.
- You can also see the small type improvement in **hooks/update-advocates.ts**.
- Review the **hooks/qamp-*-conversion-utils.ts** files that implement the fetching and conversion of the data. You can compare them to the existing file for advocates, which is very similar.
- Finally, check the test files

### DIFF
--- a/content/qamp/qamp-projects.json
+++ b/content/qamp/qamp-projects.json
@@ -1,15 +1,6 @@
 [
   {
     "cohort": {
-      "name": "QAMP Spring 23"
-    },
-    "collaborators": [],
-    "mentees": [],
-    "mentors": [],
-    "name": "#00 Other (my preferred project is not listed)"
-  },
-  {
-    "cohort": {
       "name": "QAMP Fall 22"
     },
     "collaborators": [],
@@ -21,38 +12,6 @@
       }
     ],
     "name": "#1 Add global phase gate to Qiskit Terra"
-  },
-  {
-    "cohort": {
-      "name": "QAMP Spring 23"
-    },
-    "collaborators": [],
-    "githubIssueUrl": "https://github.com/qiskit-advocate/qamp-spring-23/issues/1",
-    "mentees": [],
-    "mentors": [
-      {
-        "name": "Iskandar Sitdikov"
-      }
-    ],
-    "name": "#01 PurpleCaffeine: tracking of quantum programs and experiments"
-  },
-  {
-    "cohort": {
-      "name": "QAMP Spring 22"
-    },
-    "collaborators": [],
-    "githubIssueUrl": "https://github.com/qiskit-advocate/qamp-spring-22/issues/1",
-    "mentees": [
-      {
-        "name": "Vishnu Ajith"
-      }
-    ],
-    "mentors": [
-      {
-        "name": "Jack Woehr"
-      }
-    ],
-    "name": "#01 Transpiler Hackathon"
   },
   {
     "cohort": {
@@ -70,48 +29,6 @@
   },
   {
     "cohort": {
-      "name": "QAMP Spring 23"
-    },
-    "collaborators": [],
-    "githubIssueUrl": "https://github.com/repos/qiskit-advocate/qamp-spring-23/issues/2",
-    "mentees": [],
-    "mentors": [
-      {
-        "name": "Atsushi Matsuo"
-      }
-    ],
-    "name": "#02 Update `GroverOptimizer` in Qiskit Optimization to the latest version of the implementation"
-  },
-  {
-    "cohort": {
-      "name": "QAMP Spring 22"
-    },
-    "collaborators": [],
-    "githubIssueUrl": "https://github.com/qiskit-advocate/qamp-spring-22/issues/2",
-    "mentees": [],
-    "mentors": [
-      {
-        "name": "Julien Gacon"
-      }
-    ],
-    "name": "#02 VQE optimization with dynamic shot scheduling"
-  },
-  {
-    "cohort": {
-      "name": "QAMP Spring 23"
-    },
-    "collaborators": [],
-    "githubIssueUrl": "https://github.com/repos/qiskit-advocate/qamp-spring-23/issues/3",
-    "mentees": [],
-    "mentors": [
-      {
-        "name": "Atsushi Matsuo"
-      }
-    ],
-    "name": "#03 Implement Date-reuploading classifier in Qiskit Machine Learning"
-  },
-  {
-    "cohort": {
       "name": "QAMP Fall 22"
     },
     "collaborators": [],
@@ -123,37 +40,6 @@
       }
     ],
     "name": "#3 Implement virtual distillation, an error mitigation technique, in Qiskit"
-  },
-  {
-    "cohort": {
-      "name": "QAMP Spring 22"
-    },
-    "collaborators": [],
-    "githubIssueUrl": "https://github.com/qiskit-advocate/qamp-spring-22/issues/3",
-    "mentees": [],
-    "mentors": [
-      {
-        "name": "Dariusz Lasecki"
-      },
-      {
-        "name": "Max Rossmannek"
-      }
-    ],
-    "name": "#03 Migrate AdaptVQE from Qiskit Nature to Terra"
-  },
-  {
-    "cohort": {
-      "name": "QAMP Spring 22"
-    },
-    "collaborators": [],
-    "githubIssueUrl": "https://github.com/qiskit-advocate/qamp-spring-22/issues/4",
-    "mentees": [],
-    "mentors": [
-      {
-        "name": "Atsushi Matsuo"
-      }
-    ],
-    "name": "#04 Add performance benchmarks for Qiskit Machine Learning"
   },
   {
     "cohort": {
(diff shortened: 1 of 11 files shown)


content/qamp/qamp-projects.json
```diff
@@ -0,0 +1,1232 @@
+[
+  {
+    "cohort": {
+      "name": "QAMP Spring 23"
```
techtolentino, line 4: Is there any chance that we can rename these to `Spring 2023`? or does the team want the `QAMP Spring 23` format?

Only asking because users will already be on the `QAMP` tab, and they understand what they're looking at. Having "QAMP" in all the filters, along with all the card tags, can equate to unnecessary noise, in my opinion.

content/qamp/qamp-projects.json
```diff
@@ -0,0 +1,1232 @@
...
+        "name": "Matthew Treinish"
+      }
+    ],
+    "name": "#30 Improvements to Red Queen"
+  },
+  {
+    "cohort": {
+      "name": "QAMP Spring 23"
+    },
+    "collaborators": [
```
techtolentino, line 946: I could be mistaken, but out of all 76 entries, this is the only project that has collaborators... is that strange to anyone else?

techtolentino (commented):
Nicely done @eddybrando - I just had a few questions... but this is great!

content/qamp/qamp-projects.json
eddybrando: I like to proposal. It would keep the text simple.

Here we're only displaying the data that we get from the Airtable base. This means that the team in charge of updating that data is the one that could make this change. It would then be reflected here the next time the production build is generated.

content/qamp/qamp-projects.json
eddybrando: It does seem weird, I agree 😬, but I did check and this is what is currently stored in Airtable as well. We can ask the team in charge of the Airtable base data to be sure.

hooks/qamp-projects-conversion-utils.ts
```diff
@@ -0,0 +1,126 @@
...
+      this.recordFields = await this.getAllFieldNames(RECORD_FIELDS_IDS);
+    }
+
+    // Fetch QAMP cohorts
+    const qampCohortsAirtableRecords = new QampCohortsAirtableRecords(
+      this.apiKey
+    );
+    this.qampCohorts = await qampCohortsAirtableRecords.fetchCohorts();
+
+    // Fetch QAMP participants
```
y4izus, line 51: Same 👆

hooks/qamp-projects-conversion-utils.ts
```diff
@@ -0,0 +1,126 @@
...
+    this.qampCohorts = qampCohorts || [];
+    this.qampParticipants = qampParticipants || [];
+  }
+
+  async fetchProjects(): Promise<QampProject[]> {
+    if (!this.recordFields) {
+      this.recordFields = await this.getAllFieldNames(RECORD_FIELDS_IDS);
+    }
+
+    // Fetch QAMP cohorts
```
y4izus, line 45: I don't think this comment adds new information, the variable name is really clear

hooks/qamp-participants-conversion-utils.ts
```diff
@@ -0,0 +1,60 @@
...
+      });
+    return Promise.resolve(qampParticipants);
+  }
+
+  convertToQampParticipants(record: any): QampParticipant {
+    const qampParticipant = {
+      id: this.id,
+      name: this.getName(record),
+    };
+    return qampParticipant;
```
y4izus, line 52: You can directly return the object:

```suggestion
    return {
      id: this.id,
      name: this.getName(record),
    };
```

tests/hooks/qamp-cohorts-conversion-utils.spec.ts
```diff
@@ -0,0 +1,56 @@
+import { beforeEach, describe, expect, test } from "vitest";
+import QampCohortsAirtableRecords from "../../hooks/qamp-cohorts-conversion-utils";
```
y4izus, line 2:
```suggestion
import QampCohortsAirtableRecords from "~/hooks/qamp-cohorts-conversion-utils";
```

tests/hooks/qamp-participants-conversion-utils.spec.ts
```diff
@@ -0,0 +1,57 @@
+import { beforeEach, describe, expect, test } from "vitest";
+import QampParticipantsAirtableRecords from "../../hooks/qamp-participants-conversion-utils";
```
y4izus, line 2:
```suggestion
import QampParticipantsAirtableRecords from "~/hooks/qamp-participants-conversion-utils";
```

tests/hooks/qamp-projects-conversion-utils.spec.ts
```diff
@@ -0,0 +1,123 @@
+import { beforeEach, describe, expect, test } from "vitest";
+import QampProjectsAirtableRecords from "../../hooks/qamp-projects-conversion-utils";
```
y4izus, line 2:
```suggestion
import QampProjectsAirtableRecords from "~/hooks/qamp-projects-conversion-utils";
```

y4izus (changes requested):
Thank you!! Just a few of small changes :D

content/qamp/qamp-projects.json
```diff
@@ -0,0 +1,1232 @@
+[
+  {
+    "cohort": {
+      "name": "QAMP Spring 23"
+    },
+    "collaborators": [],
+    "mentees": [],
```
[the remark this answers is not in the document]
eddybrando: That is a good point. There are some projects that were only in the "suggestion" phase but didn't get to be a QAMP 2023 project, so we might need to add some filtering logic.

I also agree that we should continue developing this feature and adjust the final logic once we get the final feedback from the events team.

hooks/qamp-cohorts-conversion-utils.ts
```diff
@@ -0,0 +1,60 @@
...
+      });
+    return Promise.resolve(qampCohorts);
+  }
+
+  convertToQampCohort(record: any): QampCohort {
+    const qampCohort = {
+      id: this.id,
+      name: this.getName(record),
+    };
+    return qampCohort;
```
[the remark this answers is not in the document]
eddybrando: Changed in https://github.com/Qiskit/qiskit.org/pull/3244/commits/524298bf17ddd442f639fdeb640adb2450515cdc

hooks/qamp-projects-conversion-utils.ts
```diff
@@ -0,0 +1,126 @@
...
+  convertToQampProject(record: any): QampProject {
+    const qampProject = {
+      cohort: this.getCohort(record),
+      collaborators: this.getCollaborators(record),
+      githubIssueUrl: this.getGithubIssueUrl(record),
+      mentees: this.getMentees(record),
+      mentors: this.getMentors(record),
+      name: this.getName(record),
+    };
+    return qampProject;
```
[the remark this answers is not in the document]
eddybrando: Changed in https://github.com/Qiskit/qiskit.org/pull/3244/commits/524298bf17ddd442f639fdeb640adb2450515cdc

hooks/qamp-participants-conversion-utils.ts
eddybrando: Changed in https://github.com/Qiskit/qiskit.org/pull/3244/commits/524298bf17ddd442f639fdeb640adb2450515cdc

tests/hooks/qamp-participants-conversion-utils.spec.ts
eddybrando: I'm afraid this doesn't work because the when running the tests, we don't have the whole Nuxt context that sets this directory alias, so it fails to find that import.

hooks/qamp-projects-conversion-utils.ts
eddybrando: Removed in https://github.com/Qiskit/qiskit.org/pull/3244/commits/d634d6656166106d38c1066e5ccaca57583a065a

content/qamp/qamp-projects.json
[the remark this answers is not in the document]
y4izus: We already have an answer, so could you update the query to adapt it to the new requirements?

PR #3258, opened 2023-05-26
feat(qgss23): disable register cta

y4izus:
## Changes

Disable the registration CTA

## Implementation details

- Disables the CTA
- Changes the text to "General Registration June 1, 9AM EST"
- Adapts the with of the CTA to fit the content

https://github.com/Qiskit/qiskit.org/assets/17231966/9dcdc0a3-891d-4f68-b764-e9653eedad27

### DIFF
--- a/pages/events/summer-school-2023.vue
+++ b/pages/events/summer-school-2023.vue
@@ -28,7 +28,11 @@
           for more details. For any questions, please check out our FAQ below!
         </p>
         <p>See you soon!</p>
-        <UiCta class="summer-school-page__cta" v-bind="header.cta" />
+        <UiCta
+          class="summer-school-page__cta summer-school-page__cta_disabled"
+          v-bind="header.cta"
+          kind="secondary"
+        />
       </template>
       <template #card>
         <EventsItemCard v-bind="headerData.card" vertical-layout>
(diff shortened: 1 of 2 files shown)


pages/events/summer-school-2023.vue
```diff
@@ -222,6 +227,16 @@ const helpfulResourcesData = helpfulResources;
...
+
+    &_disabled {
+      pointer-events: none;
+      color: qiskit.$text-color-light;
+      width: fit-content;
+
+      &:deep(.cta__icon) {
+        display: none;
+      }
+    }
```
eddybrando, line 239: I'm not sure about how I feel about this implementation. Even though it does the job fine from a visual aspect, it renders kind of weird HTML in my opinion:

Screenshot 2023-05-26 at 14 07 21

I think that `href="false"` and `rel="false"` are invalid values. This might confuse screen readers and other assistive technologies.

Overall I think that since this element's link attribute are being disabled, this should be just a "div" or "span" instead of a modified `UiCta` component.

techtolentino: Interesting find @eddybrando - I didn't realize this before, and this is how we've been doing it for the past couple of years, _I think_.

I agree - we don't have to use the `UiCta`.

techtolentino: After discussing with team, since this is such a short-lived change (will revert back in a week or less), we can go with this implemenation.

Issue #3273, opened 2023-05-30
[Ecosystem] Show tiers as tabs


eddybrando:
### Feature description

On the [Ecosystem page](https://qiskit.org/ecosystem/), we're currently showing the tiers as a set of checkboxes that acts as a filter:

![image](https://github.com/Qiskit/qiskit.org/assets/22047320/4c65c9b2-f619-48e6-8364-2236123237f7)

We now want to display them as tabs:

![image](https://github.com/Qiskit/qiskit.org/assets/22047320/f805c483-35af-4262-8991-321fbfe820a4)

### Additional context

~Design: https://www.figma.com/file/N34caXM86bY1778Whk1c31/2503-ecosystem-page?type=design&node-id=705-7376&t=ME2tYW6PZjJE7YQi-0~

Updated design: https://www.figma.com/file/N34caXM86bY1778Whk1c31/2503-ecosystem-page?type=design&node-id=768-43517&t=u88xjTPfPl9L7xuu-0

Issue #2927, opened 2023-01-20
Ecosystem search


woodsp-ibm:
### Feature description

Provide a feature to search within the Ecosystem to locate resources of interest.

E.g. I was interested to see if an HHL algorithm was available via the Ecosystem, but had no way to search. Such a search perhaps can also include health checks, e.g. maybe I check a box if I want to include only assets whose tests are passing - having some function that no longer works is something I may not be interested in if I just want to use said function directly.

Perhaps it can start out simple but as the ecosystem grows having a flexible, comprehensive search to locate resources of interest seems highly desirable, if not essential.

PR #3278, opened 2023-05-31
feat(ecosystem): move tiers to tabs and add search bar

y4izus:
## Changes

Fix https://github.com/Qiskit/qiskit.org/issues/3273
Fix https://github.com/Qiskit/qiskit.org/issues/2927

## Implementation details

- Now the tiers are on tabs
- A search bar is added. Right now is only searching on the card description

## How to read this PR

Go to https://qiskit-org-pr-3278.dcq4xc5i083.us-south.codeengine.appdomain.cloud/ecosystem

- Tiers tabs should work
- Each tab only shows the members that are on that tier
- Search works both for searching on the description and on the title

**Out of scope**
- A solution when there are no results on the search

## Screenshots

Captura de pantalla 2023-05-31 a las 15 28 32

### DIFF
--- a/pages/ecosystem.vue
+++ b/pages/ecosystem.vue
@@ -29,7 +29,11 @@
       />
       <div class="ecosystem__tiers">
         <client-only>
-          <bx-tabs trigger-content="Select an item" value="Main">
+          <bx-tabs
+            trigger-content="Select an item"
+            value="Main"
+            @bx-tabs-selected="selectTab($event.target.value)"
+          >
             <bx-tab
               v-for="tierName in tiersNames"
               :id="`tab${tierName}`"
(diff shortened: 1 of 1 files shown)


pages/ecosystem.vue
```diff
@@ -27,85 +27,97 @@
...
+          </bx-tabs>
+          <div
+            v-for="tierName in tiersNames"
+            :id="`panel${tierName}`"
+            :key="tierName"
+            class="ecosystem__tier-panel"
+            role="tabpanel"
+            :aria-labelledby="`tab${tierName}`"
+          >
+            <bx-data-table page-size="2" start="0">
```
techtolentino, line 55: @y4izus as mentioned in our discussion, we can remove the `bx-data-table` and simply go with [the Search component](https://web-components.carbondesignsystem.com/?path=/story/components-search--default&knob-Direction%20(direction%20in%20%3Cbx-overflow-menu-body%3E)=bottom&knob-Helper%20text%20(helper-text)=Optional%20helper%20text&knob-Label%20text%20(label-text)=Dropdown%20title&knob-Overflow%20menu%20size%20(size)=xl&knob-The%20default%20content%20of%20the%20trigger%20button%20(trigger-content)=Select%20an%20item)

```vue
<bx-search
  class="ecosystem__search"
  placeholder="Search using keywords like algorithms, simulator, or machine learning"
  @bx-search-input="searchOnMembers($event.detail.value)"
/>
```

In this example above, I added the class of `ecosystem__search` so we can override the `gray` colors, with our own `cool gray` colors. Not sure if there's a better way, but this will work, lower in the same file:

```scss
<style lang="scss">
.ecosystem {
  &__search {
    --cds-field-01: #{carbon.$cool-gray-10};
    --cds-field-04: #{carbon.$cool-gray-30};
  }
}
</style>
``` 

(just an example)

y4izus: you mean outside the tier tabs too?

y4izus: Done:
- 300b06efb8074113ee7154a95a6425f327a074e0 : Updates component and styles
- db153287a848752f8c85544586450fbb7e7390b5 : Move the search bar outside the tier panels

PR #3285, opened 2023-06-01
[WIP] Prepare for more generic Preview Deploy GitHub Action

Eric-Arellano:
First part of https://github.com/Qiskit/qiskit.org/issues/3283.

This preps for the gh-actions repo removing the special casing of qiskit.org in favor of the generic `build_args` input. 

But until then, this change has no impact; the special-casing will still be used so we need to still set the `secrets` section.

--

We have to set the secrets as env vars to be able to pass the values to `with:`. GitHub Actions scrubs secrets and env vars set by them with e.g. logging, so this should be safe and is the recommended approach.

While this results in some awkward duplication, I think it's worth it so that we can keep gh-actions generic. For example, it will make it much easier for us to add future env vars without needing to update gh-actions.


Eric-Arellano:
I couldn't get this working due to issues like https://github.com/Qiskit/qiskit.org/actions/runs/5147162083.

Issue #3045, opened 2023-03-22
[ecosystem] Sort card


1ucian0:
The cards are shown in random order, which is ok for a default view (they are reshuffled, I think, weekly). It would be great to be able to sort them by:
 * stars
 * alphabetically
 * tier (and alphabetically inside)
 * downloads (information not currently available)

PR #3286, opened 2023-06-02
Sort ecosystem cards

y4izus:
## Changes

Fix https://github.com/Qiskit/qiskit.org/issues/3045

## Implementation details

Adds a dropdown to choose between two options for sorting: name (asc) and stars (desc)

## Preview
https://qiskit-org-pr-3286.dcq4xc5i083.us-south.codeengine.appdomain.cloud/ecosystem

## How to read this PR

- Go to the ecosystem page
- Check if the sort is working as expected

### DIFF
--- a/pages/ecosystem.vue
+++ b/pages/ecosystem.vue
@@ -27,7 +27,7 @@
         :label="joinAction.label"
         :url="joinAction.url"
       />
-      <div class="ecosystem__tiers">
+      <div class="ecosystem__tiers cds--grid cds--no-gutter">
         <client-only>
           <bx-tabs
             trigger-content="Select an item"
@@ -45,13 +45,14 @@
             </bx-tab>
           </bx-tabs>
           <div class="ecosystem__toolbar cds--row">
-            <bx-search
-              class="ecosystem__search cds--col-lg-14 cds--col-md-6"
-              placeholder="Search using keywords like algorithms, simulator, or machine learning"
-              @bx-search-input="searchOnMembers($event.detail.value)"
-            />
+            <div class="ecosystem__search cds--col-lg-14 cds--col-md-6">
+              <bx-search
+                placeholder="Search using keywords like algorithms, simulator, or machine learning"
+                @bx-search-input="searchOnMembers($event.detail.value)"
+              />
+            </div>
             <bx-dropdown
-              class="cds--col-lg-2 cds--col-md-2"
+              class="ecosystem__sort-dropdown cds--col-lg-2 cds--col-md-2"
               trigger-content="&#x21c5; Sort by ..."
               @bx-dropdown-selected="setSortValue($event.detail.item.value)"
             >
@@ -274,9 +275,14 @@ const joinAction: Link = {
     margin-top: carbon.$spacing-06;
   }
 
-  &__search {
+  &__search,
+  &__sort-dropdown {
     --cds-field-01: #{carbon.$cool-gray-10};
     --cds-field-04: #{carbon.$cool-gray-30};
+  }
+
+  &__sort-dropdown {
+    --cds-text-01: #{carbon.$cool-gray-60};
   }
 
   &__members {
--- a/pages/ecosystem.vue
+++ b/pages/ecosystem.vue
@@ -192,7 +192,9 @@ const filteredMembers = computed(() => {
 });
 
 function sortMembers(membersToSort: Member[]) {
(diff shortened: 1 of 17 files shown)


techtolentino (commented):
@y4izus 

I think this implementation works, but I also think it's worth seeing how the Education Platform team was able to use the `bx-overflow-menu` — because if I'm not mistaken, we use the same technologies. 

I played around with it a little bit and was able to get the overflow menu kind of working (I had to [import it in our component file here](https://github.com/Qiskit/qiskit.org/blob/main/plugins/components.client.ts)) 

I have a feeling that if we give it another day of investigation and some small rework, we can do it.

If we don't have time, I understand. 

https://github.com/Qiskit/qiskit.org/assets/6276074/4965c19d-7f9f-4c30-aea4-650ea6f582eb

pages/ecosystem.vue
```diff
@@ -44,11 +44,22 @@
...
-            @bx-search-input="searchOnMembers($event.detail.value)"
-          />
+          <div class="ecosystem__toolbar cds--row">
+            <div class="ecosystem__search cds--col-lg-14 cds--col-md-6">
+              <bx-search
+                placeholder="Search using keywords like algorithms, simulator, or machine learning"
+                @bx-search-input="searchOnMembers($event.detail.value)"
+              />
+            </div>
+            <bx-dropdown
```
techtolentino, line 54: I have a suggestion, take it or leave it.

- I think by default, we should sort by `name`
  - we can set the initial value to `sortByValue` (which we'll default to `"name"`)
- I think we should add the `label-text` to say "Sort by"
- we can remove the `trigger-content`

```vue
<bx-dropdown
  class="ecosystem__sort-dropdown cds--col-lg-2 cds--col-md-2"
  label-text="Sort by"
  :value="sortByValue"
  @bx-dropdown-selected="setSortValue($event.detail.item.value)"
>
```

pages/ecosystem.vue
```diff
@@ -243,13 +271,22 @@ const joinAction: Link = {
     margin-top: carbon.$spacing-07;
   }
 
-  &__search {
+  &__toolbar {
+    margin-top: carbon.$spacing-06;
+  }
+
+  &__search,
```
techtolentino, line 278: Since we'll add the label-text, we can remove the `&__sort-dropdown` from needing a `margin-top`. We can leave it on the `&__search` class so it balances out... 

```scss
&__search {
  margin-top: carbon.$spacing-06;

  --cds-field-01: #{carbon.$cool-gray-10};
  --cds-field-04: #{carbon.$cool-gray-30};
}
```

pages/ecosystem.vue
```diff
@@ -243,13 +271,22 @@ const joinAction: Link = {
...
+  &__sort-dropdown {
     margin-top: carbon.$spacing-06;
 
     --cds-field-01: #{carbon.$cool-gray-10};
     --cds-field-04: #{carbon.$cool-gray-30};
   }
 
+  &__sort-dropdown {
```
techtolentino, line 286: this will still need the color overrides:

```scss
&__sort-dropdown {
  --cds-text-01: #{carbon.$cool-gray-60};
  --cds-field-01: #{carbon.$cool-gray-10};
  --cds-field-04: #{carbon.$cool-gray-30};
}
```

techtolentino (changes requested):
@y4izus - we could not get the bx-overflow-menu to work properly

- tried paring w/ Yoshie
- tried asking Carbon team

Turns out, there are several parts of our code that is impacting how the overflow menu behaves, and I think there's too much investigation needed before we can get it working. (we would honestly save time if we just built it from scratch, in my opinion).

For now, I think we can go with your implementation and improve the UX by making a few small updates.

(If you agree, I can make the updates in a commit and add them as well)


https://github.com/Qiskit/qiskit.org/assets/6276074/22cf5ef1-952d-4a60-95b6-a2ed205a99dd

pages/ecosystem.vue
techtolentino: Resolved in https://github.com/Qiskit/qiskit.org/pull/3286/commits/186ea075d560882cfac850ff1f0ef5633910d571

pages/ecosystem.vue
techtolentino: Resolved in https://github.com/Qiskit/qiskit.org/pull/3286/commits/186ea075d560882cfac850ff1f0ef5633910d571

paaragon (approved):

PR #3288, opened 2023-06-02
Updated tags  on UI  cards

mariacloehb:
## Changes
Fixes issue #2852.
Restructured the layout of the. tags in the UI cards so they do not overlap anymore. Also changed the colour of the text and of the background of the tag.
## Implementation details
 Changed the display on the UI card file and changed the colour of the background and text on the variables file.
## Screenshots to see the PR
screenshot_2023-05-29_at_16 28 44


techtolentino (changes requested):
Thanks for the PR @mariacloehb 

Two comments:
1) please revert the changes from the `members.json` file that was committed, we don't need that file to be updated. (t gets automatically included when we do an `npm install`, and for some reason, it's not properly being _gitignored_).

2) When I run the code locally, the card looks like it brought all of the tags to the new row. We want to maintain the _main_ tag with the `purple-70`, then have the rest of the tags in the new row... this is what I see in this PR:

Screenshot 2023-06-02 at 6 30 45 AM

PR #3309, opened 2023-06-08
Refactor ecosystem feat branch

y4izus:
## Changes

Refactors ecosystem page

### DIFF
--- a/components/Ecosystem/EcosystemItemCard.vue
+++ b/components/Ecosystem/EcosystemItemCard.vue
@@ -30,7 +30,7 @@
         {{ description }}
       </p>
     </UiCard>
-    <bx-accordion v-if="testsResults.length != 0">
+    <bx-accordion v-if="hasTestsResults">
       <bx-accordion-item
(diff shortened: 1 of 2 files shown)


components/Ecosystem/EcosystemItemCard.vue
```diff
@@ -0,0 +1,143 @@
...
+        logsLink: res.logsLink,
+      };
+    });
+  }
+
+  return [];
+});
+
+function formatTimestamp(timestamp: number): string {
+  return new Date(timestamp * 1000).toLocaleString("en-UK", {
```
eddybrando, line 99: We're using the locale "en-US" in the rest of the codebase. I think that we should attempt to keep this consistent unless there is a good reason to have multiple locales.

eddybrando (approved):
Thank you for the refactor @y4izus!

I think it brings a lot of value to have the item cards as individual components, since the ecosystem page code was starting to handle too much logic at once.

PR #3313, opened 2023-06-09
 format typewritter effect in headers

### LINKED ISSUE #1104: [General UI] Remove the width constrains for all the titles.
As mentioned in #1097, we can remove the width constrains a check what titles are affected and how.
@JRussellHuffman also proposed several times to check all the titles with the typewriter effect that creates new lines pushing content.

paaragon:
## Changes

Fixes #1104 content jump issue for typewritter effect.

## Diagnostic

The typewriter effect causes the text to change the number of lines it occupies. When the text after the change occupies a number of lines that allows it to continue to be seen, it remains centered in the header correctly.

When it occupies such a number of lines that it would no longer be seen, the browser forces the content to continue to be seen which causes a vertical displacement of the entire content. (The video below shows how the content is displaced).

## Implementation details

To fix the error, line breaks have been set in the header text so that it does not cause new lines at any resolution.

## How to read this PR

Check the following pages with multiple resolutions:

- Ecosystem
- Events
- Providers
- Advocates

If you scroll at the middle of the heading, the content below it should not move when the typewritter effect starts.

## Screenshots

https://github.com/Qiskit/qiskit.org/assets/7396204/32efa257-beaf-4776-9cd8-fa7f486b4845


memolina2323:
Just a thought, wouldn't it look better if instead of having three lines, we could fit the typewritter effect and the "explore" word in the same line?
Screenshot 2023-06-09 at 15 15 51

paaragon:
> Just a thought: wouldn't it look better if instead of having three lines, we could fit the typewritter effect and the "explore" word in the same line? Screenshot 2023-06-09 at 15 15 51
> 
> Nice PR! 👏

Yess it would definetly look better in hight resolutions but in smaller screens, it produces those anoying jumps on the content

memolina2323:
> > Just a thought: wouldn't it look better if instead of having three lines, we could fit the typewritter effect and the "explore" word in the same line? Screenshot 2023-06-09 at 15 15 51
> > Nice PR! 👏
> 
> Yess it would definetly look better in hight resolutions but in smaller screens, it produces those anoying jumps on the content

Is it possible to adapt depending on the screen size then?

memolina2323:
> > > > Just a thought: wouldn't it look better if instead of having three lines, we could fit the typewritter effect and the "explore" word in the same line? Screenshot 2023-06-09 at 15 15 51
> > > > Nice PR! 👏
> > > 
> > > 
> > > Yess it would definetly look better in hight resolutions but in smaller screens, it produces those anoying jumps on the content
> > 
> > 
> > Is it possible to adapt depending on the screen size then?
> 
> Just added a new commit with "responsive breaklines" Hope you like it 😁

Love it! 💯

techtolentino (commented):
Thanks for the PR @paaragon 

A few thoughts:
Before we go down the path of having responsive line break tags, I think we could spend some time reconsidering if the Typewriter effect really brings user value at this point, or is it causing more issues.

Right now, this is happening:
- the typewriter effect causes inconsistent page header heights
  - heights have to be adjusted by using `<br>` tags on the Advocates and Ecosystem page
      - the `<br>` tag needs to be adjusted using responsive CSS
      
One of the original intentions of this feature was to a variety of offerings/opportunities, to the Qiskit community. 

One other possibility would be to use randomization:
- imagine that each time a user loads the page, one of the words in the array, appears
  - this presents the page heading clearly, with each refresh
  - this avoids any accessibility issues with screenreaders, reading the heading
  - this solves the changing height
  - this allows us to avoid CSS fixes
- we use this randomization pattern with the homepage qiskit definition component, and error page



I just wanted to throw out this idea before we commit to a solution... 

thanks!

y4izus:
Per offline conversations, we will continue with the solution Pablo implemented

y4izus (approved):
LGTM!

PR #3319, opened 2023-06-12
fix(footer): increase tap targets space

### LINKED ISSUE #3198: Footer links are not sized appropriately
### Feature description

A Lighthouse reporte showed that the footer links don’t have enough space around them, so that they could be easy to tap without overlapping.

We need to improve the layout of the footer.

This issue affects the UX, accessibility and SEO.

### Additional context

_No response_

eddybrando:
## Changes

This PR increases the space between the text links in the footer in order to have safe tab targets.

Closes #3198

## Implementation details

We conducted a Lighthouse report in https://pagespeed.web.dev/ and it flag some tab targets (the footer links) which were not sized properly:

image

As pointed out in https://developer.chrome.com/docs/lighthouse/seo/tap-targets/, the tab targets should be at least 48x48 in size or have enough spacing between each other.

This PR changes the padding for margin to avoid increasing the vertical size of the tab target and increasing the space to the next footer link, and also increases the margin from 8px to 12px.

A Lighthouse audit conducted locally in the Chrome dev tools didn't flag the tab targets after the change.

## Screenshots

Before:

image

After:

image

### DIFF
--- a/components/Footer/FooterSection.vue
+++ b/components/Footer/FooterSection.vue
@@ -63,7 +63,7 @@ withDefaults(defineProps<Props>(), {
     text-decoration: none;
     display: inline-block;
     width: 100%;
-    padding-bottom: carbon.$spacing-03;
+    margin-bottom: carbon.$spacing-04;
 
     &_theme_light {
       color: qiskit.$text-color-lighter;


y4izus (approved):
LGTM!! Thank you!

PR #3324, opened 2023-06-13
move reusable functionallity to common place

### LINKED ISSUE #3253: DRY Airtable hooks
### Feature description

The files **update-advocates.ts**, **update-ecosystem.ts**, **update-events.ts** and **update-qamp-projects.ts** share some code that could be [_DRYed_](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

### Additional context

There are similarities among these files. They involve importing the **fs** and **util** modules, using `util.promisify` to convert the `fs.writeFile` function into a promise-based version, and writing JSON data to a specified output folder.

We can consider creating a separate utility file that encapsulates the common functionality, like this:

```ts
import fs from "fs";
import util from "util";

export async function writeJSONToFile(
  outputFolder: string,
  filename: string,
  data: any
) {
  const writeFile = util.promisify(fs.writeFile);
  await writeFile(`${outputFolder}/${filename}`, JSON.stringify(data, null, 2));
}
```

Then we can use this function in the other files, like this:

```ts
import { writeJSONToFile } from "./conversion-utils";

...

export default async function (apiKey: string, outputFolder: string) {
  ...
  await writeJSONToFile(outputFolder, "advocates.json", advocates);
}
```

paaragon:
## Changes

 Fix #3253

Create a new promise based method to write json to file.

In the #3253 issue, @eddybrando mentioned a file that I can't find (update-qamp-projects.ts) so the changes are only made in update-advocates.ts, update-ecosystem.ts and update-events.ts

## How to read this PR

This change affects to the content update so you should force this update. To do it you have to add the GENERATE_CONTENT variable with true value (you can do it throught and .env file).

Then start the project and the new file should be created.


y4izus (approved):
LGTM! Thank you!

PR #3325, opened 2023-06-13
change qiskit.org for config variable in vue components

### LINKED ISSUE #3235: Add BASE_URL to nuxt config
### Feature description

As pointed out by @abdonrd - we can avoid having to repeat ourselves, and hardcode our base url (`https://qiskit.org`) every time we need to refer to it.

Let's consider adding it as an environment variable in our nuxt config, then leveraging the [runtimeConfig composable](https://nuxt.com/docs/api/composables/use-runtime-config)

### Additional context

_No response_

paaragon:
## Changes

Fixes #3235

I have used the `public.siteUrl` variable already existing in the nuxt config to change the links in the components that reference the `https://qiskit.org`.

## Implementation details

This change uses the `useRuntimeConfig` method. It does not change the links under the `constant` folder (because they are supposed to be constants).

### DIFF
--- a/components/Events/EventsSummerSchoolFaq2022.vue
+++ b/components/Events/EventsSummerSchoolFaq2022.vue
@@ -256,9 +256,9 @@ const joinqiskitslack: TrackedLink = {
   url: "https://qisk.it/join-slack",
   segment: { cta: "qiskit-slack-join", location: "faq" },
 };
-
+const config = useRuntimeConfig();
 const qiskitEvents: TrackedLink = {
-  url: "https://qiskit.org/events",
+  url: `${config.public.siteUrl}/events`,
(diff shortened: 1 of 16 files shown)


y4izus (changes requested):
Thank you! I've just one comment: could you remove the `members.json` file from the PR? it's a generated file and shouldn't be on this PR

y4izus:
> This file was already there, the change is because it's content it's been updated.

But the content updates are related with the automatic generation, not with your changes, so please, remove the file from the PR

paaragon:
> > This file was already there, the change is because it's content it's been updated.
> 
> But the content updates are related with the automatic generation, not with your changes, so please, remove the file from the PR

I've pushed a commit with the original members.json file

PR #3327, opened 2023-06-13
Refactor/3311 replace v bind

### LINKED ISSUE #3311: Replace `v-bind` with attributes for prop passing
### Feature description

Currently, our codebase sometimes relies on using `v-bind` to pass attributes to child components. This approach has some drawbacks, including decreased code readability and explicitness. To improve code clarity, maintainability, and leverage IDE support, we propose refactoring the codebase to use attributes instead of `v-bind` whenever possible.

### Additional context

- By using attributes directly in the template, the code becomes more readable and easier to understand.
- Explicitly defining props passed through attributes makes the data flow more apparent and aids in understanding component dependencies.
- IDEs with Vue.js support can provide autocompletion and type checking for attributes, enhancing development productivity.
- Refactoring and maintaining the codebase becomes easier as props are directly visible and can be updated without modifying scattered `v-bind` expressions.
- Using attributes promotes consistency with standard HTML attribute syntax and aligns with developer expectations.

paaragon:
## Changes

Fixes #3311 

Complete @eddybrando issue.

Now, instead of using v-bind to pass arguments to components we use only props notation.

There is only one place where I kept the v-bind notation because it is an abstract functionality where we cannot specify props.

### DIFF
--- a/components/Events/EventsSummerSchoolFaq2022.vue
+++ b/components/Events/EventsSummerSchoolFaq2022.vue
@@ -195,6 +195,8 @@
 <script setup lang="ts">
 import { Link } from "~/types/links";
 
+const config = useRuntimeConfig();
+
 const khanAcademy: Link = {
   title: "Khan Academy",
   url: "https://www.khanacademy.org/math/precalculus/x9e81a4f98389efdf:matrices/x9e81a4f98389efdf:multiplying-matrices-by-matrices/v/matrix-multiplication-intro",
@@ -261,8 +263,7 @@ const joinQiskitSlack: Link = {
   segment: { cta: "qiskit-slack-join", location: "faq" },
 };
 
-const config = useRuntimeConfig();
-const qiskitEvents: TrackedLink = {
+const qiskitEvents: Link = {
   title: "Qiskit Events",
   url: `${config.public.siteUrl}/events`,
   segment: { cta: "qiskit-org-events", location: "faq" },
--- a/components/Ui/UiLink.vue
+++ b/components/Ui/UiLink.vue
@@ -3,6 +3,7 @@
     class="link"
     :class="`link_${kind}`"
     :segment="link?.segment"
+    :title="link?.title"
     :url="link?.url"
   >
     <slot />
--- a/components/Events/EventsSummerSchoolFaq2023.vue
+++ b/components/Events/EventsSummerSchoolFaq2023.vue
@@ -189,6 +189,8 @@
 <script setup lang="ts">
 import { Link } from "~/types/links";
 
+const config = useRuntimeConfig();
+
 const pythonAndJupyter: Link = {
   title: "Introduction to Python and Jupyter Notebooks",
   url: "https://learn.qiskit.org/course/ch-prerequisites/introduction-to-python-and-jupyter-notebooks",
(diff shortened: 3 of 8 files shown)


components/Events/EventsSummerSchoolFaq2022.vue
```diff
@@ -185,84 +185,91 @@
...
-const qiskitSubscribe: TrackedLink = {
+const qiskitSubscribe: Link = {
+  title: "Subscribe to Qiskit YouTube",
   url: "https://qisk.it/sub",
   segment: { cta: "qiskit-youtube-subscribe", location: "faq" },
 };
 
-// TODO: Refactor links (https://github.com/Qiskit/qiskit.org/issues/1566)
-const joinqiskitslack: TrackedLink = {
+const joinqiskitslack: Link = {
```
techtolentino, line 258: this variable doesn't follow our casing convention

it should update to `joinQiskitSlack`, same way it shows [here](https://github.com/Qiskit/qiskit.org/pull/3327/files#diff-210de89aa72ac4cf867a2f92001d391e7fa954b96572572b1bb27eed7c58a1f3R163)

techtolentino (changes requested):
Just a small request about our variable naming style... otherwise, LGTM

components/Events/EventsSummerSchoolFaq2022.vue
paaragon: fixed in new commit

components/Ui/UiCta.vue
```diff
@@ -221,9 +221,11 @@ function isIdAnchor(url: string) {
     &#{$arrow-right-path} {
       transform: translate(4px, 0);
     }
+
```
techtolentino, line 224: is this change coming from a lint issue?

techtolentino (approved):
Just two small questions: otherwise, everything looks good to me

🚀

components/Ui/UiLink.vue
```diff
@@ -1,12 +1,21 @@
 <template>
-  <UiBasicLink class="link" :class="`link_${kind}`" v-bind="$attrs">
+  <UiBasicLink
+    class="link"
+    :class="`link_${kind}`"
+    :segment="link?.segment"
+    :title="link?.title"
```
eddybrando, line 6: The `UiBasicLink` component doesn't have a `title` property, so I think that we shouldn't pass this data.

```suggestion
```

eddybrando (changes requested):
Thanks a lot @paaragon! It's looking really good 🙂

I just added one small change request.

components/Ui/UiLink.vue
paaragon: You are right! commit accepted!

components/Ui/UiCta.vue
paaragon: yes, linter shows this solution

components/Ui/UiCta.vue
```diff
@@ -221,9 +221,11 @@ function isIdAnchor(url: string) {
...
     &#{$arrow-down-path} {
       transform: translate(0, 4px);
     }
+
```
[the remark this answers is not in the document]
paaragon: Same as above, it was the linter

eddybrando (approved):

PR #3335, opened 2023-06-14
[ecosystem]: convert stars count

### LINKED ISSUE #3333: [Ecosystem] Fix stars number on cards
Right now, if on the `member.json` file we have `starts: null`, we will have those cards on the first position when ordering by stars. We should convert that `null` to a 0, so we can see the number on the page and it will be ordered appropriately 

**Screenshot:**
Captura de pantalla 2023-06-14 a las 10 41 32

Captura de pantalla 2023-06-14 a las 10 42 36

techtolentino:
## Changes

Fixes #3333 

## Preview
https://qiskit-org-pr-3335.dcq4xc5i083.us-south.codeengine.appdomain.cloud/ecosystem

## Implementation details

- during the fetching of members data, convert `null` stars to `0`

## Screenshots

https://github.com/Qiskit/qiskit.org/assets/6276074/9f915b65-a63d-4259-bdd3-15568abc2736

### DIFF
--- a/hooks/ecosystem-conversion-utils.ts
+++ b/hooks/ecosystem-conversion-utils.ts
@@ -12,7 +12,7 @@ async function fetchMembers() {
     const membersArray: Member[] = [];
     Object.values(res.data).forEach((tier: any) => {
       Object.values(tier).forEach((member: any) => {
-        // convert null to 0
+        // convert falsy values to 0, used for correct sorting on the Ecosystem page
         if (!member.stars) {
           member.stars = 0;
         }
--- a/hooks/ecosystem-conversion-utils.ts
+++ b/hooks/ecosystem-conversion-utils.ts
@@ -12,6 +12,10 @@ async function fetchMembers() {
     const membersArray: Member[] = [];
     Object.values(res.data).forEach((tier: any) => {
       Object.values(tier).forEach((member: any) => {
+        // convert null to 0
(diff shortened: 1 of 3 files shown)


hooks/ecosystem-conversion-utils.ts
```diff
@@ -12,6 +12,10 @@ async function fetchMembers() {
     const membersArray: Member[] = [];
     Object.values(res.data).forEach((tier: any) => {
       Object.values(tier).forEach((member: any) => {
+        // convert null to 0
+        if (member.stars === null) {
```
y4izus, line 16: Not sure of the value we can get. Right now is `null`, but maybe we can get other `falsy` values, so I think it's better to directly check if there are no stars

```suggestion
        if (!member.stars) {
```

Also, could you explain on the comment that we are doing this to help us with the order?

y4izus (changes requested):
Thank you! just a small change :D

hooks/ecosystem-conversion-utils.ts
techtolentino: @y4izus, done [here](https://github.com/Qiskit/qiskit.org/pull/3335/commits/f2cbe3aa966c7eafa2e332b6e0e99f6d525d12e2)

y4izus (approved):

PR #3352, opened 2023-06-19
Create brokenlink-check.yml

### LINKED ISSUE #3366: Bug: Broken links on Qiskit.org
### Where does the issue happen?

http://qiskit.org

### What is the content issue?

There are broken links across the application. We must analyze all the links to find out which ones are broken and fix them.

### Screenshots

_No response_

### Additional information

I attach a first partial analysis of some broken links:
[3_Check.txt](https://github.com/Qiskit/qiskit.org/files/11801632/3_Check.txt)

paaragon:
Fixes #3366

New workflow to find broken links on qiskit.org

## Changes

This workflow prints logs with the broken links  it finds on every page of qiskit.org

## Implementation details

Ideally, it should open a PR with the broken links but, with the initial test, there are so many links that it is not able to finish the execution. We should fix the links it finds now and when it is a more manageable amount, study how to automatically open the PRs.

## How to read this PR

 I attatch a text file with the logs it finds. In this example, some links are included because they give a time out. I thinks those error should not be addressed at the first time and increase the default timeout limit (already increased at 60s).

### DIFF
--- a/.github/workflows/brokenlink-check.yml
+++ b/.github/workflows/brokenlink-check.yml
@@ -12,4 +12,9 @@ jobs:
         uses: ruzickap/action-my-broken-link-checker@v2
         with:
           url: https://qiskit.org
-          cmd_params: "--buffer-size=8192 --max-connections=10 --color=always --skip-tls-verification --exclude=\"(stubs|mylabs.dev|linkedin.com)\" --timeout=60"
+          cmd_params: --buffer-size=8192 \
+                      --max-connections=10 \
+                      --color=always \
+                      --skip-tls-verification \
+                      --exclude="(stubs|https://qiskit.org/documentation)" \
+                      --timeout=60
--- a/.github/workflows/brokenlink-check.yml
+++ b/.github/workflows/brokenlink-check.yml
@@ -12,9 +12,4 @@ jobs:
         uses: ruzickap/action-my-broken-link-checker@v2
         with:
           url: https://qiskit.org
-          cmd_params: --buffer-size=8192 \
-                      --max-connections=10 \
-                      --color=always \
-                      --skip-tls-verification \
-                      --exclude="(stubs|https://qiskit.org/documentation)" \
-                      --timeout=60
+          cmd_params: "--buffer-size=8192 --max-connections=10 --color=always --skip-tls-verification --exclude=\"(stubs|https://qiskit.org/documentation)\" --timeout=60"
--- a/.github/workflows/brokenlink-check.yml
+++ b/.github/workflows/brokenlink-check.yml
@@ -1,0 +1,15 @@
+name: Links
+
+on:
+  workflow_dispatch:
+
+jobs:
+  my-broken-link-checker:
+    name: Check broken links
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check
+        uses: ruzickap/action-my-broken-link-checker@v2
+        with:
+          url: https://qiskit.org
+          cmd_params: "--buffer-size=8192 --max-connections=10 --color=always --skip-tls-verification --exclude=\"(stubs|mylabs.dev|linkedin.com)\" --timeout=60"
--- a/.github/workflows/brokenlink-check.yml
+++ b/.github/workflows/brokenlink-check.yml
@@ -12,4 +12,4 @@ jobs:
         uses: ruzickap/action-my-broken-link-checker@v2
         with:
           url: https://qiskit.org
-          cmd_params: "--buffer-size=8192 --max-connections=10 --color=always --skip-tls-verification --exclude=\"(stubs|https://qiskit.org/documentation)\" --timeout=60"
+          cmd_params: "--buffer-size=8192 --max-connections=10 --color=always --skip-tls-verification --exclude=\"(https://qiskit.org/documentation|https://qiskit.org/ecosystem/.*)\" --timeout=60"


paaragon:
[3_Check.txt](https://github.com/Qiskit/qiskit.org/files/11785896/3_Check.txt)

techtolentino (commented):
As I start to review and understand this PR, just had a couple of initial comments:

1) I was shocked to see all of the broken links. I downloaded the txt file and tested a few of them randomly, and yes all the ones I've tested are in fact broken links. 

2) Is there an issue tied to this PR? What warranted the new workflow update?

3) It would help the reviewer if we had a link to the workflow / action, so we could build our understanding on what this is doing. 



(I need to read up on how this PR is working... so more comments/questions to follow, but my initial thought is that this is very interesting.)

.github/workflows/brokenlink-check.yml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+  my-broken-link-checker:
+    name: Check broken links
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check
+        uses: ruzickap/action-my-broken-link-checker@v2
+        with:
+          url: https://qiskit.org
+          cmd_params: "--buffer-size=8192 --max-connections=10 --color=always --skip-tls-verification --exclude=\"(stubs|mylabs.dev|linkedin.com)\" --timeout=60"
```
Eric-Arellano, line 15: I believe this should work too:

```yml
cmd_params: --buffer-size=8192 \
            --max-connections=10 \
            --color=always \
            --skip-tls-verification \
            --exclude="(stubs|mylabs.dev|linkedin.com)" \
            --timeout=60
```

It will be more maintainable, e.g. if we want to change the `--exclude` list.

Eric-Arellano (approved):
Nice! This makes sense and https://github.com/ruzickap/action-my-broken-link-checker is actively maintained.

In the future, I think we'll want to set this up to run on a regular schedule, which is super easy with GitHub. But I agree that only makes sense when we do something meaningful with the output, rather than just printing to the console. So this is a good first step.

--

We may want to exclude `qiskit.org/ecosystem/*` for now, e.g. `qiskit.org/ecosystem/nature`. They are a bunch of subprojects in the "Qiskit ecosystem". Those projects are important, but the even bigger priority is qiskit.org and qiskit.org/documentation (aka Terra). 

--

Please wait for approval from the core team before merging.

.github/workflows/brokenlink-check.yml
techtolentino: ✋🏼 learning here:

what is logic behind this line `stubs|mylabs.dev|linkedin.com`

I guess I don't know how they related to qiskit.org

paaragon: Hi! If a link contains any of these words, it will be excluded from the analysis (and the underlying links on those pages). 

mylabs.dev and linkedin.com are links from the example I started from. I kept them so as not to forget the notation.

Eric-Arellano: Oh, if they are not relevant to qiskit.org, then we should remove mylabs.dv and linkedin.com

.github/workflows/brokenlink-check.yml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+  my-broken-link-checker:
+    name: Check broken links
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check
+        uses: ruzickap/action-my-broken-link-checker@v2
+        with:
+          url: https://qiskit.org
+          cmd_params: "--buffer-size=8192 --max-connections=10 --color=always --skip-tls-verification --exclude=\"(stubs|https://qiskit.org/documentation)\" --timeout=60"
```
Eric-Arellano, line 15: Did you mean `/ecosystem` rather than `/documentation`? It's important that we don't have broken links for `/documentation`. `/ecosystem` is also important, but less so.

Can you please look more into how the `exclude` option works? I'm wondering whether we need to have `https://qiskit.org` in the entry. And I want to ensure it checks all subpages too.

paaragon: Hi @Eric-Arellano, in an offline conversation we thought that this workflow should focus on everything that is not documentation and that the documentation team should replicate this workflow exclusively for documentation since the volume of broken links there is much higher and it would be better to treat it separately. Do you agree with that idea or do you think there is a better approach?

Regarding the base url, it would not be strictly necessary because I think we do not have more links containing the word "documentation" but I wanted to ensure that this is in that particular position with respect to the base url. With this exactly entry, it would check all the page except for documentation and it's subpages.

paaragon:
Here is a csv with all the broken links on wiskit.org (excluding documentation)
[broken-links.csv](https://github.com/Qiskit/qiskit.org/files/11815214/broken-links.csv)

.github/workflows/brokenlink-check.yml
Eric-Arellano: Okay, that's fine with me. But you will still want to exclude `/ecosystem` too. Specifically, any subpages of `/ecosystem`, like `/ecosystem/finance`. `/ecosystem` itself is good to check.

Eric-Arellano (dismissed):
Looks good to me! But please still wait for approval from the core team.

.github/workflows/brokenlink-check.yml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+  my-broken-link-checker:
+    name: Check broken links
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check
+        uses: ruzickap/action-my-broken-link-checker@v2
+        with:
+          url: https://qiskit.org
+          cmd_params: "--buffer-size=8192 --max-connections=10 --color=always --skip-tls-verification --exclude=\"(https://qiskit.org/documentation|https://qiskit.org/ecosystem/.*)\" --timeout=60"
```
techtolentino, line 15: Just one question: 

is the check actually excluding links from `/ecosystem/.*`?, becuase it looks like they're still populating in the .csv file... 

or are we only excluding files that live _within_ the `/ecosystem` directory, and _not_ subdirectories?



https://github.com/Qiskit/qiskit.org/assets/6276074/40b4dd99-0d04-44df-b93a-0f934ecc093d

paaragon:
> Here is a csv with all the broken links on wiskit.org (excluding documentation) [broken-links.csv](https://github.com/Qiskit/qiskit.org/files/11815214/broken-links.csv)

A new csv without the ecosystem broken links:

[broken-links.csv](https://github.com/Qiskit/qiskit.org/files/11834741/broken-links.csv)

.github/workflows/brokenlink-check.yml
paaragon: Hi @techtolentino, this change was made after this file was generated. I have posted a newer file.

By the way, this workflow is not running yet in this repository (not until we merge this PR). I'm executing it on my own repo where I can execute it every when I need

.github/workflows/brokenlink-check.yml
paaragon: I tested the inline notation and it didn't work

y4izus (approved):
LGTM! Thank you!

y4izus (dismissed):

y4izus (approved):

PR #3369, opened 2023-06-20
feat(error-page): single cta w/ separate handling for docs errors

### LINKED ISSUE #3361: Improves CTAs on the 404 page
Right now we have a 404 page that always shows the same CTAs:
Captura de pantalla 2023-06-20 a las 13 03 09

It would be nice if we can have just one dynamic CTA based on the path the user is trying to going:
- URL = `qiskit.org/documentation/...` => Message = `Back to Qiskit documentation`
- URL = `qiskit.org/<other>` => Message = `Back to Qiskit home page`

techtolentino:
## Changes

Fixes #3361 

Co-authored by: @y4izus 

## Implementation details
- add a computed value for checking if the 404 `isDocumentationError`
  - if so, we render the back to Documentation CTA
  - else, show the normal back to Qiskit org home page CTA

## Preview
- unfortunately, the preview branch doesn't produce the Nuxt error page, and only results in nginx server errors
  
## Screenshots
(Showing dynamic error pages for a `500`, `404` from `/documentation/`, and regular `404`)

https://github.com/Qiskit/qiskit.org/assets/6276074/737ae425-028e-42e4-b694-6fe064aef9ba

### DIFF
--- a/error.vue
+++ b/error.vue
@@ -76,7 +76,7 @@ const backToDocsCta: Link = {
   },
 };
 
-const isDocumentationError = computed(() => {
+const isDocumentationError = computed<boolean>(() => {
   return (
     props.error.statusCode === 404 && route.path.startsWith("/documentation/")
   );
--- a/error.vue
+++ b/error.vue
@@ -12,16 +12,9 @@
               Or as we like to say, <span v-html="randomMessage" />
             </p>
             <UiCta
-              v-if="isDocumentationError"
-              :label="backToDocsCta.label"
-              :segment="backToDocsCta.segment"
-              :url="backToDocsCta.url"
-            />
-            <UiCta
-              v-else
-              :label="backToHomeCta.label"
-              :segment="backToHomeCta.segment"
-              :url="backToHomeCta.url"
+              :label="backToCta.label"
+              :segment="backToCta.segment"
+              :url="backToCta.url"
             />
           </div>
           <div class="cds--col-lg-8">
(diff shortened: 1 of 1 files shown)


error.vue
```diff
@@ -67,13 +69,19 @@ const backToHomeCta: Link = {
...
   url: "/documentation",
-  label: "Back to Qiskit Documentation",
+  label: "Back to Qiskit documentation",
   segment: {
     cta: "back-to-documentation",
     location: "error-page",
   },
 };
 
+const isDocumentationError = computed(() => {
```
eddybrando, line 79: This is just a nice to have, but I personally think that explicitly adding the expected type via generics is a good idea when using computed functions. It's easier to read and understand at a glance what is expected and it's also a safety mechanism for refactors.

```suggestion
const isDocumentationError = computed<boolean>(() => {
```

But it's totally optional.

https://vuejs.org/guide/typescript/composition-api.html#typing-computed

eddybrando (approved):
Nice job, @techtolentino! Super clean, I like it. 🧼 

I only added a tiny suggestion. Totally optional.

error.vue
```diff
@@ -12,16 +12,16 @@
               Or as we like to say, <span v-html="randomMessage" />
             </p>
             <UiCta
-              :label="backToHomeCta.label"
-              :segment="backToHomeCta.segment"
-              :url="backToHomeCta.url"
-            />
-            <UiCta
-              class="error-page__ghost-btn"
+              v-if="isDocumentationError"
```
y4izus, line 15: Instead of doing an `if-else` here, you can calculate the CTA to show in a new variable. So we will have something like this:

```html
<UiCta
    :label="backToCta.label"
    :segment="backToCta.segment"
    :url="backToCta.url"
  />
```

```js

const backToCta = computed<Link>(() => route.path.startsWith("/documentation/") 
? backToDocsCta 
: backToHomeCta);
```

Please, review types and naming because I did it quickly.

One more thing, I've removed the check for the 404 page, because I supposed we also want to have the `Back to` CTA with no matter which error, right?

y4izus (changes requested):
Thank you Randy! could you take a look at my comment?

error.vue
techtolentino: updated [here](https://github.com/Qiskit/qiskit.org/pull/3369/commits/081fc828d249ea0ef09140593eee261683462a35)

y4izus (approved):

PR #3374, opened 2023-06-22
3363 improve the dev workflow with json data

### LINKED ISSUE #3363: Improve the dev workflow with json data

paaragon:
## Changes

Fixes #3363 

We want a development workflow where we don't need the production data or API Key to generate it. We should have a default data mock that the developer can modify to test edge cases or whatever they want.

These changes store the default data in a folder called "mock".

If we set the `MOCK_CONTENT` environment variable, instead of getting the contents of Airtable, it will copy the entire `mock/content` folder into `content/`.

Al the generated content has been removed (some files were still in the repo).

## Implementation details

This change introduces two new environment variables:

- MOCK_CONTENT: if exists, copies the data instead of getting it from Airtable.
- MOCK_BASE_PATH: if a developer wants his own data instead of the existing data in the repository, he can create a new folder with his own data and reference this folder in this variable.

## How to read this PR

Start by setting the `MOCK_CONTENT` environment variable and look at the logs. It should say that the content is being mocked and the site should build without errors.

## Extra

I have done another solution from a different perspective where instead of copying all the data from one folder to another, I copy only the data that already exists. It also makes a check that the mocked data has a correct structure (is more complicated and less flexible for possible future content).

If you want to check it, is in [this branch](https://github.com/Qiskit/qiskit.org/tree/3363-improve-the-dev-workflow-with-json-data)


nuxt.config.ts
```diff
@@ -2,6 +2,7 @@ import fetchEvents from "./hooks/update-events";
 import fetchAdvocates from "./hooks/update-advocates";
 import fetchEcosystemMembers from "./hooks/update-ecosystem";
 import pkg from "./package.json";
+import { generateMocks } from "./hooks/mock/mock-service";
```
eddybrando, line 5: Nice to have: Alphabetically ordered

hooks/mock/mock-service.ts
```diff
@@ -0,0 +1,41 @@
...
+const mockBasePath =
+  process.env.MOCK_BASE_PATH || path.join(__dirname, "content");
+const contentBasePath = path.join(__dirname, "..", "content");
+
+export async function generateMocks() {
+  await copyFiles(mockBasePath, contentBasePath);
+}
+
+/**
+ * This method copies all the file structure on sourceBasePath to targetBasePath
```
eddybrando, line 13:
```suggestion
 * This method copies all the files from sourceBasePath to targetBasePath.
```

hooks/mock/content/events/upcoming-seminar-series-events.json
```diff
@@ -0,0 +1,57 @@
+[
+  {
+    "date": "June 23, 2023",
+    "startDate": "2023-06-23",
+    "endDate": "",
+    "image": "/images/events/downloaded/recXW4BBNFR06ii2a.jpg",
+    "institution": "University of Maryland ",
+    "location": "YouTube",
+    "speaker": "Cheng-Ju Lin ",
```
eddybrando, line 9: I think that it might be a good idea to use fake names and data.

This goes for all mock files.

eddybrando (changes requested):
I really like this approach @paaragon, thanks a lot! 🥳 

I added a couple of comments.

Additionally, thinking about new contributors, it might be a good idea to mock the content data by default, at least if the content directory is empty or if one required content file is missing. Most people might attempt to run `npm install && npm run dev` and expect it to work from the get go without the need of passing a configuration option, and I think that we could achieve to that.

We also need to update our README so that people know that they can use this mock option.

Additionally, we're currently adding some environment variables to the `.env.example` file, and renaming some files. These changes appear to be unrelated to this PR.

hooks/mock/content/events/upcoming-seminar-series-events.json
paaragon: Thanks! I have change the data for everything except for `ecosystem`... for me it looks ok to be the real data but if you think is better to mock this I can do it

hooks/mock/mock-service.ts
paaragon: Done! thanks!

nuxt.config.ts
paaragon: Done!

paaragon:
> I really like this approach @paaragon, thanks a lot! 🥳
> 
> I added a couple of comments.
> 
> Additionally, thinking about new contributors, it might be a good idea to mock the content data by default, at least if the content directory is empty or if one required content file is missing. Most people might attempt to run `npm install && npm run dev` and expect it to work from the get go without the need of passing a configuration option, and I think that we could make that possible and offer a great DX.
> 
> We also need to update our README so that people know that they can use this mock option.
> 
> Additionally, we're currently adding some environment variables to the `.env.example` file, and renaming some files. These changes appear to be unrelated to this PR.

Great idea the default mocked content (I added it).

README updated!

y4izus (approved):

PR #3385, opened 2023-06-26
add schema.org events to head

### LINKED ISSUE #3252: Add Event type to Schema
### Feature description

Add Event type to the existing Schema.org.

### Additional context

Qiskit.org is currently using [UseSchemaOrg](https://nuxt.com/modules/schema-org) to add the schema.org.

This Nuxt3 feature supports a determinate number of types or [identities](https://unhead-schema-org.harlanzw.com/guide/guides/identity) including organization, person, product ...

It would be beneficial for google readability if the type [event](https://schema.org/Event) could be added to the schema.

paaragon:
## Changes

Fixes #3252 

## Implementation details

Add an `application/ld+json` script in the head of `events/index.html` page o add the events information.

## How to read this PR

Run the page in dev mode and look for the script created within the HTML head tag. The correctness of the content can be tested with this page: https://search.google.com/test/rich-results?hl=es


y4izus (changes requested):
As mentioned on the issue, we are using the [useSchemaOrg](https://github.com/Qiskit/qiskit.org/pull/3374) composable, could you use it here, please?

paaragon:
With the last update, google finds 20 valid elements (the correct amount):

https://search.google.com/test/rich-results/result/r%2Fevents?id=z99Wo6f5cTegIPkGur1EGA&hl=es

image

nuxt.config.ts
```diff
@@ -1,10 +1,12 @@
+import { defineNuxtConfig } from "nuxt/config";
 import fetchEvents from "./hooks/update-events";
 import fetchAdvocates from "./hooks/update-advocates";
 import fetchEcosystemMembers from "./hooks/update-ecosystem";
 import pkg from "./package.json";
 
 const { AIRTABLE_API_KEY, GENERATE_CONTENT, NODE_ENV, SITE_URL } = process.env;
 const IS_PRODUCTION = NODE_ENV === "production";
+const siteUrl = SITE_URL || "https://qiskit.org";
```
eddybrando, line 9: Do we need this to be variable?

I'm thinking that maybe we do because of our testing environment, but I'm not sure if it makes a difference or if we could just set it to "https://qiskit.org" always.

eddybrando (changes requested):
Thanks a lot @paaragon, this is very good progress!

I added some specific comments, but also, generally, I believe that there is an issue with the current approach.

We shouldn't add the list of events to every page but only to the pages where they are listed. And also, for individual event pages, we should add the data for the individual event.

nuxt.config.ts
paaragon: There was [an issue](https://github.com/Qiskit/qiskit.org/issues/3235) for change this from static to variable. Here, I just reorder it

paaragon:
> Thanks a lot @paaragon, this is very good progress!
> 
> I added some specific comments, but also, generally, I believe that there is an issue with the current approach.
> 
> We shouldn't add the list of events to every page but only to the pages where they are listed. And also, for individual event pages, we should add the data for the individual event.

You are right, the events list should be referenced only on the event/index page (I have changed it).

I will also add the schema on `fall-fest`, `seminar-series` and `summer school (2022 and 2023)`

composables/useQiskitSchemaOrg.ts
```diff
@@ -0,0 +1,81 @@
+import { CommunityEvent } from "../types/events";
+import rawUpcomingEvents from "~/content/events/upcoming-community-events.json";
```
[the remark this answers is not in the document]
paaragon: This file no longer exists

utils/event-schemaorg.ts
```diff
@@ -0,0 +1,52 @@
+export interface EventSchemaOrg {
```
eddybrando, line 1: I believe that we're not importing this interface anywhere else, so we can remove the export:

```suggestion
interface EventSchemaOrg {
```

eddybrando (changes requested):
Thanks a lot @paaragon! It's looking very good 🙂

I left a couple of comments.

utils/event-schemaorg.ts
```diff
@@ -0,0 +1,52 @@
...
+      url: event.url,
+    });
+  }
+
+  const eventAttendanceMode =
+    event.mode === "Online"
+      ? "OnlineEventAttendanceMode"
+      : "OfflineEventAttendanceMode";
+
+  const schemaEvent: any = {
```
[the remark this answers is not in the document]
paaragon: Done!

pages/events/index.vue
```diff
@@ -282,6 +280,36 @@ const selectTab = (selectedTab: string) => {
...
 };
+
+const sortedEvents = upcomingEvents
+  .filter((event) => event.startDate)
+  .sort((a, b) => {
+    const dateA = new Date(a.startDate);
+    const dateB = new Date(b.startDate);
+
+    return dateA > dateB ? 1 : dateA < dateB ? -1 : 0;
+  });
```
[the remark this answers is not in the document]
paaragon: Done!

eddybrando (changes requested):
There are some issues after updating the branch. I'm not sure what they are related to, but might be due to the Nuxt update.

In any case, when looking at the dependencies, I realised that we're using the package `nuxt-schema-org`. According to them:

> The package `@vueuse/schema-org` has been superseded by [@unhead/schema-org](https://github.com/harlan-zw/unhead-schema-org) which offers the same simple API.

Source: https://nuxt.com/modules/schema-org#new-package

I think that as part of this PR we should replace the package. Maybe that would also bring us closer to the issue after Nuxt's update, which is that `useEventListSchemaOrg` is not found.

paaragon:
> There are some issues after updating the branch. I'm not sure what they are related to, but might be due to the Nuxt update.
> 
> In any case, when looking at the dependencies, I realised that we're using the package `nuxt-schema-org`. According to them:
> 
> > The package `@vueuse/schema-org` has been superseded by [@unhead/schema-org](https://github.com/harlan-zw/unhead-schema-org) which offers the same simple API.
> 
> Source: https://nuxt.com/modules/schema-org#new-package
> 
> I think that as part of this PR we should replace the package. Maybe that would also bring us closer to the issue after Nuxt's update, which is that `useEventListSchemaOrg` is not found.

The issue wasn't about the package but with some bug in the code. 

If you look at unhead's documentation on [how to integrate with nuxt](https://unhead-schema-org.harlanzw.com/integrations/nuxt/module), it says to use the nuxt-schema-org package

eddybrando:
> If you look at unhead's documentation on [how to integrate with nuxt](https://unhead-schema-org.harlanzw.com/integrations/nuxt/module), it says to use the nuxt-schema-org package

But on that page if you click on the GitHub icon on top, you land in the new repo https://github.com/harlan-zw/unhead-schema-org, which is `@unhead/schema-org`.

I believe they just haven't updated their documentation. But, to be fair, it's unclear either way.

I've opened an issue on their repo to ask about that: https://github.com/harlan-zw/unhead-schema-org/issues/24

components/EventsFallFest/EventsFallFestUniversityDirectory.vue
```diff
@@ -43,6 +43,7 @@
 
 <script setup lang="ts">
 import { TextLink } from "~/types/links";
+import { sortEvents } from "~/utils/event-schemaorg";
```
eddybrando, line 46: Exports in the `utils` directory are globally available, so we don't need to import it.

With that in mind, I believe that it's a good idea to have very descriptive names for these functions and files.

Since this particular function sorts all kind of events, I think that it's indeed a good idea to call it `sortEvents`, but I also think that it should be in its own file (something like **utils/sortEvents.ts** and not included in a file related to the Schema.org stuff.

pages/events/index.vue
```diff
@@ -124,6 +121,7 @@ import {
 
 import rawPastEvents from "~/content/events/past-community-events.json";
 import rawUpcomingEvents from "~/content/events/upcoming-community-events.json";
+import { sortEvents } from "~/utils/event-schemaorg";
```
eddybrando, line 124: Same comment as in the **components/EventsFallFest/EventsFallFestUniversityDirectory.vue** file:

> Exports in the `utils` directory are globally available, so we don't need to import it.
> 
> With that in mind, I believe that it's a good idea to have very descriptive names for these functions and files.
> 
> Since this particular function sorts all kind of events, I think that it's indeed a good idea to call it `sortEvents`, but I also think that it should be in its own file (something like **utils/sortEvents.ts** and not included in a file related to the Schema.org stuff.

utils/event-schemaorg.ts
eddybrando, line 1: We introduce a file in kebab-case here. The other file in the **utils** folder is in camel case. Neither one is better than the other IMO and I'm fine with both ways. But we need to be consistent, so I think that we need to either rename this file to **eventSchemaorg.ts** or rename the other file to **url-has-protocol.ts**.

When I wrote the **urlHasProtocols.ts** file I had in mind that the spirit of the **utils** folder was to put individual reusable utils, being mostly files with 1 single exported function. If that's the case, then it would be nice to have the files have the same name as the exported function, so that they are easy to find.

eddybrando (changes requested):
Thanks a lot @paaragon! I added a couple of comments 🙂

components/EventsFallFest/EventsFallFestUniversityDirectory.vue
paaragon: Thanks! I think is a good idea. Changes applied!

utils/event-schemaorg.ts
paaragon: I think we should use kebab-case as we are using in the hooks directory.

I agree with the only one exported function per file 😄

eddybrando (changes requested):

layouts/default-max.vue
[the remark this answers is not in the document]
paaragon: Deleted!

eddybrando (approved):
Thanks a lot @paaragon! I just checked the rich results in the tool linked as well and it looks great.

y4izus (approved):

PR #3388, opened 2023-06-30
perf(ecosystem): quicker list changes

### LINKED ISSUE #3370: Bug: category filter on ecosystem page is very slow

1. Go to qiskit.org/ecosystem
2. Click on any category filter
3. try to unclick the filter
4. wait 5 seconds
5. see that it takes a long time for the filter to become "unchecked"

Not sure how widespread this issue is, everything else on the qiskit.org site loads very quickly for me, but this is feature is super slow. I'm using a chrome browser if that helps

category filter is very slow to uncheck

category filter should uncheck immediately on clicking

### Screenshots

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

### Additional information

_No response_

eddybrando:
## Changes

This PR improves the performance of the ecosystem page when changing the selected tier, the category filters or the sorting option.

We are still able to improve our performance further by moving some data formatting functions from Vue components to the fetching function that stored external data.

I still think that we can continue improving the performance, but we'll have to dig deeper. This might be a task for a later point in time.

Fixes #3370 

## Implementation details

I used [computed properties instead of methods](https://vuejs.org/guide/essentials/computed.html#computed-caching-vs-methods) to leverage caching based on the reactive properties.

I also avoided to render the content of the tabs that were not active.

I also changed the order in which we performed certain operations on the members list to optimise the space and time complexity.

Finally, I sorted the ecosystem project members by name by default during the fetching function, so that we can avoid having to sort during rendering.

### DIFF
--- a/components/Ecosystem/EcosystemItemCard.vue
+++ b/components/Ecosystem/EcosystemItemCard.vue
@@ -65,6 +65,7 @@ const hasTestsResults = computed(() => props.member.testsResults.length !== 0);
 const testRows = computed(() => {
   if (props.member.testsResults) {
     return props.member.testsResults.map((res) => {
+      // TODO: Format properties during data fetching instead for better performance.
       const timestamp = formatTimestamp(res.timestamp);
       // Convert package name to title case
       let packageName;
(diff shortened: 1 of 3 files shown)


paaragon (approved):
LGTM

PR #3430, opened 2023-07-18
fix(seo): removed duplicated meta description tag from head

### LINKED ISSUE #3429: Bug: SEO - meta description tags are included twice

1. Go to qiskit.org
2. Inspect the DOM `<head>`
4. See that meta `description` tags are included twice

meta `description` tags are included twice do to the description being defined in two places:

1. nuxt config
2. useSeoMeta composable

the meta description tag should only be included once

### Screenshots

Screenshot 2023-07-17 at 7 57 08 AM (2)

### Device

_No response_

### OS

_No response_

### Browser

_No response_

### Additional information

_No response_

jackoconnordev:
## Changes

Fix  #3429

Removes duplicated meta description tag in DOM head.

## Implementation details

Removes generic meta description tag which is applied to all pages by `nuxt.config.js`. This leaves `useSeoMeta` as the sole provider of page description tags, which applies unique descriptions to each page for SEO purposes.

## Screenshots

Note 1: In the before screenshot the second description tag uses the description text from the homepage since nuxt is not configured to apply unique descriptions to each page. Keeping `useSeoMeta` seems like the preferred choice of the two options since now in the after image the sole description is specific to the advocates page.

Note 2: `og:description` uses the same text as `description`, but it is a different tag name so is not a straight up duplicate tag. `og:description` gets its text from the same source as `useSeoMeta` and is page specific.

**Before**

image

**After**

image

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -2,7 +2,6 @@ import fetchEvents from "./hooks/update-events";
 import fetchAdvocates from "./hooks/update-advocates";
 import fetchEcosystemMembers from "./hooks/update-ecosystem";
 import { generateMocks } from "./hooks/mock/mock-service";
-import pkg from "./package.json";
 
 const { AIRTABLE_API_KEY, GENERATE_CONTENT, NODE_ENV, SITE_URL, MOCK_CONTENT } =
   process.env;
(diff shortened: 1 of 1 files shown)


techtolentino (approved):
Thanks @jackoconnordev - these updates work for me.

@y4izus @eddybrando - tagging you in case you wanted to review.

Screenshot 2023-07-19 at 6 11 43 AM (2)

PR #3436, opened 2023-07-19
fix(ui-cta): logic and UI for external and internal links

### LINKED ISSUE #1228: [General UI] The icon in the landing CTA should be an arrow

1. Navigate to qiskit.org
2. Notice the icon of the "Get started" CTA

The icon is the one indicating a navigation outside qiskit.org.

Since `qiskit.org/documentation/install.html` is under qiskit.org, the icon should be an arrow.

eddybrando:
## Changes

External links from the same domain (qiskit.org) should look like internal links and open in the same window. Other external links should open in a new window and show the respective icon.

Fixes #1228

## Implementation details

We have 4 scenarios for out CTAs that we want to support:

1. The CTA points to an internal fragment link (starts with `#`)
2. The CTA points to another internal link
3. The CTA points to an external link to a page within the same domain (qiskit.org but outside of this Nuxt project, like the documentation page)
4. The CTA points to another external link

For the user, whether the linked page is part of this Nuxt project or not, is irrelevant. Hence, visually, it should be displayed as an internal link and behave as such, opening in the same window.

The new logic in the *UiCta* supports this cases.

The **UiLink** component has been rewritten as well to support the `target` and `external` attributes' logic for the `NuxtLink` component wrapped. Using a regular anchor component brings no longer any benefit, so it was also removed.

A new [Nuxt util](https://nuxt.com/docs/guide/directory-structure/utils) has been create to check whether a URL has a protocol, based on https://github.com/unjs/ufo/blob/v1.1.2/src/utils.ts#L40.

## How to read this PR

1. Start looking at the **UiCta** logic. It might be easier to avoid looking at the diff first and instead just read the new code to follow along the logic.
2. Review the **UiLink** component's logic similarly.
3. Take a look at the **utils/urlHasProtocol.ts** utility function.
4. Check the fix added to **components/Providers/ProvidersToc.vue**.

## Screenshots

image


components/Ui/UiCta.vue
```diff
@@ -49,35 +48,43 @@ const props = withDefaults(defineProps<Props>(), {
...
+  // Same domain
+  if (urlString.value.startsWith(config.public.siteUrl)) {
+    return false;
   }
-  if (isExternal(url as string)) {
+
+  return isExternalLogically.value;
+});
+
+const isFragment = computed<boolean>(() => {
```
techtolentino, line 75: just curious if there's a difference between fragment and anchor

techtolentino (dismissed):
Awesome PR @eddybrando 

I checked the preview branch, and locally - everything looks great.

I only had some questions for my learning.

🚀

components/Ui/Link/UiLink.vue
```diff
@@ -1,60 +1,83 @@
...
-  </component>
+  </NuxtLink>
 </template>
 
 <script setup lang="ts">
 import { CtaClickedEventProp } from "~/types/segment";
 
+type NuxtLinkTarget =
+  | "_blank"
+  | "_parent"
```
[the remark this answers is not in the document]
eddybrando: That's true, we don't have those scenarios. They are also not the most common in _regular_ websites, but I wanted to include them to give NuxtLink the whole support, just in case.

components/Ui/UiCta.vue
eddybrando: As I understand it, _anchor_ is the `<a>` element and what we're referring to here is if the anchor points to a _document fragment_ via the `href` attribute (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a?retiredLocale=de#href).

components/Ui/Link/UiLink.vue
```diff
@@ -1,60 +1,83 @@
...
+  target?: NuxtLinkTarget;
   title?: string;
-  url?: string;
+  url: string;
 }
 
 const props = withDefaults(defineProps<Props>(), {
+  external: undefined,
   segment: undefined,
+  target: undefined,
```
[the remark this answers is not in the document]
eddybrando: The goal of the `target` prop here is to have the option to pass a the `target` value explicitly and have it been passed forward to `NuxtLink`.

Further down, in the `computedTarget` function we check if the `target` has been explicitly passed (not `undefined`).

If it has been defined, we pass that to `NuxtLink`.

Then we check for the other scenarios that concern us:

- `isExternalFromSameDomain`: `target="_self"`
- `isExternalFromDifferentDomain`: `target="_blank"`

Otherwise we do pass `undefined`, which won't render any `target` attribute in the final anchor's HTML which indeed, as you said, defaults to the same logic as with `target="_self"`.

components/Ui/Link/UiLink.vue
```diff
@@ -1,60 +1,83 @@
...
+    return "_self";
+  } else if (isExternalFromDifferentDomain.value) {
+    return "_blank";
+  }
 
-const isAnchor = computed(() => {
-  return (
-    isExternal.value || isMail.value || isIdAnchor.value || isInternal.value
-  );
+  return undefined;
```
y4izus, line 81: Not sure why `NuxtLinkTarget` has a `null` type but you return here an `undefined`

components/Ui/UiCta.vue
```diff
@@ -49,35 +48,43 @@ const props = withDefaults(defineProps<Props>(), {
...
+  return urlString.value.startsWith("#");
+});
+
+const iconPerLinkType = computed(() => {
+  if (isExternalVisually.value) {
     return Launch16;
-  }
-  if (isIdAnchor(url as string)) {
+  } else if (isFragment.value) {
```
y4izus, line 82: You have a `return` on the previous check, do you need an `else` here?

y4izus (changes requested):
Finally we have this!! thank you!! I've just small requests :D

components/Ui/Link/UiLink.vue
```diff
@@ -1,60 +1,83 @@
...
+type NuxtLinkTarget =
+  | "_blank"
+  | "_parent"
+  | "_self"
+  | "_top"
+  | (string & {})
+  | null;
+
 interface Props {
+  external?: boolean;
```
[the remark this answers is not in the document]
eddybrando: As far as I know optional parameters in Typescript automatically _translate_ to "or `undefined`". (https://www.typescriptlang.org/docs/handbook/2/objects.html#optional-properties)

But this made me realise that I was adding unnecessary `undefined` default values to the props, which can be safely removed. I did that here: https://github.com/Qiskit/qiskit.org/pull/3436/commits/e6d3fcae2d79f096e486d4d95b0263eefb0b67a1

components/Ui/Link/UiLink.vue
```diff
@@ -1,60 +1,83 @@
...
   segment?: CtaClickedEventProp | undefined;
+  target?: NuxtLinkTarget;
```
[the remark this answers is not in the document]
eddybrando: (Please see https://github.com/Qiskit/qiskit.org/pull/3436#discussion_r1270685934)

components/Ui/Link/UiLink.vue
```diff
@@ -1,60 +1,83 @@
...
   title?: string;
```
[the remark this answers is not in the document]
eddybrando: (Please see https://github.com/Qiskit/qiskit.org/pull/3436#discussion_r1270685934)

components/Ui/Link/UiLink.vue
eddybrando: The `target` prop in **NuxtLink** accepts `'_blank' | '_parent' | '_self' | '_top' | (string & {}) | null` and `undefined` since it's an optional Typescript property.

Here we're just trying to not interfere with their logic by passing the original value if one of our special cases doesn't need to rewrite it. The `computedTarget` function can pass both `null` and `undefined`.

components/Ui/UiCta.vue
eddybrando: Changed in https://github.com/Qiskit/qiskit.org/pull/3436/commits/b108d9ca02da6d1ee6ae7f2bb8968cd76865957f.

components/Ui/Link/UiLink.vue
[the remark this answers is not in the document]
eddybrando: As discussed offline, regarding having this prop have `false` as default instead of `undefined`:

> In this case I think that it makes sense to leave it with `undefined` because we might actually not define the value when we pass these props. Also, the check that is done to the `external` prop is to see if it is not _falsey_, which works fine as-is. Additionally, this is closer the NuxtLink's implementation.
> 
> In summary, I think that this is "more correct", having it default to `undefined` in this specific case.

components/Ui/Link/UiLink.vue
y4izus: Understood , thank you!

y4izus (approved):
LGTM! Thank you Eddy!

PR #3453, opened 2023-07-26
feat: add Alchemer Beacon

### LINKED ISSUE #3439: Add Alchemer survey to Qiskit.org
We need to add a survey to our Qiskit.org site with **[Alchemer](https://www.alchemer.com/)** using what they call [website intercept](https://help.alchemer.com/help/website-intercept).

- [x] Add Alchemer's Beacon (JavaScript code) to the site
- [x] ~Create CSS to style survey box~ (The modal uses an iframe with a fixed CSS width inside. We cannot modify it from outside of the iframe)

The scope has been extended after a review:

- [x] Add "feedback" tab (fixed to the right, centered vertically, with text from top to bottom that triggers the survey)
- [x] Change survey text (requires input from @1ucian0)

eddybrando:
## Changes

This PR adds support for Alchemer surveys.

Closes #3439 

## Implementation details

According to [Alchemer's documentation](https://help.alchemer.com/help/website-intercept#create-a-beacon), we need to add their JavaScript code (Beacon) at the end of the `body` tag. This is achieved using the `nuxt.config.ts`'s `app.head.script` configuration.

In order to avoid having a full-width survey (this was requested) we looked at the example page given to us (https://kevinmaddox.com/web-intercept-css.html) and tried our best to create a similar styling for the `.sg-b-in` class, which, after inspecting the injected `intercept.js` JavaScript from the Beacon script, appears to be the wrapper around the iframe. This style also needed to be included in all pages, so it was imported in `main.scss` from the new `_alchemer.scss`.

Also a small sentence was added to the documentation under the "Third-party integrations" chapter.

## How to read this PR

It seems like we can only review this from a code POV, since the survey won't load automatically on staging or local domains, but rather only on the qiskit.org domain (security policy configuration most likely).

If you want to see the style however, you can manually inspect the code in your dev tools in the staging environment and add the following HTML:

```html
<div class="sg-b-in">
  <iframe
    src="//survey.alchemer.com/s3/7446243/Qiskit-org-Feedback-Survey"
    frameborder="0"
    class="sg-b-p-i"
    title="survey"
  ></iframe>
</div>
```

That should bring up the survey somewhere on the page, but I don't think that what is seeing will reflect 1 to 1 to what will be the end result. I was just using this method as a reference for the width adjustments only. I asume some more styles will load with the real script in production to get the div also positioned in front of the site and maybe the height dynamically calculated, but at this point I can only guess.

## Screenshots

Screenshot 2023-07-26 at 17 59 24


techtolentino (commented):
I think this looks good @eddybrando 

My only questions are: 

- will these placeholders get dynamically replaced somehow? 
- is there some configuration where we say "on this route, get feedback about X feature?"

how does this part work... 🤔 





Screenshot 2023-07-26 at 9 55 41 AM

eddybrando:
We're currently working on making the survey available for testing in the preview page. Once we have it we'll be able to make a more precise review.

eddybrando:
The [preview build](https://qiskit-org-pr-3453.dcq4xc5i083.us-south.codeengine.appdomain.cloud/) should now display the survey after 5 seconds in the page.

techtolentino (approved):
this works nicely and I have no issues with the code

thanks @eddybrando

PR #3455, opened 2023-07-28
refactor(layout): make "default-max" the default

### LINKED ISSUE #2267: Remove old "default" layout without max-size support
Implementing the max-size layout in all pages will allow us to consolidate the default layout and DRY some code.

Currently we have two "default layouts". One with and one without the support for the max-size breakpoint. We should only have one default layout (the one that supports said breakpoint).

In order to remove the old default layout, we first need to complete these tasks:

- [x] #1873
- [x] #1874
- [x] https://github.com/Qiskit/qiskit.org/issues/1872

eddybrando:
## Changes

This PR remove the old **default** layout which was no longer used. The actual default was **default-max**, so that one will be the new default layout.

Closes #2267

## Implementation details

* Removed the **default** layout
* Renamed the **default-max** to **default**
* Removed the specific `layout` definitions in the page components so that they use the default layout


techtolentino (approved):
Thank you for this PR @eddybrando 

We've discussed this before so it's finally nice to clean up and refactor.

PR #3457, opened 2023-07-28
ci: update GitHub actions to use Node 16

### LINKED ISSUE #2942: Fix warnings on Build and Deploy gh-actions
The [Build and Deploy](https://github.com/Qiskit/qiskit.org/blob/main/.github/workflows/deploy.yml) gh-action [shows several warnings](https://github.com/Qiskit/qiskit.org/actions/runs/4041781403) because we are using deprecated commands and a deprecated version of Node.

We should update the gh-action to stop using deprecated things

eddybrando:
## Changes

This PR updates [actions/checkout](https://github.com/actions/checkout) and [actions/setup-node](https://github.com/actions/setup-node) to v3 so that they use Node 16.

Closes #2942

## Implementation details

The v2 of both actions still used Node 12. We needed to update to v3 as stated in the GitHub blog: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/


techtolentino (approved):
LGTM

PR #3459, opened 2023-07-28
fix(events): calendar copy feedback text

### LINKED ISSUE #3454: Copy button text change for calendar subscription links
## Current behaviour

Current copy button text after pressing the copy button is "Copy calendar sync url".

This could be misleading as it implies that there is another action to take in order to copy the link text to the clipboard.

## Suggested change

Change the popup text to "Copied calendar sync url" to make it clear that the link has already been copied to the clipboard after pressing the button the first time.

I've included screenshots of the what the suggested change would look like. I'm not a UX guy by any means so if this is not a good idea I'd love to know why.

### Current behaviour

Screenshot 2023-07-26 at 20 44 09

### Suggested Change

Screenshot 2023-07-26 at 20 48 01

eddybrando:
## Changes

This PR changes the feedback text for the "copy calendar URL" action in the Events page to "Copied!".

Closes #3454


jackoconnordev (approved):
Copied looks spelled right to me anyway 😉

techtolentino (approved):
🐇

PR #3460, opened 2023-07-28
fix(providers): support missing CTAs

### LINKED ISSUE #3146: Bug: Providers YAML has invalid link type

This is a code issue.

https://github.com/Qiskit/qiskit.org/blob/main/content/providers/list/1.primitives.yaml#L76 has an object `websiteCta` that is typed as `GeneralLink`, but an object with that type requires a string `label` attribute and a string `url` attribute. The file mentioned defines both of them as `null`.

This is related to this _FIX_ comment: https://github.com/Qiskit/qiskit.org/blob/main/components/providers/ProvidersAccordionLayout.vue#L45

- The data in the YAML file(s) need to conform with the defined types.
- We might need to introduce stricter testing as well to avoid content in the wrong format.
- https://github.com/Qiskit/qiskit.org/blob/main/components/providers/ProvidersAccordionLayout.vue#L45 needs to be fixed. The type check should occur ideally in TypeScript or during automated testing to avoid redundant checking.

### Screenshots

_No response_

### Device

_No response_

### OS

_No response_

### Browser

_No response_

### Additional information

_No response_

eddybrando:
## Changes

This PR supports the providers data missing CTAs.

Closes #3146

## Implementation details

Some providers don't have certain URLs (website, source or docs). We changed the props in **components/Providers/Accordion/ProvidersAccordionContent.vue** to accept the CTAs optionally.

We also removed the CTA objects from the providers data files when the CTA had the URL `null`.

### DIFF
--- a/components/Providers/Accordion/ProvidersAccordionContent.vue
+++ b/components/Providers/Accordion/ProvidersAccordionContent.vue
@@ -33,7 +33,7 @@
 import { TextLink } from "~/types/links";
 import { ProviderCodeExample } from "~/types/providers";
 
-export interface AccordionLayoutProps {
+interface Props {
   title: string;
   description: string;
   installation: string;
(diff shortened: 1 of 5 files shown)


HuangJunye (approved):
Sorry for the delay. I am back to work now. It looks good to me.

eddybrando:
This PR is no longer relevant.

PR #3461, opened 2023-07-28
feat: page transitions

### LINKED ISSUE #2107: Add page transitions
When navigating between routes, Nuxt gives us the option to add transitions pretty easily: https://nuxtjs.org/docs/2.x/features/transitions

This is not only pretty (personal opinion) but also useful for the user to understand navigation changes with visual queues.

eddybrando:
## Changes

This PR adds page transitions.

Closes #2107

## Implementation details

We're implementing simple page transitions as explained in the Nuxt 3 docs: https://nuxt.com/docs/getting-started/transitions

Only internal links with NuxtLink are expected to work. The top header, since it's a web component with regular links, won't trigger animations.

### DIFF
--- a/app.vue
+++ b/app.vue
@@ -1,0 +1,18 @@
+<template>
+  <NuxtLayout>
+    <NuxtPage />
+  </NuxtLayout>
+</template>
+
+<style lang="scss">
+.page-enter-active,
+.page-leave-active {
+  transition: all 0.4s;
+}
+
+.page-enter-from,
+.page-leave-to {
+  opacity: 0;
+  filter: blur(1rem);
+}
+</style>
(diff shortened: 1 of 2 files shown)


techtolentino (commented):
I love the page transitions, it gives the experience an interesting touch, and makes it more smooth as users navigate between pages and gives an sped up page load effect... 

Not sure if we need to get approval before implementing to production; just commenting that I support this PR ✋🏼

y4izus:
Although I like the animation when going to one page to another, I don't think we should merge this change. We are trying to reduce differences when going from a qiskit.org page to another (see https://github.com/Qiskit/qiskit.org/issues/1228) and this change add a clear difference. @1ucian0 , could you give your opinion here?

y4izus:
Per offline conversation, we decided to not merge this PRs, so I close it

PR #3468, opened 2023-07-31
fix(404): documentation link

eddybrando:
Ensure that the documentation link gets treated as an external link logically.

## Changes

This PR fixes the `/documentation` link in the 404 pages.

## Implementation details

Error pages that pointed to `/documentation` had a non-working CTA. We needed to pass the absolute URL (`https://qiskit.org/documentation/`) to ensure that our links' logic treats this link as an external link logically instead of attempting to use internal Nuxt routing logic.

## How to read this PR

In the preview build (or locally), go to a page like `/documentation/this-page-does-not-exist` to get an error page with the documentation CTA, and test if the CTA works when clicked.

### DIFF
--- a/error.vue
+++ b/error.vue
@@ -61,7 +61,7 @@ const backToHomeCta: TextLink = {
 };
 
 const backToDocsCta: TextLink = {
-  url: "/documentation",
+  url: "https://qiskit.org/documentation/",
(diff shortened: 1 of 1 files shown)


techtolentino (approved):
Thanks for the quick fix @eddybrando

Issue #3473, opened 2023-07-31
Bug: immediately remove wrong links


techtolentino:
### Where does the issue happen?

https://qiskit.org/learn/course/quantum-states-and-qubits

### What is the content issue?

The [first two links](https://qiskit.org/learn/course/quantum-states-and-qubits#:~:text=Michael%20Nielsen%2C%20et,Chapters%201%2D3) in the External Recommended Reading section need to be removed or replaced with links to the publisher's site.

- [x] remove incorrect links from all Learn pages (handled in https://github.com/Qiskit/qiskit.org/pull/3474)
- [x] update with correct links to publisher's website (if possible)

### Screenshots

https://github.com/Qiskit/qiskit.org/assets/6276074/19541b0a-d5f3-4336-8070-b1350e4df06d

techtolentino:
@frankharkins - can you help us update these links with ones that take the user to the publisher's official site? 

thank you

PR #3474, opened 2023-07-31
fix: remove incorrect links to readings

techtolentino:
## Changes

Part of https://github.com/Qiskit/qiskit.org/issues/3473

## Implementation details

- removed links that we know take users to an unwanted URL


vabarbosa (approved):
thank you @techtolentino

PR #3443, opened 2023-07-20
feat(ecosystem): tier filters and new header

techtolentino:
## Changes

Closes #3404 

## Implementation details

- [x] remove tier Tabs
- [x] remove results logic
- [x] add Tier filters and filter logic
- [x] update page header with Tier explanation cards

## Preview link
https://qiskit-org-pr-3443.dcq4xc5i083.us-south.codeengine.appdomain.cloud/ecosystem

## How to read this PR

- please review code updates and deletions in `ecosystem.vue`
- need feedback on how to DRY tier and category filters / functionality
- review the updated page header and pictogram selection

PR #3477, opened 2023-08-01
feat(ecosystem): add tier project count to feature cards

### LINKED ISSUE #3472: [Ecosystem] add project count to Leadspace tier cards
Now that we'll have tier cards in the Ecosystem page Leadspace, we should include how many projects each tier has, so that a visitor can have a good idea of the ecosystem.

- calculate how many projects each tier has
- add this count to the Leadspace ecosystem tier cards

### Additional information
Depends on https://github.com/Qiskit/qiskit.org/pull/3443

techtolentino:
## Changes

Fixes #3472 

## Implementation details

- adds function to `getTierProjectCount`
- pass in count into the `joinSectionCards` to be used as a `detail` prop
- added `detail` prop to feature card, and conditionally render if exists
- styled overrides so the feature card is responsive

## How to read this PR

Please review

- method for getting project count (is this the best way?)
- the new `card.detail` prop added
- overriding styles added to ensure the Ecosystem page feature cards are responsive

## Screenshots

https://github.com/Qiskit/qiskit.org/assets/6276074/14ca8e39-d9a8-44d4-b470-f3979d3cbf84

### DIFF
--- a/components/Ui/UiCompactFeature.vue
+++ b/components/Ui/UiCompactFeature.vue
@@ -7,6 +7,9 @@
     <div class="ui-compact-feature__description">
       {{ description }}
     </div>
+    <div v-if="detail" class="ui-compact-feature__detail">
+      {{ detail }}
+    </div>
   </article>
 </template>
 
@@ -15,6 +18,7 @@ interface Props {
   description: string;
   icon: string;
   title: string;
+  detail: string;
 }
 
 defineProps<Props>();
@@ -30,6 +34,7 @@ $feature-icon-size: 2.25rem;
   background-color: qiskit.$background-color-secondary;
   height: 100%;
   padding: carbon.$spacing-05;
+  position: relative;
 
   @include carbon.breakpoint-down(lg) {
     padding-bottom: carbon.$spacing-06;
@@ -52,6 +57,12 @@ $feature-icon-size: 2.25rem;
 
   &__description {
     color: qiskit.$text-color-white;
+  }
+
+  &__detail {
+    color: qiskit.$text-color-white;
+    position: absolute;
+    bottom: carbon.$spacing-05;
   }
 }
 </style>
--- a/components/Ui/UiJoinWithCards.vue
+++ b/components/Ui/UiJoinWithCards.vue
@@ -21,6 +21,7 @@
           :description="card.description"
           :icon="card.icon"
           :title="card.name"
+          :detail="card.detail"
         />
       </div>
     </div>
@@ -34,6 +35,7 @@ interface Card {
   name: string;
   description: string;
   icon: string;
+  detail: string;
 }
 
 interface Props {
--- a/pages/ecosystem.vue
+++ b/pages/ecosystem.vue
@@ -156,17 +156,20 @@ const joinSectionCards = [
   {
     name: "Main tier",
     description: "The main Qiskit packages maintained by IBM Quantum.",
+    detail: `${getTierProjectCount("Main")} projects`,
     icon: "quantum.svg",
   },
   {
     name: "Extensions tier",
     description: "IBM Quantum supported Qiskit extensions.",
+    detail: `${getTierProjectCount("Extensions")} projects`,
     icon: "transform--02.svg",
   },
   {
     name: "Community tier",
     description:
       "Software packages supported by the Qiskit community, not maintained by IBM Quantum.",
+    detail: `${getTierProjectCount("Community")} projects`,
     icon: "group.svg",
   },
 ];
@@ -301,6 +304,17 @@ const filteredMembersSorted = computed<Member[]>(() => {
 
   return filteredMembers.value.sort((a, b) => a.name.localeCompare(b.name));
 });
+
+function getTierProjectCount(tierName: string) {
+  const tierCount = members.reduce((acc, member) => {
+    if (member.tier === tierName) {
+      acc += 1;
+    }
+    return acc;
+  }, 0);
+
+  return tierCount;
+}
 </script>
 
 <style lang="scss" scoped>
@@ -399,6 +413,20 @@ const filteredMembersSorted = computed<Member[]>(() => {
 
   &::part(expando) {
     background-color: carbon.$cool-gray-20;
+  }
+}
+</style>
+
+<style lang="scss">
+@use "~/assets/scss/carbon.scss";
+
+.ecosystem-page__join-section {
+  .join-with-cards__feature {
+    height: 15.5rem;
+
+    @include carbon.breakpoint-down(md) {
+      height: 14.5rem;
+    }
   }
 }
 </style>


components/Ui/UiCompactFeature.vue
```diff
@@ -15,6 +18,7 @@ interface Props {
   description: string;
   icon: string;
   title: string;
+  detail: string;
```
y4izus, line 21: if you have a `v-if` on L10 checking for `detail`, should this line be:

```suggestion
  detail?: string;
```
??

y4izus (changes requested):
just 2 small things Randy!

components/Ui/UiCompactFeature.vue
```diff
@@ -53,5 +58,11 @@ $feature-icon-size: 2.25rem;
   &__description {
     color: qiskit.$text-color-white;
   }
+
+  &__detail {
+    color: qiskit.$text-color-white;
+    position: absolute;
```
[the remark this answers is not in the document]
techtolentino: updated to use flex options here https://github.com/Qiskit/qiskit.org/pull/3477/commits/e15b1773d8a1ad58d7e026508ae06330c11dc3e8

pages/ecosystem.vue
```diff
@@ -301,6 +304,17 @@ const filteredMembersSorted = computed<Member[]>(() => {
 
   return filteredMembers.value.sort((a, b) => a.name.localeCompare(b.name));
 });
+
+function getTierProjectCount(tierName: string) {
```
eddybrando, line 308: To increase type security, I think that we should probably have a type `TierName` exported from the **types/ecosystem.ts** file.

We can also reuse it in that file in the `Member` interface and in the `Tier` interface, and probably in some other places as well.

pages/ecosystem.vue
```diff
@@ -402,3 +416,17 @@ const filteredMembersSorted = computed<Member[]>(() => {
   }
 }
 </style>
+
+<style lang="scss">
+@use "~/assets/scss/carbon.scss";
```
eddybrando, line 421: I think that this import is not necessary.

pages/ecosystem.vue
```diff
@@ -402,3 +416,17 @@ const filteredMembersSorted = computed<Member[]>(() => {
...
+
+.ecosystem-page__join-section {
+  .join-with-cards__feature {
+    height: 15.5rem;
+
+    @include carbon.breakpoint-down(md) {
+      height: 14.5rem;
+    }
+  }
+}
```
eddybrando, line 431: Why do we need to set a fix height here? Is it because of the space between the tier description and the project count? If so, I believe this should be handled with a margin or a flex/grid gap.

components/Ui/UiJoinWithCards.vue
```diff
@@ -34,6 +35,7 @@ interface Card {
   name: string;
   description: string;
   icon: string;
+  detail: string;
```
eddybrando, line 38: This will cause an error in the **advocates** page, where we also use this component.

eddybrando (changes requested):
Thank you very much @techtolentino, it looks very good!

I left a couple of comments.

pages/ecosystem.vue
techtolentino: These heights were set a while ago when they were a part of the `AdvocatesJoin` component, I'm just overriding them to make the project count fit, without cramming content. 

https://github.com/Qiskit/qiskit.org/pull/3443/files#diff-51ac0a67d015e16594a8062805ae88a7ef4741c62735c55e77c618823d262087L75-L90

If we want to refactor these cards, we could handle it in a separate PR?

pages/ecosystem.vue
techtolentino: Doesn't this give us the ability to use the `carbon.breakpoint-down(md)`?

pages/ecosystem.vue
techtolentino: Addressed [here](https://github.com/Qiskit/qiskit.org/pull/3477/commits/bab37d63bbe1aef52efa3d89c1f35d1e544cc5f8)

pages/ecosystem.vue
eddybrando: You're right, I didn't see the breakpoint line.

pages/ecosystem.vue
eddybrando: I think that overriding is not a good idea in this case, to be honest. It leads to more technical debt.

In my opinion, we should get rid of these height rules in the **pages/ecosystem.vue** file. Instead, we should add a top margin in the **components/Ui/UiCompactFeature.vue** for the `ui-compact-feature__copy` `div` containing the `detail` text, and then also change the `height` properties for `min-height` properties for instance in the **components/Ui/UiJoinWithCards.vue** file.

techtolentino: > _I think that overriding is not a good idea in this case, to be honest. It leads to more technical debt._

(I agree, good point)



Addressed [here](https://github.com/Qiskit/qiskit.org/pull/3477/commits/4718ee992e5c939a4ae206cfe5df3dee6470396d)

y4izus (changes requested):

components/Ui/UiCompactFeature.vue
```diff
@@ -1,11 +1,19 @@
 <template>
   <article class="ui-compact-feature">
-    <nuxt-img class="ui-compact-feature__icon" :src="`/images/icons/${icon}`" />
-    <h3 class="h4 ui-compact-feature__title">
-      {{ title }}
-    </h3>
-    <div class="ui-compact-feature__description">
-      {{ description }}
+    <div>
```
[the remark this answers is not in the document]
techtolentino: Since I'm using CSS flex, I used this div to group the content of the card, separately from the detail of the card.

the content div includes the icon, title, and description
the detail div is separate
and w/ css flex, I justify w/ space-between, so that the project count aligns at the bottom, across all 3 cards

otherwise, the other approach to align the project count across 3 cards, was to position absolute...

components/Ui/UiCompactFeature.vue
```diff
@@ -50,8 +62,9 @@ $feature-icon-size: 2.25rem;
     }
   }
 
-  &__description {
+  &__copy {
     color: qiskit.$text-color-white;
+    margin-bottom: carbon.$spacing-05;
```
eddybrando, line 67: Screenshot 2023-08-03 at 17 50 29

I think that this might add too much space at the end of the card. Maybe the best approach here would be to only add a margin-top to the "detail" div.

eddybrando (changes requested):

components/Ui/UiCompactFeature.vue
eddybrando: Nicely implemented in https://github.com/Qiskit/qiskit.org/pull/3477/commits/d1108671665325a9056820f665cacccd06c0665b, thank you!

eddybrando (approved):
Nice job @techtolentino! Thanks a lot! 💪

components/Ui/UiCompactFeature.vue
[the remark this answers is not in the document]
y4izus: is it not enough grouping it with the `<article>`?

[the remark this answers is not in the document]
techtolentino: @y4izus it's not... the `<article>` _is_ the card itself.

- Within the card, we want to have the project count aligned across all 3 cards (unless I'm wrong about this)
  - the way we're doing that is CSS flex
  - to push the project count (`.__detail` div) to the bottom and have it aligned, we justify space-between
  - the `space-between` separates the 'content' (icon, title, and description), from the 'detail'
  
This is what happens if I remove the `<div>`... 

https://github.com/Qiskit/qiskit.org/assets/6276074/5d23c4a9-4cc7-4aa5-a648-0efb99252875

If there's a better way of achieving this and I'm just not seeing it, I'm open to suggestions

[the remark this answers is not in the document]
y4izus: correct me if I'm wrong: when you remove `<div>`, the only thing that is not aligned is the icon, why don't you use `align-self: start;` on it?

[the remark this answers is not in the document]
y4izus: ok, don't spend more time with this, we can merge ;)

y4izus (approved):

PR #3478, opened 2023-08-01
feat(ecosystem): update default cards order

### LINKED ISSUE #3471: [Ecosystem] add sort by tier

techtolentino:
## Changes

Fixes #3471 
Co-authored by @eddybrando 

## Implementation details

- remove conversion utils alphabetic sorting (it wasn't doing anything)
- updated tests to pass
- add `Default` option
- sort cards by default, in alphabetic order, by tier... `['Main', 'Extensions', 'Community']`

## Preview link
https://qiskit-org-pr-3478.dcq4xc5i083.us-south.codeengine.appdomain.cloud/ecosystem

## Screenshots

https://github.com/Qiskit/qiskit.org/assets/6276074/ca56f8b5-8fd3-4e44-945b-b49ecc2a9a4f


pages/ecosystem.vue
```diff
@@ -297,9 +298,11 @@ const filteredMembers = computed<Member[]>(() => {
 const filteredMembersSorted = computed<Member[]>(() => {
   if (selectedSortingOption.value === "stars") {
     return filteredMembers.value.sort((a, b) => b.stars - a.stars);
+  } else if (selectedSortingOption.value === "name") {
+    return filteredMembers.value.sort((a, b) => a.name.localeCompare(b.name));
   }
 
-  return filteredMembers.value.sort((a, b) => a.name.localeCompare(b.name));
+  return filteredMembers.value.sort((a, b) => b.tier.localeCompare(a.tier));
```
eddybrando, line 305: I think that since the inverse alphabetical order (M, E, C) **only coincidentally** matches our "default" order logic we should do one of the following:

- either replace the alphabetical order with an explicit order matching "Main", "Ecosystem" and "Community", or
- add a comment similar to `// HACK: This sorting matches the default order: "Main", "Ecosystem", "Community"` for better code clarity and maintainability

For all intends and purposes of this PR, the second option seems enough to me.

hooks/ecosystem-conversion-utils.ts
```diff
@@ -21,9 +21,6 @@ async function fetchMembers() {
       });
     });
 
-    // Sort ecosystem projects alphabetically
-    membersArray.sort((a, b) => a.name?.localeCompare(b.name));
-
```
eddybrando, line 26: This is actually useful. It sorts the projects alphabetically so that we can avoid having to sort them again in the **pages/ecosystem.vue** file when the user selects the "Name" order option. It reduces the time complexity from **O(n * log n)** to **O(1)**.

pages/ecosystem.vue
```diff
@@ -297,9 +298,11 @@ const filteredMembers = computed<Member[]>(() => {
 const filteredMembersSorted = computed<Member[]>(() => {
   if (selectedSortingOption.value === "stars") {
     return filteredMembers.value.sort((a, b) => b.stars - a.stars);
+  } else if (selectedSortingOption.value === "name") {
+    return filteredMembers.value.sort((a, b) => a.name.localeCompare(b.name));
```
eddybrando, line 302: If we keep the function that sorts the projects by name in the **hooks/ecosystem-conversion-utils.ts**, I believe that we can simply return `filteredMembers.value` here, since they would be already sorted by name, so we can avoid sorting again.

eddybrando (changes requested):
Thank you very much @techtolentino!

I added some change requests 🙂. Let me know what you think.

Additionally, I believe that we should try and sort the projects already during the hook process that fetches and restructures them and put them in the default order. This was done when "Name" was the default, so it should be changed to the new "Default" sorting logic (although I would suggest to sort them by both, first by name and then by tier). With this in place, we could display the page quicker without needing to sort on the fly, since they would be already sorted for the default case.

pages/ecosystem.vue
techtolentino: updated [here](https://github.com/Qiskit/qiskit.org/pull/3478/commits/b64976347e6f182f05b96a45b4f65ead7d93cee2)

pages/ecosystem.vue
techtolentino: updated [here](https://github.com/Qiskit/qiskit.org/pull/3478/commits/b64976347e6f182f05b96a45b4f65ead7d93cee2)

eddybrando (approved):
Nice job @techtolentino! Thanks a lot!

Please check my commits first before merging and let me know if there is anything you would like to change.

PR #3515, opened 2023-08-24
feat(ui-link): remove `noreferrer`

### LINKED ISSUE #3514: [Banner]: update `rel` attribute for CTAs that point to Docs and Learning
### Feature description

As pointed out by @abdonrd:

We should update the homepage announcement banner CTAs to remove the `noreferrer` keyword in the `rel` attribute, because it's preventing us from tracking the referer to the destination... and we want to track when users come from qiskit org.

Screenshot 2023-08-23 at 9 51 35 AM (2)

### Additional context

[rel=noreferrer](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noreferrer) documentation

eddybrando:
When linking to external links, we want to preserve `noopener` but remove `noreferrer`.

## Changes

This PR removes the [`noreferrer`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noreferrer) keyword for external links since we want to be able to use the `Referer` header.

Closes #3514

## Implementation details

We instruct NuxtLink explicitely to only use `noopener` for external links.

https://nuxt.com/docs/api/components/nuxt-link

### DIFF
--- a/components/Ui/Link/UiLink.vue
+++ b/components/Ui/Link/UiLink.vue
@@ -2,6 +2,7 @@
   <NuxtLink
     v-track-click="segment"
     :external="computedExternal"
+    :rel="computedExternal ? 'noopener' : undefined"
     :target="computedTarget"
     :title="title"
     :to="url"


abdonrd:
The preview is not working? Redirect to qiskit.org

abdonrd (approved):
Thanks! It works in the preview!

PR #3563, opened 2023-10-18
Update banners content

aitanaviudes:
## Changes

- Updated the links in banners and notification bars so they point to the corresponding sites.


eddybrando (approved):
LGTM!

PR #3631, opened 2024-01-16
fix: multiple TypeScript issues

eddybrando:
## Changes

This PR fixes several TypeScript issues in the codebase.


y4izus (changes requested):
Thank you! I just have a request to follow consistency. Sometimes you use `import type {...` and sometimes `import { type ...}`. Following [TS docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export), I would suggest `import type {...` form, what do you think?

y4izus (approved):
LGTM

PR #3651, opened 2024-03-14
Turn off PR preview

Eric-Arellano:
We are trying to simplify the https://github.com/Qiskit/gh-actions repo as much as possible, and likely archiving it to inline it into the main existing user: qiskit/documentation.

Now that this repo barely gets development and will be archived soon, we can turn off live previews. You can still preview the app locally.


techtolentino (approved):
LGTM

Thanks @Eric-Arellano

PR #3665, opened 2024-05-05
What do you think about adding an .nvmrc file?

eduardfossas:
Despite not everyone uses nvm, I think it is still good practice to add it despite defining the engines in the package.json. What do you think?


techtolentino:
@eduardfossas - I agree with you, however, this repo and application will soon be deprecated, so for now, we will not be making any updates.

eduardfossas:
@techtolentino, ah, did not know that, sorry! Either way, I am happy to help in the new repo.

